PR #10, opened 2017-06-30
Refactor admin app routers

GeorgeChatzigiannis:
- Relocated logic behind routers to appropriate controllers
- Extracted router helpers
- Split admin route (previously named school) to two routes(school and pupil)


admin/helpers/routes.js
```diff
@@ -0,0 +1,34 @@
+const mongoose = require('mongoose')
+
+const School = require('../models/school')
+const Pupil = require('../models/pupil')
+const Answer = require('../models/answer')
+
+const fetchPupilsData = async (req, next) => {
```
jon-shipley: * This function has dependencies on the session (under `req`) for the school associated with the user, and `next`.  In my view it would be cleaner if this function took only parameter - the school id and threw an error rather than called `next()`.  I don't think it is the responsibility of the helper function to decide what to do.

* The name of this function is `fetchPupilsData` but it returns pupils mongoose objects and school data.  This mix is inconsistent - as one is a plain javascript object, and one is an array of mongoose objects.

* This function does two fetches in series, but they should be done in parallel as the school id is known before the first call.

* This looks like it ought to be in a service (e.g. `pupilService`) rather than a helper function and should behave appropriately (as described above).

GeorgeChatzigiannis: @jon-shipley check my latest commit and let's discuss it from there. Even though the fetchPupilsData returns an object with a lean and a non-lean mongoose object it's quite convenient as in most cases both are required to be used.

admin/controllers/pupil.js
```diff
@@ -0,0 +1,257 @@
...
+const Pupil = require('../models/pupil')
+const errorConverter = require('../lib/error-converter')
+const ValidationError = require('../lib/validation-error')
+const addPupilErrorMessages = require('../lib/errors/pupil').addPupil
+const pupilValidator = require('../lib/validator/pupil-validator')
+const { fetchPupilsData, fetchPupilAnswers } = require('../services/pupilService')
+
+const getAddPupil = async (req, res, next) => {
+  res.locals.pageTitle = 'Add pupil'
+  const schoolId = req.user.School // school object from session
```
jon-shipley: Documentation is incorrect.  School is just an ID here.

GeorgeChatzigiannis: These are carried changes from before.
Let me know if there's anything else.
Updated in the last commit

admin/services/pupilService.js
```diff
@@ -0,0 +1,32 @@
...
+      Pupil.getPupils(schoolId).exec()
+    ]).catch((error) => {
+      throw new Error(error)
+    })
+    return {
+      schoolData,
+      pupils
+    }
+  },
+  fetchPupilAnswers: async (id, next) => {
```
jon-shipley: I don't think a service function should be making decisions like calling `next()` Better just to throw.

GeorgeChatzigiannis: Changed

PR #16, opened 2017-07-04
Feature/newrelic toggle

GuyHarwood:
@GeorgeChatzigiannis this is a simple one, but will need refactoring to use `config` object once you have completed that work.  Can stay here until you are ready with that.


GuyHarwood:
@GeorgeChatzigiannis do you have the capacity to do the above in this sprint?

PR #17, opened 2017-07-05
Project config files

GeorgeChatzigiannis:
- Introduced config files for admin and pupil apps
- Support existing node env variables but could potentially extend to more


admin/app.js
```diff
@@ -19,9 +17,10 @@ const session = require('express-session')
 const MongoStore = require('connect-mongo')(session)
 const breadcrumbs = require('express-breadcrumbs')
 const fs = require('fs')
+const config = require('./config')
 
 mongoose.promise = global.Promise
-const connectionString = process.env.MONGO_CONNECTION_STRING || 'mongodb://localhost/mtc'
+const connectionString = config.MONGO_CONNECTION_STRING || 'mongodb://localhost/mtc'
```
jon-shipley: I would have thought it would make sense to default MONGO_CONNECTION_STRING to localhost in the config, rather than in the calling code.  It could also be used in the migration config.

admin/app.js
```diff
@@ -72,7 +71,7 @@ const mongoStoreOptions = {
 }
 const sessionOptions = {
   name: 'staff-app.sid',
-  secret: process.env.NODE_ENV === 'production' ? process.env.SESSION_SECRET : 'anti tamper for dev',
+  secret: process.env.NODE_ENV === 'production' ? config.SESSION_SECRET : 'anti tamper for dev',
```
jon-shipley: It would make sense to ensure that SESSION SECRET was populated earlier in the code, e.g. during the config initialisation.

admin/config.js
```diff
@@ -0,0 +1,14 @@
...
+  MONGO_CONNECTION_STRING: process.env.MONGO_CONNECTION_STRING,
+  MTC_AUTH_PRIVATE_KEY: process.env.MTC_AUTH_PRIVATE_KEY,
+  MTC_SERVICE: process.env.MTC_SERVICE,
+  NCA_TOOLS_AUTH_URL: process.env.NCA_TOOLS_AUTH_URL,
+  NEW_RELIC_LICENSE_KEY: process.env.NEW_RELIC_LICENSE_KEY,
+  PORT: process.env.PORT,
+  SESSION_SECRET: process.env.SESSION_SECRET,
+  STD_LOG_FILE: process.env.STD_LOG_FILE,
+  TSO_AUTH_PUBLIC_KEY: process.env.TSO_AUTH_PUBLIC_KEY
+}
```
jon-shipley: I think the onus is on the config object to check that all values are populated.  If not then if the NODE_ENV is production then terminate node, if on development it could issue a warning.  We may want to differentiate between values we need on dev and those we don't.  NCA_TOOLS_AUTH_URL will never be set on dev for instance

pupil/config.js
```diff
@@ -0,0 +1,13 @@
...
+  GOOGLE_SVC_PK: process.env.GOOGLE_SVC_PK,
+  GOOGLE_TRACKING_ID: process.env.GOOGLE_TRACKING_ID,
+  GOOGLE_SVC_DOC_KEY: process.env.GOOGLE_SVC_DOC_KEY,
+  MTC_SERVICE: process.env.MTC_SERVICE,
+  MONGO_CONNECTION_STRING: process.env.MONGO_CONNECTION_STRING,
+  NEW_RELIC_LICENSE_KEY: process.env.NEW_RELIC_LICENSE_KEY,
+  PORT: process.env.PORT,
+  SESSION_SECRET: process.env.SESSION_SECRET,
+  STD_LOG_FILE: process.env.STD_LOG_FILE
+}
```
jon-shipley: I think config should ensure that all values are set to non-empty strings or issue a warning on dev and an error on prod.  It should not start up if the config is missing on prod.

admin/app.js
GeorgeChatzigiannis: updated

admin/app.js
GeorgeChatzigiannis: updated

admin/config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  AZURE_STORAGE_CONNECTION_STRING: process.env.AZURE_STORAGE_CONNECTION_STRING,
+  AZURE_STORAGE_LOGGING_ENABLED: process.env.AZURE_STORAGE_LOGGING_ENABLED,
+  GOOGLE_TRACKING_ID: process.env.GOOGLE_TRACKING_ID,
+  MONGO_CONNECTION_STRING: process.env.MONGO_CONNECTION_STRING || 'mongodb://localhost/mtc',
+  MTC_AUTH_PRIVATE_KEY: process.env.MTC_AUTH_PRIVATE_KEY,
+  MTC_SERVICE: process.env.MTC_SERVICE,
+  NCA_TOOLS_AUTH_URL: process.env.NCA_TOOLS_AUTH_URL,
+  NEW_RELIC_LICENSE_KEY: process.env.NEW_RELIC_LICENSE_KEY,
+  PORT: process.env.PORT || '3000',
```
jon-shipley: This should be port 3001 or it will conflict with the pupil app in development.

admin/app.js
```diff
@@ -19,9 +19,23 @@ const session = require('express-session')
...
+const unsetVars = []
+Object.keys(config).map((key) => {
+  if (!config[key] && !whitelist.includes(key)) {
+    unsetVars.push(`${key}`)
+  }
+})
+
+if (unsetVars.length > 0) {
+  console.log(`The following environment variables need to be defined:\n${unsetVars.join('\n')}`)
+  process.exit()
```
jon-shipley: According to https://nodejs.org/api/process.html#process_process_exit_code this is misuse.  They reccommend: 

```
if (unsetVars.length > 0) {
  console.log(`The following environment variables need to be defined:\n${unsetVars.join('\n')}`)
  process.exitCode = 1
  throw new Error('Bad config')
}
```

You can get the error from the shell immediately after the process has exited by looking at `$?` E.g. 

`echo $?`

To confirm that the exit code is non-zero (to indicate an error occurred)

admin/app.js
```diff
@@ -19,9 +19,23 @@ const session = require('express-session')
 const MongoStore = require('connect-mongo')(session)
 const breadcrumbs = require('express-breadcrumbs')
 const fs = require('fs')
+const config = require('./config')
+const whitelist = require('./whitelist')
+
+const unsetVars = []
+Object.keys(config).map((key) => {
+  if (!config[key] && !whitelist.includes(key)) {
```
jon-shipley: I don't think there should be any concept of a whitelist for production - any missing config value should terminate.  The whitelist vars should really be devWhitelist.

admin/app.js
GeorgeChatzigiannis: That's correct I completely forgot about it. It's updated now

admin/app.js
GeorgeChatzigiannis: Yes I made it dev specific

admin/config.js
GeorgeChatzigiannis: I thought I had that pushed but apparently I didn't. Done now

PR #20, opened 2017-07-06
Piping feature

GeorgeChatzigiannis:
Piping can detect changes and hot reload(not browser) without having to restart the server.

Please have a run and let me know your thoughts


jon-shipley:
It looks like `piping` will still be running even in `production`.  If so, I'd suggest not doing that - this should be a developer convenience tool.

GeorgeChatzigiannis:
@jon-shipley made changes to prevent it from running on production

pris54:
@GeorgeChatzigiannis will it also detect SASS changes?

GeorgeChatzigiannis:
@pris54 It does detect SASS changes but we also need to recompile once there is a change.

admin/app.js
```diff
@@ -49,8 +49,8 @@ const index = require('./routes/index')
 const testDeveloper = require('./routes/test-developer')
 const admin = require('./routes/admin')
 
-let app
-if (piping()) app = express()
+if (process.env.NODE_ENV !== 'production') piping()
```
jon-shipley: As per our offline review (@GuyHarwood, @GeorgeChatzigiannis)  - please can you change this check to look for 'development' in order to run `piping`

PR #21, opened 2017-07-06
Feature/13224 code coverage 2

jon-shipley:
Add code coverage for nca-tools-auth-service
Add new devDep `proxyquire`


jon-shipley:
Looking for 2 reviews as the authentication was touched

PR #23, opened 2017-07-07
Accessibility markup.

pris54:
Some of the tools used: 

1) Chrome extension WAVE
2) Chrome extension Validity
3) Chrome extension 'Siteimprove' 
4) Mac Voice Over tool
5) ZoomText

Guidelines:
https://www.w3.org/WAI/intro/wcag
http://webaim.org/techniques/screenreader/#differences

Please note: Some of the tools are more strict than others, what a check tool may report as "redundant role" may come up as a recommendation or even an error on a different tool.

### DIFF
--- a/admin/views/test-developer/manage-check-forms.ejs
+++ b/admin/views/test-developer/manage-check-forms.ejs
@@ -94,7 +94,7 @@
                                 </div>
                             </td>
                             <td tabindex="0" aria-label="Assigned to <%= form.checkWindows.map( cw => { return cw.name;  }).join(', ') || 'pending' %>.">
-                                <%= form.checkWindows.map( cw => { return cw.name;  }).join(', ') || '(Pending)'; %>
+                                <%= form.checkWindows.map( cw => cw.name ).join(', ') || '(Pending)'; %>
                             </td>
                             <td tabindex="0" aria-label="Uploaded on <%= formatGdsDate(form.createdAt) %>."> <%= formatGdsDate(form.createdAt) %> </td>
                         </tr>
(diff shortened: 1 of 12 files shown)


admin/views/test-developer/manage-check-forms.ejs
```diff
@@ -80,14 +83,20 @@
...
+                                           value="<%= form._id %>"
+                                           tabindex="0"
+                                           aria-label="Tick form <%= form.name %>"
+                                           aria-checked="false">
+                                    <label for="check-form-<%= form._id %>"><a href="/test-developer/view-form/<%= form._id %>" aria-label="Edit form <%= form.name %>"><%= form.name %> </a></label>
                                 </div>
                             </td>
-                            <td>
-                                <%= form.checkWindows.map( cw => { return cw.name;  }).join(', '); %>
+                            <td tabindex="0" aria-label="Assigned to <%= form.checkWindows.map( cw => { return cw.name;  }).join(', ') || 'pending' %>.">
```
GeorgeChatzigiannis: Can we change this to: form.checkWindows.map( cw => cw.name ).join(', ')
The curly bracers and return should be redundant after the arrow.

pris54: @GeorgeChatzigiannis yes, I can, but please know that the curly brackets where already there, what I added is the double pipe and the default content. 
I'm always amazed by how you guys never pick up these things in the original PRs... but always in mines...

GeorgeChatzigiannis: @pris54 I was aware they were already there but I thought since you made changes it would be a good time to improve bits of code like this.

pris54: @GeorgeChatzigiannis Change done.

PR #26, opened 2017-07-11
Feature/13548 remove unused models

jon-shipley:
Remove unused models; code cleanup

* removed various pin models from the pupil app as they are no longer needed
* synchronised the pupil and school models between both apps
* moved the code coverage task: run `npm run coverage` to see coverage
* various lint fixes


jon-shipley:
## Pupil app tests
```
55 scenarios (55 passed)
140 steps (140 passed)
4m9.688s
```

## Admin app tests
```
132 scenarios (132 passed)
358 steps (358 passed)
4m59.378s
```

admin/models/school.js
```diff
@@ -5,9 +5,9 @@ const Schema = mongoose.Schema
 const randomGenerator = require('../lib/random-generator')
 
 const School = new Schema({
-  _id: { type: Number },
+  _id: {type: Number},
   leaCode: {
-    type: Number,
+    type: Schema.Types.Number,
```
GeorgeChatzigiannis: Why changed that from Number to Schema.Types.Number?
I was under the impression it was the same

jon-shipley: The approach I took was to diff the pupil and school models and copy the changes from pupil to admin, and then overwrite the pupil model with the more substantial admin model.

So these changes will have come from differences between the admin model and the pupil model.

GeorgeChatzigiannis: In that case we need to keep one for consistency though

PR #32, opened 2017-07-13
12698  Fix page breaks by adding multiple tables

GeorgeChatzigiannis:
CSS page break inside is very inconsistent and does generate issues (_extra page at the end_) when used in a table element.
Only solution I could came up with is use multiple tables.

### DIFF
--- a/admin/views/partials/_content.ejs
+++ b/admin/views/partials/_content.ejs
@@ -1,8 +1,8 @@
-<div id="content" role="main">
+<main id="content" role="main">
     <div class="phase-banner no-print">
         <p tabindex="0" aria-describedby="beta-service">
             <span id="beta-service"><strong class="phase-tag" tabindex="0">BETA</strong>This is a new service – your <a target="_blank" href="http://www.smartsurvey.co.uk/s/mtc-trial-june-2017/">feedback</a> will help us to improve it.</span>
         </p>
     </div>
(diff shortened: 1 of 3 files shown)


GuyHarwood:
@GeorgeChatzigiannis @shekss  one of those test failures (empty result missing) occurred for me yesterday but passed on the subsequent run.  I think QA need to look at this as the result is unreliable.

admin/views/partials/_content.ejs
```diff
@@ -1,8 +1,8 @@
-<main id="content" role="main">
+<div id="content" role="main">
```
jon-shipley: This will affect the main container for every page and is not GDS compliant.

I would suggest changing it back to a `<main>`
 element and applying a print style sheet that would override the style that are causing the extra page to be printed.

Examples:

* https://github.com/UKGovernmentBEIS/rifs-frontend-play/blob/master/src/main/assets/stylesheets/application-print.scss
* https://github.com/alphagov/government-frontend/blob/master/app/assets/stylesheets/print.scss

Although after a cursory look at the stylesheets above, it doesn't look like they are dealing with extra blank pages produced when printing.  

If you can get to the bottom of it and find any evidence that it is caused by the GDS styles then we should raise an issue to at least get the documentation updated to point this out, possibly including our fix if it will work for everybody.

PR #36, opened 2017-07-17
13370 and 13631 Configurable timings. Admin.

pris54:
See stories:
https://agilefactory.visualstudio.com/MTC/_workitems?id=13445&_a=edit
https://agilefactory.visualstudio.com/MTC/_workitems?id=13464&_a=edit

This is the admin side of the stories.

Designs:
http://share.fluent-interaction.co.uk/W50CNV/#g=1&p=service_manager_home
http://share.fluent-interaction.co.uk/W50CNV/#g=1&p=check_settings (click save)

Max allowed values are to be confirmed.

Roles have not been implemented yet, for that reason once you've logged in you must navigate manually to http://localhost:3001/administrator, then follow the 2nd link

PR #40, opened 2017-07-18
13658 13660 custom timing

pris54:
This is the pupil part of stories 13370 and 13631 - Please refer to PR #36 and the stories for further details.
UPDATE (1): public/javascript/mtc-check.js updated to agreed code standards
UPDATE (2): UX requested "Remaining time" text to be removed


admin/lib/validator/settings-validator.js
```diff
@@ -0,0 +1,36 @@
...
+    errorMessage: settingsErrorMessages.loadingTimeLimit
+  }
+}
+
+module.exports.validate = function (req) {
+  return new Promise(async function (resolve, reject) {
+    let validationError = new ValidationError()
+    try {
+      req.checkBody(settingsValidationSchema)
+      let result = await req.getValidationResult()
```
GeorgeChatzigiannis: should be const

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
...
+    })
+  } catch (error) {
+    next(error)
+  }
+}
+
+const getUpdateTiming = async (req, res, next) => {
+  res.locals.pageTitle = 'Check settings'
+  let settings
+  const successfulPost = req.params.status || false
```
GeorgeChatzigiannis: Can be simplified as !!req.params.status

pupil/public/javascripts/mtc-check.js
```diff
@@ -135,20 +135,20 @@
...
-    GOVUK.setTextContent(pageTimer, questionTimeLimit.toString())
-    setInterval(function () {
-      let remaining = Math.ceil((deadline - Date.now()) / 1000)
-      if (remaining < 0) {
-        remaining = 0
-      }
-      GOVUK.setTextContent(pageTimer, remaining.toString())
-    }, 100)
-  }
+  // GOVUK.updatePageTimer = function () {
```
GeorgeChatzigiannis: I am not personally a big fan of keeping commented code in the file since we can always revert if we need using git

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
...
+
+  try {
+    const settingsRecord = await Settings.find().exec()
```
jon-shipley: Would `findOne` be appropriate here instead of `find`?

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
+const Settings = require('../models/setting')
+const SettingsLog = require('../models/setting-log')
+const settingsErrorMessages = require('../lib/errors/settings')
+const settingsValidator = require('../lib/validator/settings-validator')
+const config = require('../config')
+
+const getAdministration = async (req, res, next) => {
```
jon-shipley: File is missing jsDoc

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
...
+    next(error)
+  }
+}
+
+const setUpdateTiming = async (req, res, next) => {
+  res.locals.pageTitle = 'Check settings'
+  let settings
+  let status
+
+  const settingsRecord = await Settings.find().exec()
```
jon-shipley: Would `findOne` be appropriate here?

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
...
+  if (settingsRecord[0]) {
+    settings = settingsRecord[0]
+  } else {
+    settings = new Settings()
+  }
+  settings.questionTimeLimit = Math.round(req.body.questionTimeLimit * 100) / 100
```
jon-shipley: Do we need to validate that the user has provided something that looks like a number before assigning it?

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
...
+  if (validationError.hasError()) {
+    res.locals.pageTitle = 'Check settings'
+    req.breadcrumbs(res.locals.pageTitle)
+    return res.render('administrator/check-settings', {
+      settings: req.body,
+      error: validationError.errors,
+      errorMessage: settingsErrorMessages,
+      breadcrumbs: req.breadcrumbs()
+    })
+  } else {
```
jon-shipley: As you have included `return` in the error short circuit above, you don't need the else part to be indented - it only adds a level of indenting that is not required.

admin/controllers/administrator.js
```diff
@@ -0,0 +1,104 @@
...
+    try {
+      await settings.save()
+      status = 'updated'
```
jon-shipley: This is the only place in the method that status is updated, and it is only used once below in the redirect - so it looks like the redirect could be hard-coded.

pupil/controllers/mtc-check.js
```diff
@@ -42,8 +44,23 @@ const getQuestion = async (req, res, next) => {
     }
   }
 
+  let loadingTime = config.TIME_BETWEEN_QUESTIONS
+  let questionTimeLimit = config.TIME_BETWEEN_QUESTIONS
+
+  try {
+    const timeSettings = await Setting.find().exec()
```
jon-shipley: `findOne` might be useful here.

pupil/controllers/mtc-warmup.js
```diff
@@ -30,8 +32,23 @@ const getQuestion = async (req, res, next) => {
     }
   }
 
+  let loadingTime = config.TIME_BETWEEN_QUESTIONS
+  let questionTimeLimit = config.TIME_BETWEEN_QUESTIONS
+
+  try {
+    const timeSettings = await Setting.find().exec()
```
jon-shipley: `findOne` might be useful here.

pupil/public/javascripts/mtc-check.js
```diff
@@ -1,25 +1,24 @@
...
-  var GOVUK = global.GOVUK || {}
-  var startTime
-  var pageInterval = 5  // seconds allowed per page
-  var preloadInterval = 2 // seconds for student preparation
-  var deadline
-
-  var registerInputs = []
-  var isTouch = 'ontouchstart' in window || navigator.msMaxTouchPoints || navigator.maxTouchPoints ||
+
+  const GOVUK = global.GOVUK || {}
```
jon-shipley: `const` is only available from IE11 onwards - but we must support IE9.

pupil/public/javascripts/mtc-check.js
```diff
@@ -1,25 +1,24 @@
...
+  const defaultQuestionTimeLimit = 5  // seconds allowed per page
+  const defaultPreloadInterval = 2 // seconds for student preparation
+  const isTouch = 'ontouchstart' in window || navigator.msMaxTouchPoints || navigator.maxTouchPoints ||
     (window.Touch && document instanceof Touch) || (window.DocumentTouch && document instanceof DocumentTouch)
+
+  let deadline
```
jon-shipley: `let` won't work in IE9
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

pupil/public/javascripts/mtc-check.js
```diff
@@ -122,60 +124,86 @@
...
-    // set the page timer to the countdown on page load
-    var pageTimer = document.getElementById('js-page-timer')
-    GOVUK.setTextContent(pageTimer, pageInterval.toString())
-
-    setInterval(function () {
-      var remaining = Math.ceil((deadline - Date.now()) / 1000)
-      if (remaining < 0) { remaining = 0 }
-      GOVUK.setTextContent(pageTimer, remaining.toString())
-    }, 100)
+  // GOVUK.updatePageTimer = function () {
```
jon-shipley: Why is this commented out?

pupil/views/check/question-virtual-keyboard.ejs
```diff
@@ -9,13 +9,7 @@ if (typeof nextQuestion === 'number') {
...
 <div class="hidden" id="js-content-div">
 
-    <div class="remaining-time">
-        Remaining time: <span id="js-page-timer" class="bold timer"></span>
-    </div>
-
-    <br style="clear:both"/>
-
-    <div class="question-container">
+    <div class="question-container" id="js-page-time-settings" data-value="<%= questionTimeLimit %>">
```
jon-shipley: I'm sure this will work, but I would have dropped the `questionTimeLimit` and `loadingTime` straight into the javascript with whatever values were correct for the test.  This seems a longer way round to me, and causes an additional dom lookup.  Is there a reason for it?

pupil/views/check/question.ejs
```diff
@@ -9,13 +9,6 @@
 <script src="/javascripts/app.js"></script>
 
 <div class="hidden" id="js-content-div">
-
```
jon-shipley: I didn't see anything in the A/C about dropping the timer.

pupil/public/javascripts/mtc-check.js
GeorgeChatzigiannis: Only way to use this would be with a transpiler like babel but that's part of future conversation

admin/lib/validator/settings-validator.js
pris54: Will be modified in the next commit. How about hdf-validator.js? it's using let as well.

admin/controllers/administrator.js
pris54: There is only one document in this collection, AFAIK we are expecting more fields (further settings) but not further documents so 'find' is meant to find the only document available (that is, the first one found, or [0])

admin/controllers/administrator.js
pris54: Please see the previous explanation, it's the same case.

admin/controllers/administrator.js
pris54: Form validation is in place, nothing but a valid number with up to two decimals can be submitted.

pupil/controllers/mtc-check.js
pris54: See previous explanation re' use of 'find'

pupil/controllers/mtc-warmup.js
pris54: Same

pupil/public/javascripts/mtc-check.js
pris54: I've replaced const with var across the file. It's in next commit.

pupil/public/javascripts/mtc-check.js
pris54: As explained in the PR's description, 'time remaining' text has been removed as per UX request (see story as well).

pupil/views/check/question-virtual-keyboard.ejs
pris54: Let's please have a chat.

pupil/public/javascripts/mtc-check.js
```diff
@@ -1,25 +1,24 @@
...
   'use strict'
-  var GOVUK = global.GOVUK || {}
-  var startTime
-  var pageInterval = 5  // seconds allowed per page
-  var preloadInterval = 2 // seconds for student preparation
-  var deadline
 
-  var registerInputs = []
+  var GOVUK = global.GOVUK || {}
+  var defaultQuestionTimeLimit = 5  // seconds allowed per page
```
jon-shipley: It's a little surprising to see these figures here.   Before this change these figures were stored in a single place, at the top of the mtc-check file.  Now they are stored in two config objects and this file, as well as being overridden by the database values.

Surely, if the default values are in the pupil config then they don't need to be replicated here as well?

admin/controllers/administrator.js
pris54: I'll add it, what should we do about the rest of the files in the project? None of them has jsDoc

pupil/views/check/question.ejs
pris54: Requested by UX (please see PR's description and story discussion)

pupil/public/javascripts/mtc-check.js
pris54: Server side, defaults are stored in config, front-end wise this default value should never be needed but it's in place for peace of mind.

admin/controllers/administrator.js
GeorgeChatzigiannis: I introduced it on (admin) pupilService but I agree here we need to extent the usage throughout the project.

pupil/public/javascripts/mtc-check.js
jon-shipley: I would have thought this would make maintenance harder as tracking down all these values requires prior knowledge.

If you don't share the same point of view then would you consider documenting it all somewhere?

pupil/views/check/question.ejs
jon-shipley: that's fine

PR #46, opened 2017-07-24
Feature/june trial authentication report

jon-shipley:
Add authentication report
Update the psychometrician's report with 2 new columns and a bugfix for the TimeTaken column


admin/bin/psychometrician-report.js
```diff
@@ -73,16 +73,19 @@ function getRow (answer) {
     'TestDate': moment(answer.creationDate).format('YYYYMMDD'),
     'TimeComplete': moment(answer.pupil.checkEndDate).format('h:mm:ss a'),
     'TimeStart': moment(answer.pupil.checkStartDate).format('h:mm:ss a'),
-    'TimeTaken': moment(moment(answer.pupil.checkEndDate).diff(moment(answer.pupil.checkStartDate))).format('h:mm:ss'),
+    'TimeTaken': moment(moment(answer.pupil.checkEndDate).diff(moment(answer.pupil.checkStartDate))).format('HH:mm:ss'),
     'AppId': answer.isElectron ? 'electron' : 'web'
   }
+  const p = (idx) => { return 'Q' + (idx + 1).toString() }
```
GeorgeChatzigiannis: dont need curly bracers and return here

jon-shipley: Fixed

jon-shipley:
@GeorgeChatzigiannis - it's possible that in the future these reports will be moved into some sort of reporting system, but it's too early to tell.  

These have been developed purely as standalone ad-hoc reports, and checking them in is a matter of making sure we don't lose the work.

PR #47, opened 2017-07-25
Feature/rename school upn to urn

jon-shipley:
Business nomenclature fix: rename school.upn to the correct school.urn


jon-shipley:
Looking for 1 dev and 1 test review

PR #48, opened 2017-07-25
Rerun exit code and timings

activemq:
Added Dynamic waiting in some places and also solved exit code issues when the rerun passes but the exit code is still 1

### DIFF
--- a/admin/test/features/step_definitions/manage_check_forms_steps.rb
+++ b/admin/test/features/step_definitions/manage_check_forms_steps.rb
@@ -30,7 +30,7 @@ When(/^I upload a csv file$/) do
 end
 
 Then(/^it should be added to the list of forms$/) do
-  wait_until{expect(manage_check_forms_page.available_checks.rows.count).to eql(@current_form_count + 1)}
+  wait_until{(manage_check_forms_page.available_checks.rows.count).eql?(@current_form_count + 1)}
 end
 
 Given(/^I am ready to upload a csv file$/) do
@@ -79,7 +79,7 @@ When(/^I have assigned the check form to another check window$/) do
 end
 
 Then(/^the check form should be displayed as being assigned to multiple check window$/) do
-  wait_until{expect(manage_check_forms_page.available_checks.rows.first.assigned_to.text).to eql @first_window_name + ', ' + @last_window_name}
+  wait_until{(manage_check_forms_page.available_checks.rows.first.assigned_to.text).eql? @first_window_name + ', ' + @last_window_name}
 end
 
 When(/^I choose to preview the check form questions$/) do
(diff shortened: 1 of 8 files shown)


admin/test/features/step_definitions/manage_check_forms_steps.rb
```diff
@@ -30,8 +30,7 @@
 end
 
 Then(/^it should be added to the list of forms$/) do
-  sleep 2 # need to address this sleep , test fails randomly with out this
-  expect(manage_check_forms_page.available_checks.rows.count).to eql(@current_form_count + 1)
+  wait_until{expect(manage_check_forms_page.available_checks.rows.count).to eql(@current_form_count + 1)}
```
shekss: i think if you put an expectation in the wait utility library, it ignores it. you may have to do this:
wait_until{manage_check_forms_page.available_checks.rows.count ==(@current_form_count + 1)}
expect(manage_check_forms_page.available_checks.rows.count).to eql(@current_form_count + 1)
looks repetitive but is robust.
try it once by giving a false condition like 
wait_until{expect(manage_check_forms_page.available_checks.rows.count).to_not eql(@current_form_count + 1)}
if the step still passes with a not condition, then the wait utility gem ignore expectations.

admin/test/features/step_definitions/manage_check_forms_steps.rb
```diff
@@ -80,8 +79,7 @@
 end
 
 Then(/^the check form should be displayed as being assigned to multiple check window$/) do
-  sleep 2 # need to address this sleep , test fails randomly with out this
-  expect(manage_check_forms_page.available_checks.rows.first.assigned_to.text).to eql @first_window_name + ', ' + @last_window_name
+  wait_until{expect(manage_check_forms_page.available_checks.rows.first.assigned_to.text).to eql @first_window_name + ', ' + @last_window_name}
```
shekss: same here like the comment above

admin/test/features/step_definitions/manage_check_forms_steps.rb
activemq: @shekss I have reworked this to the following : 
wait_until{(manage_check_forms_page.available_checks.rows.count).eql?(@current_form_count + 1)}

I don't think having a expect after this is of any use as if the above fails it will throw a timeout error and if it passes it will pass the expect so we can treat the wait_until as a dynamic waiting assertion.

admin/test/features/step_definitions/manage_check_forms_steps.rb
activemq: Changed using dynamic waiting assertions

PR #51, opened 2017-07-25
Restoring time countdown as requested by PO

pris54:
Related to stories 13631 and 13370


jon-shipley:
The question time limit is still getting set to 2 seconds rather than the 5 expected:

```
<div class="remaining-time" id="js-page-time-settings" data-value="2">
        Remaining time: <span id="js-page-timer" class="bold timer"></span>
    </div>
```

jon-shipley:
There are significant differences between this page and the previous page with timer:

screen shot 2017-07-26 at 03 39 11

GuyHarwood:
@shekss Would you prefer we get the headless chrome work done, and then come back to this PR?

shekss:
@GuyHarwood  the headless bit is mostly done in this PR  but i have more questions on it, I would still prefer a separate PR. Me and @pris54 wanted to get this PR demoed asap. as soon as the tests pass here on headless chrome, I suggest we merge this PR.

PR #55, opened 2017-07-26
Foundation for Pupil SPA

jon-shipley:
This is the default angular app for the pupil check

* replaced favicon with GDS one
* added polyfills for IE8 and above


pupil-spa/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see http://editorconfig.org
```
GeorgeChatzigiannis: do we need editorconfig? isn't standardJS covering most of this when used on the IDE(webstorm etc)

GuyHarwood: i would rather it was left in

jon-shipley: As we are using typescript we will be using tslint - rules are defined in `tslint.json` - which we can tweak.

I don't think there is any harm in leaving it in, and perhaps much good.

PR #66, opened 2017-08-02
Add new check complete component - static page

jon-shipley:
Add new component CheckComplete to display the /check/complete page.


jon-shipley:
Looking for 1 approval

Please note that this page is routed via '/check/complete' and is not currently linked to from any other page.  I can tweak that if required... just let me know.

PR #77, opened 2017-08-08
Add Question Service

jon-shipley:
Add QuestionService


jon-shipley:
QuestionService now depends on the StorageService - please review @GuyHarwood @pris54

PR #78, opened 2017-08-08
Feature/add auth guard to routes

jon-shipley:
Add a guard to protect routes that require the user be authenticated


jon-shipley:
Conflicts resolved: please can I get 1 review from @GuyHarwood or @pris54 to make this mergeable?

PR #80, opened 2017-08-08
Feature/update question component

jon-shipley:
Port HTML over from the old pupil app for the question page.

Pick up factor1 & factor1 vars from @Input() definition, so the component can be initialised like this:
`<app-question [factor1]="4" [factor2]="5"></app-question>`


jon-shipley:
Please can I get new approvals on this PR now that I have resolved the conflicts?

PR #83, opened 2017-08-09
Remove Promises from StorageService (tests will be fixed in separate branch)

GuyHarwood:


pupil-spa/src/app/storage.service.ts
```diff
@@ -3,40 +3,30 @@ import { Inject, Injectable } from '@angular/core';
...
-      } else {
-        localStorage.setItem(key, value);
-        resolve();
-      }
-    });
+  setItem(key, value): void {
+
+    if (!key) {
+      throw new Error('key is required');
+    } else {
```
jon-shipley: There is no need for the `else` branch - it can just read 

```
if (bailoutCondition) {
 throw ()
}
return expression
```

This is cleaner code I think.

GuyHarwood: @jon-shipley fixed in the upcoming PR for test promise removal

PR #85, opened 2017-08-09
Feature/create check component

jon-shipley:
Initial cut of the CheckComponent

* viewState is managing the component display - this may need to expand to include the entire check process
* questionService could be expanded to save question state and store it in case the app got reloaded


pupil-spa/src/app/question.service.ts
```diff
@@ -55,4 +54,22 @@ export class QuestionService {
...
+  getConfig() {
+    return this.config;
+  }
+
+  reset() {
+    this.questions = null;
+    this.config = null;
+  }
+
+  initialise() {
```
GuyHarwood: i assume this will be called from the constructor?

GuyHarwood: is there any reason why you aren't put the contents of `initialise()` in the constructor?  It would save you from having to call it in every test.

pupil-spa/src/app/check/check.component.html
```diff
@@ -0,0 +1,5 @@
+<div [ngSwitch]="viewState">
```
jon-shipley: No.  Why?  Is there a better way?

pupil-spa/src/app/question.service.ts
jon-shipley: QuestionService is a singleton, and the state is kept throughout the lifetime.  So, we need to reload the questions on login and clear them on logout.

pupil-spa/src/app/check/check.component.html
GuyHarwood: I missed the fact that it was loading those in :)

pupil-spa/src/app/question.service.ts
GuyHarwood: makes sense 😎

PR #86, opened 2017-08-10
Feature/user service unit test fixes

pris54:
Unit test modifications.


pupil-spa/src/app/user.data.mock.json
```diff
@@ -0,0 +1,23 @@
+{
```
GuyHarwood: if you look in `/design/spa/checksession.json` you will find the actual spec for the user data.  Probably best to replace it as many of the fields below do not exist in our dataset

pris54: This file was deleted... its' not needed, I'll recheck.

PR #87, opened 2017-08-10
Feature/14512 user service update

GuyHarwood:
Better structure to promises and tests


pupil-spa/src/app/user.service.ts
```diff
@@ -2,48 +2,47 @@ import { Injectable } from '@angular/core';
...
-      if (err) {
-        return reject(err);
-      }
-      if (data) {
-        return resolve(data);
-      }
+          const data = response.json();
+          this.loggedIn = true;
+          this.storageService.setItem(sessionDataKey, data['session']);
+          this.storageService.setItem(questionsDataKey, data['questions']);
```
jon-shipley: I think we also need to set the config section too?

pupil-spa/src/app/user.service.ts
```diff
@@ -2,48 +2,47 @@ import { Injectable } from '@angular/core';
...
+        (err) => {
+          reject(err);
+        }).catch(error => reject(error));
     });
   }
 
   logout() {
-    this.storageService.removeItem(auth_token);
-    this.storageService.removeItem('data');
+    /*   this.storageService.removeItem(auth_token);
```
jon-shipley: Why are these now commented out?  Aren't they needed?

jon-shipley:
Please merge the latest master in and resolve the conflicts.
.

pupil-spa/src/app/user.service.ts
GuyHarwood: yes, its not in the checkSessionSchema.json, so will add that and the assertion

pupil-spa/src/app/user.service.ts
GuyHarwood: good catch.  I hadn't saved the file!

PR #91, opened 2017-08-14
run ng test during travis build on SPA

GuyHarwood:
This PR can be considered complete when the travis build executes the unit tests via the `test-single` script added to package.json.

The pupil tests have also been omitted as this is no longer under development and will soon be removed.


GuyHarwood:
all working as expected

PR #92, opened 2017-08-14
strongly type Storage Keys to avoid mismatches

GuyHarwood:
created Union Type called StorageKey which narrows the keys used in the storage service to those which reflect each storage entity (questions, config, session etc...)

refactored question service and spec to accommodate this as tests were failing.

### DIFF
--- a/pupil-spa/src/app/question.service.spec.ts
+++ b/pupil-spa/src/app/question.service.spec.ts
@@ -13,7 +13,9 @@ describe('QuestionService', () => {
       getItem() {
       }
     };
-    spyOn(mockStorageService, 'getItem').and.returnValue(JSON.stringify(responseMock));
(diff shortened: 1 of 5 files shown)


pupil-spa/src/app/question.service.spec.ts
```diff
@@ -29,6 +31,8 @@ describe('QuestionService', () => {
   it('getQuestion() returns a Question', inject([QuestionService], (service: QuestionService) => {
     service.initialise();
     const q = service.getQuestion(1);
+    //console.log('question:', q);
```
GeorgeChatzigiannis: do you want to commit this commented code?

GuyHarwood: good spot, removed

PR #101, opened 2017-08-16
revert code that cleared storage when loading login component.

GuyHarwood:
we should only clear storage on logout, currently.

next sprint includes work to time out the session.


activemq:
Tested that once logged in if the user navigates to the sign in page the local storage is not cleared. Removed test scenario that is now invalid

PR #105, opened 2017-08-17
Base api url environment variable per config and gulp task

GeorgeChatzigiannis:
Load api url variable from environment and include gulp replace task for build process

### DIFF
--- a/pupil-spa/gulpfile.js
+++ b/pupil-spa/gulpfile.js
@@ -1,0 +1,8 @@
+const gulp = require('gulp')
+const replace = require('gulp-string-replace')
+
+gulp.task('replace', () => {
+  gulp.src(['./src/environments/environment.prod.ts'])
+    .pipe(replace('API_URL', process.env.API_URL))
+    .pipe(gulp.dest('./src/environments/'))
+})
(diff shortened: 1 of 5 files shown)


pupil-spa/gulpfile.js
```diff
@@ -0,0 +1,8 @@
+const gulp = require('gulp')
+const replace = require('gulp-string-replace')
+
+gulp.task('replace', () => {
```
GuyHarwood: please give it a more distinct name, such as 'setAppEnvironmentVariables'

pupil-spa/src/environments/environment.prod.ts
```diff
@@ -1,3 +1,4 @@
 export const environment = {
+  apiURL: 'API_URL',
```
GuyHarwood: i couldnt get this to work, but then realised i had overwritten the entry with undefined.  I had to switch it back so it would pick up the real value.  This could be problematic in development, but we should only really be updating this file during the docker image build, outside of local development.  If it proves problematic then we should have a separate 'template' file for environment.prod.ts and update the target.

pupil-spa/gulpfile.js
GeorgeChatzigiannis: done

PR #107, opened 2017-08-17
Feature/14848 instructions page config

jon-shipley:
Remove hardcoded numbers from the Instructions page and use the actual config vars instead

### DIFF
--- a/pupil-spa/src/app/instructions/instructions.component.html
+++ b/pupil-spa/src/app/instructions/instructions.component.html
@@ -5,8 +5,8 @@
 <div>
   <ul class="text list list-bullet">
     <li>This check has <strong>{{count}} questions</strong>.</li>
-    <li>Each question will show for <strong>5 seconds</strong>.</li>
-    <li>There will be a <strong>2-second</strong> pause between each question.</li>
+    <li>Each question will show for <strong>{{questionTime}} seconds</strong>.</li>
+    <li>There will be a <strong>{{loadingTime}}-second</strong> pause between each question.</li>
     <li>You can answer the questions using:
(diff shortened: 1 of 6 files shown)


pupil-spa/src/app/instructions/instructions.component.html
```diff
@@ -5,8 +5,8 @@ <h1 class="heading-xlarge">Instructions</h1>
 <div>
   <ul class="text list list-bullet">
     <li>This check has <strong>{{count}} questions</strong>.</li>
-    <li>Each question will show for <strong>5 seconds</strong>.</li>
-    <li>There will be a <strong>2-second</strong> pause between each question.</li>
+    <li>Each question will show for <strong>{{questionTime}} seconds</strong>.</li>
+    <li>There will be a <strong>{{loadingTime}}-second</strong> pause between each question.</li>
```
activemq: Do we need the '-', as now its reads like:
"There will be a 3-second pause between each question."
@jon-shipley

jon-shipley: @activemq - that's exactly as the copy is intended to read, and indeed did read when it was hardcoded to `2-second pause`

jon-shipley:
@activemq - heads up - I suspect this has broken the build.  Can you review the cucumber tests please?  They are now dynamic and can be either from config or the the database.

pupil-spa/src/app/instructions/instructions.component.html
activemq: 👍

activemq:
35 scenarios (35 passed)
79 steps (79 passed)
1m17.946s

PR #111, opened 2017-08-17
Integrate Audit Service calls

GuyHarwood:
# WIP DO NOT MERGE

This PR is merely to enable in-context discussion on the code itself to aid resolution.

Problem:  assertions on the calls to `auditService.addEntry` by the component fail, stating it was called zero times.  It is in fact called by the component during the test run.

The only way i have managed to get the assertion to pass is to comment out the fake created for component.ngAfterViewInit, which was explicitly added by @jon-shipley to avoid the timeout calls.

See inline comments added below for more info.

### DIFF
--- a/pupil-spa/src/app/question/question.component.spec.ts
+++ b/pupil-spa/src/app/question/question.component.spec.ts
@@ -1,14 +1,21 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { QuestionComponent } from './question.component';
+import { AuditService } from "../audit.service";
+import { AuditServiceMock } from "../audit.service.mock";
+import { QuestionRenderedAuditEntry } from "../auditEntry";
 
 describe('QuestionComponent', () => {
   let component: QuestionComponent;
   let fixture: ComponentFixture<QuestionComponent>;
+  let auditServiceMock = new AuditServiceMock();
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      declarations: [ QuestionComponent ]
+      declarations: [QuestionComponent],
+      providers: [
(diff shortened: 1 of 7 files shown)


pupil-spa/src/app/question/question.component.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('QuestionComponent', () => {
     component = fixture.componentInstance;
     fixture.detectChanges();
     // prevent Timeout's being set
-    spyOn(component, 'ngAfterViewInit').and.returnValue(null);
+    //spyOn(component, 'ngAfterViewInit').and.returnValue(null);
```
GuyHarwood: @jon-shipley commenting this out and then explicitly calling `component.ngAfterViewInit()` within the test is the only way i have managed to make the test pass.

pupil-spa/src/app/question/question.component.spec.ts
```diff
@@ -115,23 +115,28 @@ describe('QuestionComponent', () => {
...
-        console.log('addEntry called with:', entry);
+      let auditService = fixture.debugElement.injector.get(AuditService);
+      spyOn(auditService, 'addEntry').and.callFake((entry) => {
+        auditEntryInserted = entry;
       });
     });
-    it('adds audit entry on question rendered', async(() => {
+    it('is added on question rendered', async(() => {
       fixture.whenRenderingDone().then(() => {
+        component.ngAfterViewInit();
```
GuyHarwood: explicit call to ngAfterViewInit - feels unnecessary

GuyHarwood:
Working after merge with master, will discuss in the morning. 😴

PR #116, opened 2017-08-18
Fix asset copying

jon-shipley:


GuyHarwood:
cloned fresh repo and works locally. 👍

PR #121, opened 2017-08-21
Fix admin unit tests - call chain has now changed

jon-shipley:
Unit test fix to reflect latest controller code


jon-shipley:
Tests are good: 
```
Finished in 0.983 seconds
102 tests, 233 assertions, 0 failures, 0 skipped
```

PR #122, opened 2017-08-22
Feature/14928 integrate answer service

GeorgeChatzigiannis:
- Answer service integration with check component
- Added tests

### DIFF
--- a/pupil-spa/src/app/check/check.component.ts
+++ b/pupil-spa/src/app/check/check.component.ts
@@ -35,6 +35,14 @@ export class CheckComponent implements OnInit {
     this.config = this.questionService.getConfig();
   }
 
+  @HostListener('document:touchend', [ '$event' ])
+  handleTouchEndEvent() {
+    // IMPORTANT: Prevent double-tap zoom on Ipad
+    event.preventDefault();
+    event.target.dispatchEvent(new Event('click'));
+    return false;
+  }
+
   ngOnInit() {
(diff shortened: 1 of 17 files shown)


jon-shipley:
@George - please resolve the conflicts so I can review

GeorgeChatzigiannis:
@jon-shipley resolved

pupil-spa/src/app/check/check.component.ts
```diff
@@ -48,11 +49,15 @@ export class CheckComponent implements OnInit {
 
   manualSubmitHandler(answer: string) {
     // console.log(`check.component: manualSubmitHandler(): ${answer}`);
+    const answerSet = { factor1: this.question.factor1, factor2: this.question.factor2, answer: parseInt(answer, 10) };
```
jon-shipley: I think should be a string here as we need to record leading zero's `in the answer`.

jon-shipley: It would also be good to store the sequence number too.  Is this Q1 or Q9?

GeorgeChatzigiannis: @jon-shipley modified the answer to be a string

GeorgeChatzigiannis: @jon-shipley is that required? I would assume you will always have the answers in the proper order, hence you will be able to distinguish the sequence.

PR #126, opened 2017-08-24
14517 Submission service implementation

GeorgeChatzigiannis:
- Submission service includes a call to get all local storage items and attempt a post to store data in db.
- #### Http call is currently commented out until implemented
- Tests included for both storage and submission service


pupil-spa/src/app/services/submission/submission.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(private http: Http, private storageService: StorageService) {
+  }
+
+  async submitData() {
+    const localStorageData = this.storageService.getAllItems();
+    const headers = new Headers();
+    headers.append('Content-Type', 'application/json');
+    const requestArgs = new RequestOptions({headers: headers});
+    // TODO: build Check complete API call
```
GeorgeChatzigiannis: Will be uncommented once the api call exists and tests will be amended/improved accordingly

pupil-spa/src/app/services/question/question.service.ts
```diff
@@ -55,9 +57,9 @@ export class QuestionService {
 
     if (currentQuestionNumber === this.getNumberOfQuestions()) {
       // we are already on the last question, there isn't another question number.
+      this.submissionService.submitData().catch(error => new Error(error));
```
jon-shipley: Putting the submission service here is incongruous, and violates the single responsibility principle.

GeorgeChatzigiannis: @jon-shipley What's wrong with calling a service from another service?

GeorgeChatzigiannis: I see what you mean but this block really did nothing before apart from exiting.

jon-shipley: @GeorgeChatzigiannis  - so a caller calls `getNextQuestionNumber()` an as a side-effect it submits to the backend?  That's not good.

jon-shipley: @GeorgeChatzigiannis - control flow for the Check itself is managed by the Check Component - the submission must be made from there.

PR #127, opened 2017-08-24
Feature/14780 implement auth token

jon-shipley:
JWT

Add JWT to the questions API
Store the JWT the pupil-spa
New jwtService in the admin app that can verify the incoming token


admin/spec/lib/validator/hdf-validation-spec.js
```diff
@@ -16,9 +16,9 @@ describe('HDF validator', function () {
     }
   }
 
-  function notAllowed () {
-    return [':', ';', ',', '.', '!', '@', '#', '$', ' ', '%', '^', '&', '*', '(', ')', '_', '+', '=']
-  }
+  // function notAllowed () {
```
GeorgeChatzigiannis: do you need these comments?

admin/services/jwt-service.js
```diff
@@ -0,0 +1,73 @@
...
+      iss: 'MTC Admin',                                       // Issuer
+      sub: pupil._id,                                         // Subject
+      exp: Math.floor(Date.now() / 1000) + (60 * 60),         // Expiry
+      nbf: Math.floor(Date.now() / 1000),                     // Not before
+      jwi: jwtId                                              // JWT token ID
+    }
+
+    // Construct a JWT token
+    const token = await jwt.sign(payload, pupil.jwtSecret)
+    return token
```
GeorgeChatzigiannis: you don't need the variable here you can just return await jwt.sign(...

admin/spec/lib/validator/hdf-validation-spec.js
jon-shipley: removed!

admin/services/jwt-service.js
jon-shipley: true - but I _like_ it this way.  The compiler will handle it.

GeorgeChatzigiannis:
@jon-shipley conflicts

PR #133, opened 2017-08-29
feature/14932-check-api

GeorgeChatzigiannis:


admin/models/completed-checks.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+const mongoose = require('mongoose')
+mongoose.Promise = global.Promise
+const Schema = mongoose.Schema
+
+const CompletedChecks = new Schema({
+  data: {
+    type: Object,
```
jon-shipley: @GeorgeChatzigiannis @GuyHarwood - do we want the model to be more restrictive?  Checking that all required keys are present?

GuyHarwood: @jon-shipley we discussed this yesterday.  In hindsight, i think we can be fairly confident about the keys that are going to be used.  But if we do make it more restrictive we have to be mindful that this will need changing, should we introduce new keys.

pupil-spa/src/app/services/submission/submission.service.ts
```diff
@@ -11,21 +11,20 @@ export class SubmissionService {
...
-  //       },
-  //       () => {
-  //       }).catch(error => new Error(error));
+    await this.http.post(`${environment.apiURL}/api/completed-check`,
+      { audit, inputs, answers, access_token },
+      requestArgs)
+      .toPromise()
+      .then((response) => {
+          if (response.status !== 201) {
+            return new Error('Login Error:' + response.status + ':' + response.statusText);
```
jon-shipley: This needs to handle other response codes doesn't it?

admin/models/completed-checks.js
jon-shipley: fine

PR #142, opened 2017-09-05
Bug fix for feedback unit tests

pris54:


pupil-spa/src/app/services/feedback/feedback.service.ts
```diff
@@ -19,7 +19,7 @@ export class FeedbackService {
       const accessToken = this.storageService.getItem('access_token');
 
       if (!storedFeedback || !accessToken) {
-        return reject('Missing data in local storage');
+        return false;
```
GeorgeChatzigiannis: I just noticed that this is wrapper in both a promise and an async func.
I think one of these two is only required.
I'd keep just the async/await for simplicity, see the example on submission.service.ts

pupil-spa/src/app/services/feedback/feedback.service.ts
```diff
@@ -9,43 +9,37 @@ export class FeedbackService {
...
-      const storedFeedback = this.storageService.getItem('feedback');
-      const accessToken = this.storageService.getItem('access_token');
+    const storedFeedback = this.storageService.getItem('feedback');
+    const accessToken = this.storageService.getItem('access_token');
 
-      if (!storedFeedback || !accessToken) {
-        return reject('Missing data in local storage');
-      }
+    if (!storedFeedback || !accessToken) {
+      return false;
```
jon-shipley: async functions will turn a `return` into a `resolve` and a `throw` into a `reject`.  Judging by the previous code we should use `throw` here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

GuyHarwood:
i'm getting this error when running `ng test`.....

```
ERROR: 'Unhandled Promise rejection:', 'Cannot read property 'postFeedback' of undefined', '; Zone:', 'ProxyZone', '; Task:', 'Promise.then', '; Value:', TypeError{__zone_symbol__currentTask: ZoneTask{_zone: Zone{_properties: ..., _parent: ..., _name: ..., _zoneDelegate: ...}, runCount: 0, _zoneDelegates: null, _state: 'notScheduled', type: 'microTask', source: 'Promise.then', data: undefined, scheduleFn: undefined, cancelFn: null, callback: function () { ... }, invoke: function () { ... }}}, 'TypeError: Cannot read property 'postFeedback' of undefined
    at http://localhost:9876/_karma_webpack_/webpack:/Users/guy/dev/MTC/pupil-spa/src/app/feedback/feedback.component.spec.ts:129:34
```

PR #150, opened 2017-09-07
move old pupil app away from root

GuyHarwood:
`/pupil` moved to `/design/reference/old-pupil-app`

Why?  less keystrokes in tab completion to get to SPA directory 🥂


jon-shipley:
We may need to tell codacy to ignore this path.

PR #152, opened 2017-09-07
Feature/13940 add single pupil page

jon-shipley:
Show the add-pupil screen


admin/assets/stylesheets/application.scss
```diff
@@ -310,3 +315,17 @@ table.spacious-two-column {
     width: 50%;
   }
 }
+
+.master-detail {
+  margin-top: 1.1em;
```
pris54: @jon-shipley whenever possible I'd like to stick to pixels, we've got a few scattered units out there but in general I've been trying to use only pixels and 'gutter' variables as much as possible.

jon-shipley: @pris54 - I believe we *ought* to standardise on using the pem() function.
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/functions.md

Would you agree to that?  We would specify in pixels, but output would be in em?

activemq:
Looks like something broke, looking in to it

PR #171, opened 2017-09-14
Feature/13944 edit page

GeorgeChatzigiannis:
Ready for review


admin/views/school/_pupil-form.ejs
```diff
@@ -92,7 +92,7 @@ if (formData._id) {
                     </p>
                     <p>
                         If your pupil does not have a UPN there is
-                        <a target="_blank" href="https://www.gov.uk/government/publications/unique-pupil-numbers">
+                        <a href="https://www.gov.uk/government/publications/unique-pupil-numbers">
```
jon-shipley: This looks odd to me.  Why are we removing the `target="_blank"`?  I only added it in recently.  Did the requirements change already or is this some sort of merge error?

GeorgeChatzigiannis: Probably my mistake there, I undid this change

PR #175, opened 2017-09-15
Add/edit check windows.

pris54:
Stories 15007 and 15097


pris54:
@activemq could you please look at the error thrown by Travis?

admin/models/check-window.js
```diff
@@ -11,28 +11,30 @@ const CheckWindow = new Schema({
...
     }]
   }
 }, {timestamps: true})
 
+/**
+ * Retrieve all check windows and sort by arguments passed.
+ * @param sortBy
+ * @param direction
+ * @returns {Promise}
+ */
```
jon-shipley: Good!

admin/routes/index.js
```diff
@@ -41,60 +39,7 @@ router.get('/profile', isAuthenticated(), (req, res) => getProfile(req, res))
 /* Administration page */
 router.get('/administrator', isAuthenticated(config.ROLE_TEST_DEVELOPER), (req, res) => getAdministration(req, res))
 /* Health check */
-async function getPing (req, res) {
```
jon-shipley: We seem to have lost some functionality here.`getPing()` should restore commitId and build number now.

admin/views/administrator/check-windows-form.ejs
```diff
@@ -0,0 +1,199 @@
...
+                </label>
+                <div class="form-date">
+                    <div class="form-group form-group-day">
+                        <label class="form-label" for="checkStartDay">Day</label>
+                        <input class="form-control <%= error.checkStartDay ? 'form-control-error' : '' %>"
+                               id="checkStartDay"
+                               name="checkStartDay"
+                               type="number"
+                               value="<%= checkWindowData.checkStartDay %>"
+                               aria-required="true"
```
jon-shipley: Good to see the aria-* attributes used here. Perhaps you could come up with documentation to standardise this (in another PR).

admin/controllers/administrator.js
```diff
@@ -195,9 +198,137 @@ const getCheckWindows = async (req, res, next) => {
...
+      errorMessage: checkWindowErrorMessages,
+      actionName,
+      urlActionName,
+      breadcrumbs: req.breadcrumbs()
+    })
+  }
+
+  if (req.body.checkWindowId !== '') {
+    try {
+      checkWindow = await CheckWindow.findOne({_id: req.body.checkWindowId}).exec()
```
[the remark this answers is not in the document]
pris54: @jon-shipley that's true however, if there is a human error during editing, the id will come for sure from the hidden field

admin/views/administrator/check-windows-form.ejs
pris54: @jon-shipley Good idea, I'll bring the need for a document during sprint planning.

admin/routes/index.js
pris54: @jon-shipley Odd... Code restored.

admin/controllers/administrator.js
```diff
@@ -195,9 +198,137 @@ const getCheckWindows = async (req, res, next) => {
...
+ */
+const checkWindowsForm = async (req, res, next) => {
+
+  let error = ''
+  let errorMessage = ''
+  let checkWindowData = ''
+  let successfulPost = false
+  let actionName = (req.params.action === 'add' && req.params.id === undefined) ? 'Create' : 'Edit'
+  let urlActionName = req.params.action
+  let currentYear = Date.now()
```
jon-shipley: This is not exactly the current year...

admin/controllers/administrator.js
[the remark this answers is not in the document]
jon-shipley: I'm not fussed about the param - but in one you use `findOne` and in the other you use `CheckWindow.getCheckWindow`.  Seems inconsistent.  Are they doing the same thing?

admin/routes/index.js
```diff
@@ -94,7 +93,7 @@ function getBuildNumber () {
   })
 }
 
-router.get('/ping', (req, res) => getPing(req, res))
+router.get('/ping', (req, res) => res.status(200).send('OK'))
```
jon-shipley: Still doesn't look right... should return the commit and and buildId

admin/controllers/administrator.js
pris54: Fixed, please have a look

admin/routes/index.js
pris54: This has to be a bad merge... I've fixed it.

admin/controllers/administrator.js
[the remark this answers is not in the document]
pris54: @jon-shipley yes, you're right, getCheckWindow was created later on. I've replace line 290 to use this method too.

PR #181, opened 2017-09-19
Feature/13943 ux changes

jon-shipley:
UX review changes

* Add padding between the green highlight bar and the pupil name
* Reduce the height of the green highlight bar


jon-shipley:
New look: 

screen shot 2017-09-19 at 10 08 37

admin/views/school/pupil-register.ejs
```diff
@@ -21,7 +21,7 @@
         <table id='register-pupils' class="spacious">
             <thead>
                 <tr>
-                    <th scope="col">
+                    <th scope="col" style="width: 66%">
```
GeorgeChatzigiannis: can this be included in css to avoid inline css?
I know we've been doing this for a while but we'll be losing track of what's going on with the styles if we continue with this practice.

jon-shipley:
Updated look:
screen shot 2017-09-20 at 11 16 37

admin/assets/stylesheets/application.scss
```diff
@@ -258,6 +258,13 @@ input {
   @include button($grey-3);
 
   margin-top: em(20);
+  &:not(:last-child) {
+    margin-right: pem(10px);
+  }
+
+  @include ie(8) {
```
pris54: @jon-shipley Should this be moved to application-ie8.css?

jon-shipley: @pris54 - yes that's exactly what happens.  application-ie8.css is a generated file from these includes.

PR #182, opened 2017-09-19
Feature/15594 pupils not taking check scaffold

pris54:


admin/views/school/pupils-not-taking-check.ejs
```diff
@@ -0,0 +1,40 @@
...
+    <div class="column-two-thirds" id="checkWindowsForm">
+        <h1 class="heading-xlarge" id="heading-xlarge" tabindex="0">Pupils not taking the check</h1>
+
+        <p class="lede">Click on the button below to add reasons for pupils who are not taking the check</p>
+
+        <p>
+            <a class="button button-primary" href="#">Add reason for pupil</a>
+        </p>
+
+        <br />
```
jon-shipley: Why are all these `br`s here?

pris54: @jon-shipley to add space between the button and the back to top link, so to match the prototype. I don't want a new css class just to add space.

GeorgeChatzigiannis: Can we replace brs with css to match the expected behaviour?

pris54: @GeorgeChatzigiannis I really don't want another CSS class just to add space once ever.

PR #189, opened 2017-09-20
Bug/pupil register page fix

jon-shipley:
Update the sidebars following UX review on the pupil-register page

### DIFF
--- a/admin/views/administrator/administration-home.ejs
+++ b/admin/views/administrator/administration-home.ejs
@@ -37,12 +37,16 @@
         </div>
     </div>
 
-    <div class="column-one-third sidebar-top-border" tabindex="0" role="group" aria-label="sidebar">
-        <h3 class="heading-medium" tabindex="0">Before you start</h3>
-        <ul class="list">
-            <li>
-                <a href="/PDFs/MTC_administration_guidance_June-2017-trial.pdf" target="_blank">Guidance</a>
-            </li>
-        </ul>
+    <div class="column-one-third add-title-margin" tabindex="0" aria-label="sidebar">
+        <aside role="complementary" class="support-column">
+            <h3 class="heading-medium" tabindex="0">Before you start</h3>
+            <nav role="navigation">
+                <ul class="list">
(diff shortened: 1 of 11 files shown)


admin/views/administrator/administration-home.ejs
```diff
@@ -37,12 +37,16 @@
...
-                <a href="/PDFs/MTC_administration_guidance_June-2017-trial.pdf" target="_blank">Guidance</a>
-            </li>
-        </ul>
+    <div class="column-one-third add-title-margin" tabindex="0" aria-label="sidebar">
+        <aside role="complementary" class="support-column">
+            <h3 class="heading-medium" tabindex="0">Before you start</h3>
+            <nav role="navigation">
+                <ul class="list">
+                    <li>
+                        <a href="/PDFs/MTC_administration_guidance_June-2017-trial.pdf" target="_blank">Guidance</a>
```
GeorgeChatzigiannis: I am not a big fan of exposing our directories in the frontend.
How about we handle this with a router get method?
Same as my PR.

admin/views/administrator/check-windows.ejs
```diff
@@ -53,21 +53,28 @@
...
-                <a href="#">Progress report</a>
-            </li>
-        </ul>
+    <div class="column-one-third add-title-margin" tabindex="0" aria-label="sidebar">
+        <aside role="complementary" class="support-column">
+            <h3 class="heading-medium" tabindex="0">Before you start</h3>
+            <nav role="navigation">
+                <ul class="list">
+                    <li>
+                        <a href="/PDFs/MTC_administration_guidance_June-2017-trial.pdf" target="_blank">Guidance</a>
```
GeorgeChatzigiannis: same call here

admin/views/administrator/administration-home.ejs
jon-shipley: @GeorgeChatzigiannis I think this is a topic for a different discussion.  We may not even want Express to serve static assets.

PR #191, opened 2017-09-21
15764/force ssl on azure for admin app

GuyHarwood:
Azure terminates SSL at router and `x-forwarded-proto` header is not received by app.

Instead we get `x-arr-ssl` to indicate whether request is secure or not.  

This patch re-introduces the `x-forwarded-proto` header so express `req.protocol` works correctly and we redirect request to https if insecure request detected.

### DIFF
--- a/admin/app.js
+++ b/admin/app.js
@@ -27,7 +27,7 @@ const flash = require('connect-flash')
 const helmet = require('helmet')
 const config = require('./config')
 const devWhitelist = require('./whitelist-dev')
-const forceSsl = require('express-enforces-ssl')
+const azure = require('./azure')
 
 const unsetVars = []
 Object.keys(config).map((key) => {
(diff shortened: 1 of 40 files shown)


pupil-spa/nginx/forcehttps.conf
```diff
@@ -0,0 +1,10 @@
+# removed until working
+
+  # if not running locally
+  # try $server_port != 3002
+  if ($remote_addr !=0.0.0.0) {
```
jon-shipley: looks like you need a space after `!=`

admin/package.json
```diff
@@ -30,6 +30,7 @@
     "express": "~4.15.2",
     "express-breadcrumbs": "0.0.2",
     "express-busboy": "^6.0.1",
+    "express-enforces-ssl": "^1.1.0",
```
jon-shipley: If adding new packages please also update the yarn lock file (for travis)

admin/app.js
```diff
@@ -81,10 +84,38 @@ app.use(helmet.contentSecurityPolicy({
...
 app.use(helmet.hsts({
   maxAge: oneYearInSeconds,
-  includeSubDomains: true,
-  preload: true
+  includeSubDomains: false,
+  preload: false
 }))
 
+// identify azure by specific environment variable
+function isAzure () {
```
GeorgeChatzigiannis: Seems like a helper function to me.

I would think this can be moved outside of app.js to keep it cleaner.

@jon-shipley @GuyHarwood opinions?

admin/package.json
GuyHarwood: thanks for the heads up, adjustment incoming...

admin/app.js
GuyHarwood: will create azure specific file for this

PR #198, opened 2017-09-22
Feature/16000 - integrate azure application insights

GuyHarwood:
Provides rich telemetry in Azure when configured via environment variable.

for more info see here - https://docs.microsoft.com/en-us/azure/application-insights/app-insights-overview


admin/azure.js
```diff
@@ -1,8 +1,17 @@
+let appInsights = require('applicationinsights')
```
jon-shipley: should be const

PR #207, opened 2017-09-28
Submit multiple pupils data

GeorgeChatzigiannis:
Tests, service introduction - refactory kept for next PR


admin/controllers/pupil.js
```diff
@@ -37,16 +42,16 @@ const getAddPupil = async (req, res, next) => {
...
 const postAddPupil = async (req, res, next) => {
   res.locals.pageTitle = 'Add pupil'
+  req.breadcrumbs(res.locals.pageTitle)
   let school
   try {
     school = await School.findOne({_id: req.body.school}).exec()
-    if (!school) {
-      throw new Error(`School [${req.body.school}] not found`)
-    }
+    // if (!school) {
```
GuyHarwood: why not just delete it?

GeorgeChatzigiannis: Will be on the next commit, I forgot about it.

admin/controllers/pupil.js
```diff
@@ -95,27 +81,142 @@ const postAddPupil = async (req, res, next) => {
...
+      if (csvData.some(p => p[6])) {
+        const errorsCsv = []
+        headers.push('Errors')
+        errorsCsv.push(headers)
+        csvData.forEach((p) => errorsCsv.push(p))
+        const writeToString = promisify(csv.writeToString)
+        const cvsStr = await writeToString(errorsCsv, { headers: true })
+        // Upload csv to Azure
+        try {
+          const remoteFilename = moment().format('YYYYMMDDHHmmss') + '-' + 'error.csv'
```
GuyHarwood: i would use something more unique here - it is entirely possible that 2 or more users could be uploading files that contain errors at the same time.

GuyHarwood: @GeorgeChatzigiannis  i would use something more unique here - it is entirely possible that 2 or more users could be uploading files that contain errors at the same time.

jon-shipley: @GuyHarwood @GeorgeChatzigiannis - as node often aligns to unix I had assumed there would be a `mktemp` available.  There isn't, but there is a `mkdtemp` which will create a temp directory.  Otherwise there is a package for it: https://www.npmjs.com/package/mktemp

GeorgeChatzigiannis: @GuyHarwood I added school id in the beginning of the file name. 
Let me know if we need further uniqueness

admin/controllers/pupil.js
```diff
@@ -159,7 +160,7 @@ const postAddMultiplePupils = async (req, res, next) => {
         const cvsStr = await writeToString(errorsCsv, { headers: true })
         // Upload csv to Azure
         try {
-          const remoteFilename = `${school._id}-${moment().format('YYYYMMDDHHmmss')}-error.csv`
+          const remoteFilename = `${school._id}_${uuidv4()}_${moment().format('YYYYMMDDHHmmss')}_error.csv`
           const streamLength = 512 * 1000
```
GuyHarwood: 👍

admin/test/features/add_multiple_pupil.feature
```diff
@@ -0,0 +1,30 @@
...
+    Given I am logged in
+    And I am on the add multiple pupil page
+
+  Scenario: Add Multiple Pupil Landing page
+    Then I should see a heading for the Add Multiple pupils
+    And I can see sub heading 'Download example' on Add Multiple Pupil Page
+    And I can see sub heading 'Upload file' on Add Multiple Pupil Page
+    And I can see the Info message on adding Multiple Pupils
+    And I should have the option to choose a csv file for adding multiple pupil
+
```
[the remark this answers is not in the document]
sinhaamit1980: Done

admin/controllers/pupil.js
```diff
@@ -95,27 +82,142 @@ const postAddPupil = async (req, res, next) => {
...
     next(error)
   }
 }
 
+const postAddMultiplePupils = async (req, res, next) => {
+  let school
+  try {
+    school = await School.findOne({_id: req.user.School}).exec()
+    if (!school) {
+      throw new Error(`School [${req.body.school}] not found`)
```
jon-shipley: I don't understand why `req.body.school` is used in the error message but `req.user.School` is used for the actual find?

admin/controllers/pupil.js
```diff
@@ -95,27 +82,142 @@ const postAddPupil = async (req, res, next) => {
...
+  const stream = fs.createReadStream(uploadFile.file)
+  const csvStream = csv()
+    .on('data', (data) => { csvData.push(data) })
+    .on('end', async() => {
+      // Remove error column and headers from data
+      if (csvData.some(p => p[6])) csvData.map((r) => r.splice(6, 1))
+      let headers = csvData.shift(0)
+      // validate each pupil
+      const pupils = []
+      csvData = await Promise.all(csvData.map(async(p) => {
```
jon-shipley: This needs to be re-factored.  There is way too much going on here.

admin/controllers/pupil.js
```diff
@@ -59,33 +65,14 @@ const postAddPupil = async (req, res, next) => {
...
-        combinedValidationError.removeError('dob')
-      }
-      return res.render('school/add-pupil', {
-        school: school.toJSON(),
-        formData: req.body,
-        error: combinedValidationError,
-        breadcrumbs: req.breadcrumbs()
-      })
-    }
+    Object.keys(error.errors).forEach((e) => { error.errors[e] = error.errors[e].replace(/\\/g, '') })
```
jon-shipley: This code seems far too low-level to be in a controller.  Why?

admin/controllers/pupil.js
```diff
@@ -95,27 +82,142 @@ const postAddPupil = async (req, res, next) => {
...
+          const dob = p[4].split('/')
+          const pupilData = Object.assign({
+            'dob-day': dob[1],
+            'dob-month': dob[0],
+            'dob-year': dob[2]
+          }, pupil._doc)
+          await validatePupil(pupil, pupilData)
+        } catch (err) {
+          p[6] = []
+          Object.keys(err.errors).forEach((e) => p[6].push(err.errors[e].replace(/ \\"/g, '')))
```
jon-shipley: Why even have the smart quotes in the error message if you remove them?

admin/controllers/pupil.js
```diff
@@ -95,27 +82,142 @@ const postAddPupil = async (req, res, next) => {
...
+          savedPupils.map((p) => pupilIds.push(p._id))
+          req.flash('info', `${pupils.length} new pupils have been added`)
+        } catch (err) {
+          return next(err)
+        }
+        pupilIds = JSON.stringify(pupilIds)
+        res.redirect(`/school/pupil-register/lastName/true?hl=${pupilIds}`)
+      }
+    })
+  stream.pipe(csvStream)
```
jon-shipley: Well this is a fun function.  Does it work?  Have you got TDD unit tests?

admin/controllers/school.js
```diff
@@ -72,8 +72,9 @@ const getPupils = async (req, res, next) => {
   })
   try {
     req.breadcrumbs(res.locals.pageTitle)
+    if (req.query.hl) req.query.hl = JSON.parse(req.query.hl)
     res.render('school/pupil-register', {
-      highlight: new Set(Array.isArray(req.query.hl) ? req.query.hl : [ req.query.hl ]),
+      highlight: new Set(req.query.hl),
```
jon-shipley: Yes - that's better 👍

admin/lib/errors/pupil.js
```diff
@@ -17,9 +17,9 @@ module.exports.addPupil = {
   'dobNoFuture': 'Date of birth can\'t be in the future',
   'dobRequired': 'Date of birth can\'t be blank',
   'dobInvalidChars': 'Entry must be a number',
-  'dob-day': 'Please check “Day”',
-  'dob-month': 'Please check “Month”',
-  'dob-year': 'Please check “Year”',
+  'dob-day': 'Please check \\“Day\\”',
```
jon-shipley: I suggest we don't need to escape the quotes by default - the string is fine.  If you want to write them into csv data then the csv package should be responsible for escaping where needed.

admin/lib/validator/pupil-validator.js
```diff
@@ -129,14 +84,43 @@ module.exports.validate = async function (req) {
...
-
+  // Gender Validation
+  if (!(/^(M|F)$/).test(pupilData['gender'])) {
+    validationError.addError('gender', addPupilErrorMessages.genderRequired)
+  }
+  // UPN Validation
+  const upn = pupilData['upn'].trim().toUpperCase()
+  const expected = upnService.calculateCheckLetter(upn.substring(1))
+  if (expected !== upn[0]) {
+    console.log(`UPN check letter validation failed for [${upn}]: expected [${expected}] but got [${upn[0]}]`)
```
jon-shipley: Can we turn these console.logs off?

admin/controllers/pupil.js
GeorgeChatzigiannis: @jon-shipley as i stated this will be refactored in a separate PR. I already added a comment about it in the description.

admin/controllers/pupil.js
GeorgeChatzigiannis: its wrong, update coming

admin/controllers/pupil.js
GeorgeChatzigiannis: @jon-shipley It won't live in the controller as soon as I refactor it as part of the next PR

admin/controllers/pupil.js
GeorgeChatzigiannis: The double quotes within single quotes causes an issue on Excel. These content within the double quotes will appear as an ascii representation which is obviously unreadable.

admin/controllers/pupil.js
GeorgeChatzigiannis: tests will accompany each service I will have as part of the refactor

admin/lib/validator/pupil-validator.js
GeorgeChatzigiannis: done

admin/lib/errors/pupil.js
GeorgeChatzigiannis: @jon-shipley any suggestion on how to escape double quotes within single quotes w/o escape chars?
That was my original intention but it doesn't seem like it's possible.

admin/controllers/school.js
GeorgeChatzigiannis: 🥇

admin/lib/errors/pupil.js
jon-shipley: Why do you need to escape them?  The error message string is fine without the escapes.

admin/controllers/pupil.js
jon-shipley: Can you provide an example of the output that is causing an issue?  Is it only Excel that has the issue?

admin/controllers/pupil.js
jon-shipley: You will need to unit test the controller as a part of the refactor.

admin/controllers/pupil.js
GeorgeChatzigiannis: So far I only have noticed it on Excel.
'Please check “Year”' will appear as Please check â€œYearâ€

jon-shipley: It should be fine in UTF-8.... however it may be more accessible if you use only ASCII characters.  Does it work if you use normal straight double-quotes rather than curlies?

admin/controllers/pupil.js
GeorgeChatzigiannis: I guess I have to...

admin/test/features/add_multiple_pupil.feature
```diff
@@ -4,26 +4,30 @@ Feature: Add Multiple Pupil
...
     Then I should be taken to the Pupil register page
     And I should see a flash message for the multiple pupil upload
     And I can see the new pupil in the list
     And I delete the Add Multiple Pupil CSV file
 
-  Scenario: User uploading a CSV files with errors
+  Scenario: User uploading a CSV files with errors can see error message and link to download error file
     When I Upload a CSV file with errors to add Multiple Pupil
     And I can see the error message for adding Multiple Pupil
     And I can see link to download Error File
```
activemq: @sinhaamit1980 I meant that all the scenarios have alot going on...Could you maybe try and stick to having only one conjunction (And) and if required try summarise the steps

admin/controllers/pupil.js
```diff
@@ -95,27 +82,142 @@ const postAddPupil = async (req, res, next) => {
   } catch (error) {
     next(error)
   }
-  res.redirect(`/school/pupil-register/lastName/true?hl=${pupil._id}`)
+  const pupilId = JSON.stringify([pupil._id])
+  res.redirect(`/school/pupil-register/lastName/true?hl=${pupilId}`)
```
pris54: As a suggestion, I would make this into a friendly URL (no ?=)

admin/controllers/pupil.js
```diff
@@ -218,7 +320,8 @@ const postEditPupil = async (req, res, next) => {
   }
 
   // pupil saved
-  res.redirect(`/school/pupil-register/lastName/true?hl=${pupil._id}`)
+  const pupilId = JSON.stringify([pupil._id])
+  res.redirect(`/school/pupil-register/lastName/true?hl=${pupilId}`)
```
pris54: Same as previous comment, a user friendly, slash only URL would look better.

admin/controllers/pupil.js
jon-shipley: Setting the highlight via the query string in my work.   I would rather use the qs than session, personally.

GeorgeChatzigiannis: @pris54 we need to pass down a list of people that will display the green indicator.

I am more than happy to hear your suggestions on how to achieve this.

admin/controllers/pupil.js
GeorgeChatzigiannis: check my previous comment

admin/controllers/pupil.js
```diff
@@ -1,11 +1,17 @@
 const moment = require('moment')
+const uuidv4 = require('uuid/v4')
+const fs = require('fs-extra')
+const csv = require('fast-csv')
+const { promisify } = require('bluebird')
+const azure = require('azure-storage')
+const blobService = azure.createBlobService()
```
GuyHarwood: this is likely the cause of the travis build failure.  As soon as you call `azure.createBlobService()` it is looking for the env var `AZURE_STORAGE_CONNECTION_STRING` which is not present.  Mock the blobService out.

PR #209, opened 2017-09-29
Feature/14132 pupils not taking check save

pris54:
Note: Controller to be refactored and made into one or more services in next iteration.
Unit tests are pending.
PR includes stories 14132, 15627


admin/controllers/school.js
```diff
@@ -444,14 +444,47 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+
+  if (req.body.attendanceCode === undefined || req.body.pupil === undefined) {
+    return res.redirect('/school/pupils-not-taking-check/select-pupils')
+  }
+
+  const todayDate = moment(moment.now()).format()
+  let pupils = req.body.pupil
+
+  pupils.forEach(async (id) => {
+    let pupil = await fetchOnePupil(id, req.user.School)
```
jon-shipley: This is fetching each pupil in sequence, and likely to be slower than needed.  Why not fetch them all in parallel?

admin/controllers/school.js
```diff
@@ -444,14 +444,47 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+    if (pupil) {
+      pupil.attendanceCode = {
+        _id: req.body.attendanceCode,
+        dateRecorded: new Date(todayDate),
+        byUser: req.user.UserName
+      }
+      await pupil.save()
```
jon-shipley: Equally adding a save inside the loop is not recommended.  Please change the flow of execution to a) fetch all pupils, b) process the pupils (the loop), and then c) save the pupils.

admin/services/pupil.service.js
```diff
@@ -48,6 +48,22 @@ const pupilService = {
     }
   },
   /**
+   * Fetch one pupil filtered by pupil id and school id
+   * @param pupilId
+   * @param schoolId
+   * @returns {Promise.<*>}
+   */
+  fetchOnePupil: async (pupilId, schoolId) => {
```
jon-shipley: As the pupilId is a primary key, what is the purpose of the schoolId check?

admin/services/pupil.service.js
```diff
@@ -48,6 +48,22 @@ const pupilService = {
...
+    // TODO: Introduce integration tests
+    try {
+      return await Pupil
+      .findOne({'_id': pupilId, 'school': schoolId})
+      .exec()
+    } catch (error) {
+      throw new Error(error)
```
jon-shipley: There is no need for this try/catch block - suggest you remove.

admin/views/school/save-pupils-not-taking-check.ejs
```diff
@@ -0,0 +1,67 @@
...
+    </div>
+    <div class="column-one-half">
+        <%- partial('partials/_name_and_sign_out') %>
+    </div>
+</div>
+
+<div class="grid-row" id="pupils-not-taking-checks" aria-describedby="heading-xlarge">
+
+    <div class="column-two-thirds">
+        <h1 class="heading-xlarge" id="heading-xlarge" tabindex="0">Pupils not taking the check</h1>
```
jon-shipley: Have you a link to any research that suggests adding a tab index is useful for `h` elements?  It is not included on the gov.uk documentation: https://govuk-elements.herokuapp.com/typography/#typography-headings for instance.

admin/controllers/school.js
pris54: @jon-shipley Modified as suggested to insertMany outside of the loop.

admin/controllers/school.js
pris54: Done.

admin/views/school/save-pupils-not-taking-check.ejs
pris54: I've removed it for now but I'll bring this back eventually when we revise accessibility. The purpose is to improve the keyboard navigation.

admin/services/pupil.service.js
pris54: I'd rather keep it for now as it's the same pattern used for all the methods in the service. Maybe best to revise them all at once?

admin/services/pupil.service.js
pris54: IIRC it was discussed that one pupil could be temporarily registered in more than one school (and not knowing yet how we are going to deal with this), the school id is to make sure we display the one for the school in the current session.

admin/controllers/school.js
```diff
@@ -444,14 +444,49 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+  if (req.body.attendanceCode === undefined || req.body.pupil === undefined) {
+    return res.redirect('/school/pupils-not-taking-check/select-pupils')
+  }
+
+  const todayDate = moment(moment.now()).format()
+  let pupils = req.body.pupil
+  let pupilsProcessed
+
+  pupils.forEach(async (id) => {
+    let pupil = fetchOnePupil(id, req.user.School)
```
jon-shipley: We shouldn't fetch individual pupils inside a loop - we can fetch all pupils in one go which will minimise round trips.

admin/controllers/school.js
```diff
@@ -444,14 +444,49 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+    if (pupil) {
+      pupil.attendanceCode = {
+        _id: req.body.attendanceCode,
+        dateRecorded: new Date(todayDate),
+        byUser: req.user.UserName
```
jon-shipley: UserName is not unique - `EmailAddress` probably is however.  We don't have a pattern for this. It may be best to use both Name and EmailAddress fields as `byUserName` `byUserEmail`.

@GuyHarwood @GeorgeChatzigiannis - thoughts?

admin/controllers/school.js
```diff
@@ -444,14 +444,49 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+      }
+      pupilsProcessed.push(pupil)
+    }
+  })
+  await pupilsProcessed.insertMany()
```
jon-shipley: Should there be a check for any errors here?

admin/controllers/school.js
GuyHarwood: @pris54 this will not scale well.  Can we please do bulk operations where possible

admin/controllers/school.js
GuyHarwood: use both.  we need to explore this in more detail.  have added a card to architecture backlog

admin/services/pupil.service.js
GuyHarwood: @pris54 please can you do this change as part of the PR, as it's superfluous, and we will likely forget if its left as-is.

admin/services/pupil.service.js
```diff
@@ -37,16 +37,23 @@ const pupilService = {
...
   },
+  fetchMultiplePupils: async (pupilIds) => {
+    try {
+      return await Pupil
+        .find({'_id': { $in: pupilIds }})
+        .exec()
+    } catch (error) {
+      throw new Error(error)
+    }
+  },
```
GuyHarwood: Much better :)  Now we just need the error wrapping removed

admin/services/pupil.service.js
pris54: This method has been replaced, the new one doesn't use a try/catch block however, I'm not refactoring the remaining methods of the service at this point.

admin/controllers/school.js
pris54: This is in the next PR (story 15595)

admin/controllers/school.js
```diff
@@ -449,14 +450,55 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+  let pupilsToSave = []
+
+  await pupilsData.map(async (pupil) => {
+    if (pupil) {
+      pupil.attendanceCode = {
+        _id: req.body.attendanceCode,
+        dateRecorded: new Date(todayDate),
+        byUser: req.user.UserName
+      }
+      await pupilsToSave.push(pupil)
```
jon-shipley: I don't think you need to await on an `Array.push()`.  This whole block should be synchronous.

admin/controllers/school.js
```diff
@@ -449,14 +450,55 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+  if (req.body.attendanceCode === undefined || req.body.pupil === undefined) {
+    return res.redirect('/school/pupils-not-taking-check/select-pupils')
+  }
+
+  const todayDate = moment(moment.now()).format()
+  const pupilsData = await fetchMultiplePupils(Object.values(req.body.pupil))
+  let pupilsToSave = []
+
+  await pupilsData.map(async (pupil) => {
+    if (pupil) {
```
jon-shipley: Do we need this `if` condition?  Won't the result of `fetchMultiplePupils` either be an array of pupils or null?

admin/controllers/school.js
```diff
@@ -449,14 +450,55 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
...
+      pupil.attendanceCode = {
+        _id: req.body.attendanceCode,
+        dateRecorded: new Date(todayDate),
+        byUser: req.user.UserName
+      }
+      await pupilsToSave.push(pupil)
+    }
+  })
+
+  await Pupil.create(pupilsToSave, function (err, pupil) {
```
jon-shipley: This looks weird - async _and_ a callback.  In any case it seems `Model.create([], cb)` is not a bulk update, mongoose will turn this into a loop and save each object individually.

admin/models/pupil.js
```diff
@@ -58,14 +58,16 @@ const Pupil = new Schema({
...
+    required: false,
+    ref: 'AttendanceCode'
   }
 }, {
   timestamps: true
 })
 
 /**
- * Validation / sanitisation
+ * Validation / sanitation
```
jon-shipley: I think sanitisation is correct.  Sanitation relates to public health.

admin/services/pupil.service.js
GuyHarwood: @pris54 its a 5 minute job, and i would like to see consistency across our components.  Please refactor them all.

admin/controllers/school.js
pris54: [model].create works the same as [model].insertMany, the second is more recommended for bulk inserts except that it won't deal with updates. This is the best method I found for a bulk update, please show me an example of a better one and I'll refactor.

From Mongoose website:
http://mongoosejs.com/docs/api.html#model_Model.create
http://mongoosejs.com/docs/api.html#model_Model.insertMany

admin/services/pupil.service.js
```diff
@@ -33,34 +33,31 @@ const pupilService = {
...
     // TODO: Introduce integration tests
-    try {
-      return await Answer.findOne({
-        pupil: mongoose.Types.ObjectId(id),
-        result: {$exists: true}
-      }).sort({createdAt: -1}).exec()
-    } catch (error) {
-      throw new Error(error)
-    }
+    const answers = await Answer.findOne({
```
GeorgeChatzigiannis: @pris54 don't you want to catch the mongoose error though if it occurs?

jon-shipley: @GeorgeChatzigiannis - the caller can catch any errors

GuyHarwood:
@pris54 please delete the branch if it is no longer required

PR #211, opened 2017-10-04
Prevent refresh breaking check

jon-shipley:
# Preparatory work to getting refresh to do the right thing

* fix UserService to look for the correct token to determine if the user is logged in
* update the question service and warmup question service to re-initialise from storage if storage is available
* store the check state


GeorgeChatzigiannis:
@jon-shipley let me know when it's RFR, I updated your branch to master

pupil-spa/src/app/services/user/user.service.ts
```diff
@@ -16,7 +15,7 @@ export class UserService {
   data: any = {};
 
   constructor(private http: Http, private storageService: StorageService) {
-    this.loggedIn = !!this.storageService.getItem(sessionDataKey);
+    this.loggedIn = !!this.storageService.getItem(accessTokenKey);
```
GuyHarwood: @jon-shipley bear in mind the access token is going to be HttpOnly soon.

jon-shipley: Yes, that's fine.  At which point we can delete this.

PR #216, opened 2017-10-05
Feature/15595 pupils not taking check render saved pupils

pris54:
NOTE: PR #209 needs to be approved and merged first.


admin/assets/stylesheets/elements/_flash-messages.scss
```diff
@@ -1,7 +1,17 @@
+// Flash message
 .info-message {
   border: 5px solid $green;
   padding: 1em;
-  margin-top: 45px;
+  margin-top: 40px;
```
pris54: This is to align with the top border of the sidebar

admin/controllers/school.js
```diff
@@ -458,6 +471,13 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
   })
 }
 
+/**
```
GuyHarwood: where are the unit tests for this controller?

admin/controllers/school.js
```diff
@@ -467,8 +487,13 @@ const savePupilNotTakingCheck = async (req, res, next) => {
...
 
   const todayDate = moment(moment.now()).format()
-  const pupilsData = await fetchMultiplePupils(Object.values(req.body.pupil))
+  const postedPupils = req.body.pupil
+  const pupilsData = await fetchMultiplePupils(Object.values(postedPupils))
+
   let pupilsToSave = []
+  let pupilsList
+  let attendanceCodes
+  let flashMessage = 'Reasons updated'
```
GeorgeChatzigiannis: 'Reason updated' is not needed as value right?
that can just be let flashMessage

admin/controllers/school.js
```diff
@@ -486,19 +511,52 @@ const savePupilNotTakingCheck = async (req, res, next) => {
...
-  let pupilsList = []
+  // Get attendance code index
+  try {
+    attendanceCodes = await AttendanceCode.getAttendanceCodes().exec()
+  } catch (error) {
+    console.log('ERROR getting attendance codes', error)
+    return next(error)
+  }
+
+  let pupils = await fetchPupilsWithReasons(req.user.School)
```
GeorgeChatzigiannis: that can be a const right?

admin/controllers/school.js
```diff
@@ -486,19 +511,52 @@ const savePupilNotTakingCheck = async (req, res, next) => {
...
   return res.render('school/save-pupils-not-taking-check', {
     breadcrumbs: req.breadcrumbs(),
-    pupilsList
+    pupilsList,
+    messages: req.flash('info')
   })
 }
 
+const removePupilNotTakingCheck = async (req, res, next) => {
+  if (req.params.pupilId === undefined) {
```
GeorgeChatzigiannis: Unless you expect the pupilId to ever be 0,empty array or object or nul this can just be simplified as:
`if (req.params.pupild) {`

admin/controllers/school.js
GeorgeChatzigiannis: In addition to guys comment you need to also test the new service function

pris54: @GuyHarwood @GeorgeChatzigiannis As explained and agreed, merging as it is comes first, while signing off happens I'm already in the process of simplifying controllers, writing services all using a TDD approach - so refactoring and unit tests come in the next PR.

PR #217, opened 2017-10-06
Feature/15858 submit multiple validation rules

GeorgeChatzigiannis:


admin/spec/service/data-access/azure-file.data.service.spec.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
```
GuyHarwood: what are you testing in this file?

GeorgeChatzigiannis: In case azure storage connection string is defined in the environment the tests will check the response objects for both download and upload methods.
In the opposite scenario they will check the mock response they receive.

I am questioning though whether these tests are really useful as they expect a container that perhaps would be better to provide through config.

@GuyHarwood thoughts?

GuyHarwood: You are testing a mock when the connection string is not present.
When the connection string is present, you are testing a library maintained by a 3rd party

admin/services/data-access/insert-pupils.data.service.js
```diff
@@ -0,0 +1,10 @@
+const Pupil = require('../../models/pupil')
+
+module.exports.insertMany = async (pupils) => {
+  const savedPupils = await Pupil.insertMany(pupils, (err) => {
+    if (err) {
+      throw new Error(err)
```
jon-shipley: Why re-throw the error in an async function?  It will be rejected automatically in the await/async promise implementation.   Just leave it out, the caller still has to catch errors all the same.

admin/services/pupil-upload.service.js
```diff
@@ -0,0 +1,71 @@
...
+  let stream
+  const pr = await new Promise((resolve) => {
+    let csvData = []
+    stream = fs.createReadStream(uploadFile.file)
+    csv.fromStream(stream)
+      .on('data', (data) => {
+        csvData.push(data)
+      })
+      .on('end', async () => {
+        // Remove error column and headers from data and validate rows
```
jon-shipley: This entire block is rather long, so it would be better if it was extracted out into another method.

admin/services/pupil-upload.service.js
```diff
@@ -0,0 +1,71 @@
...
+          return resolve({
+            csvErrorFile: csvBlobFile.name,
+            hasValidationError: true
+          })
+        } else {
+          // Save pupils if validation is successful
+          let pupilIds = []
+          let savedPupils
+          try {
+            savedPupils = await insertMany(pupils)
```
jon-shipley: I would prefer this to read `pupilDataService.insertMany(pupils)`  Doesn't that make things clearer to the reader?

admin/spec/service/data-access/azure-file.data.service.spec.js
```diff
@@ -0,0 +1,25 @@
...
+      'John,Lawrence,Smith,L822200014001,5/22/1005,M,"Please check ""Year"", More than 1 pupil record with same UPN"\n' +
+      'Maria,Stella,Brown,A8222000140021,7/15/2005,F,UPN invalid (character 13 not a recognised value)'
+    const blobFile = await azureUploadFile('csvuploads', remoteFilename, csvStr, streamLength)
+    expect(blobFile).toBeDefined()
+    expect(blobFile.name).toBeDefined()
+  })
+
+  it('will download a file from azure blob storage and expect it to have content', async () => {
+    const csvErrorFile = 'error.csv'
+    const blobFile = await azureDownloadFile('csvuploads', csvErrorFile)
```
jon-shipley: You do not appear to be mocking out the azure-file.data.service dependencies.  You should, and then you can test the internal promise resolve and reject paths.

GeorgeChatzigiannis: @jon-shipley I believe these tests are redundant as all they do is wrap the library.
I could potentiall only create a stub to check if the library function has been called but that's only gonna prove that the promise is working as intended.

admin/services/pupil-upload.service.js
GeorgeChatzigiannis: @jon-shipley I changed that to be a single pupil data service to cover both insertMany and create pupil mongoose entity.
Any other future Pupil(mongoose model) service method should be introduced in here

admin/controllers/pupil.js
```diff
@@ -119,91 +109,27 @@ const postAddMultiplePupils = async (req, res, next) => {
...
-    })
-  stream.pipe(csvStream)
+  const uploadFile = req.files && req.files.csvTemplateFile
+  const fileErrors = await fileValidator.validate(uploadFile)
+  if (fileErrors.hasError()) {
+    res.hasError = true
+    res.fileErrors = fileErrors
+    return getAddMultiplePupils(req, res, next)
+  }
+  const csvUploadResult = await pupilUploadService.upload(school, uploadFile, req, res, next)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: they were not used at all and I forgot to remove them.
Updated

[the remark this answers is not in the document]
GeorgeChatzigiannis: Wrapped in try/catch

admin/controllers/pupil.js
```diff
@@ -212,14 +138,7 @@ const getAddMultiplePupilsCSVTemplate = async (req, res) => {
...
 const getErrorCSVFile = async (req, res) => {
-  const blobFile = await new Promise((resolve, reject) => {
-    blobService.getBlobToText('csvuploads', req.session.csvErrorFile,
-      (error, result) => {
-        if (error) reject(error)
-        else return resolve(result)
-      }
-    )
-  })
+  const blobFile = await azureDownloadFile('csvuploads', req.session.csvErrorFile)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: 💯

admin/controllers/pupil.js
```diff
@@ -119,91 +109,27 @@ const postAddMultiplePupils = async (req, res, next) => {
...
+  }
+  const csvUploadResult = await pupilUploadService.upload(school, uploadFile, req, res, next)
+  // upload error
+  if (csvUploadResult.error) next(csvUploadResult.error)
+  // render with errors
+  if (csvUploadResult.hasValidationError) {
+    req.session.csvErrorFile = csvUploadResult.csvErrorFile
+    res.hasError = csvUploadResult.hasValidationError
+    res.fileErrors = csvUploadResult.fileErrors
+    getAddMultiplePupils(req, res, next)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: correct.
Updated

admin/services/pupil-upload.service.js
```diff
@@ -0,0 +1,49 @@
...
+      hasValidationError: true
+    })
+  } else {
+    // Save pupils if validation is successful
+    let pupilIds = []
+    let savedPupils
+    try {
+      savedPupils = await pupilDataService.insertMany(pupils)
+    } catch (error) {
+      return (error)
```
jon-shipley: How does the caller differentiate between an error and the actual result?

admin/services/data-access/check-unique-upn.data.service.js
```diff
@@ -0,0 +1,6 @@
+const Pupil = require('../../models/pupil')
+
+module.exports.isUnique = async (upn, id) => {
```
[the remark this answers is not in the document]
jon-shipley: There is now no need for this file as you can use the more general pupil-data.service.js to get the same result.

admin/controllers/pupil.js
```diff
@@ -119,107 +109,41 @@ const postAddMultiplePupils = async (req, res, next) => {
...
+  } else {
+    req.flash('info', `${csvUploadResult.pupils && csvUploadResult.pupils.length} new pupils have been added`)
+    const ids = JSON.stringify(csvUploadResult.pupilIds)
+    res.redirect(`/school/pupil-register/lastName/true?hl=${ids}`)
+  }
 }
 
 const getAddMultiplePupilsCSVTemplate = async (req, res) => {
-  const file = 'assets/csv/multiple_pupils_template.csv'
+  const file = 'assets/csv/MTC Pupil details template - Sheet 1.csv'
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: changed

admin/services/data-access/check-unique-upn.data.service.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: this service is now removed as correctly pointed.

admin/services/data-access/check-unique-upn.data.service.js
```diff
@@ -0,0 +1,6 @@
...
+  const pupil = await Pupil.findOne({ upn: upn }).exec()
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: indeed, the service is removed though as it's no longer needed

admin/services/pupil-upload.service.js
GeorgeChatzigiannis: the caller should catch the exception coming from mongoose so that try/catch block is not needed.
I am removing it.

GeorgeChatzigiannis: Since the caller (pupil controller) is listening for errors I suppose that try/catch is redundant.
The only case that is not captured is when no pupils are saved in which case a separate error must be thrown.

admin/test/features/step_definitions/add_multiple_pupil_validation_steps.rb
```diff
@@ -0,0 +1,249 @@
...
+  expect(@hashed_data[0][:errors].to_s.include?("Date of birth can't be in the future")).to be_truthy, "Expected: 'Date of birth can't be in the future'... But Got: #{@hashed_data[0][:errors].to_s}"
+end
+
+And(/^I can see the validation error for duplicate UPN from attendance register for multiple pupil upload$/) do
+  expect(@hashed_data[0][:errors].to_s.include?('UPN is a duplicate of a pupil already in your register')).to be_truthy, "Expected: 'UPN is a duplicate of a pupil already in your register'... But Got: #{@hashed_data[0][:errors].to_s}"
+end
+
+# And(/^I can see the validation error for duplicate UPN from within the file for multiple pupil upload$/) do
+#   expect(@hashed_data[0][:errors].to_s.include?('More than 1 pupil record with same UPN')).to be_truthy, "Expected: 'More than 1 pupil record with same UPN'... But Got: #{@hashed_data[0][:errors].to_s}"
+# end
```
activemq: @sinhaamit1980 Remove any commented code if its not needed

sinhaamit1980: Done

admin/controllers/school.js
```diff
@@ -54,7 +54,7 @@ const getPupils = async (req, res, next) => {
   const { pupils } = await fetchPupilsData(req.user.School)
   let pupilsFormatted = await Promise.all(pupils.map(async (p) => {
     const { foreName, lastName, _id } = p
-    const dob = moment(p.dob).format('DD/MM/YYYY')
+    const dob = moment(p.dob).format('DD MMM YYYY')
```
jon-shipley: There is a date service for almost this format.  Why use 'DD' rather than 'D'? The GDS standard is https://www.gov.uk/service-manual/design/dates which states no leading zero.

Suggest you use the DateService to prevent a proliferation of date formats.

GeorgeChatzigiannis: @jon-shipley updated

PR #218, opened 2017-10-06
Feature/15596 pupils not taking check remove

pris54:
Story 15596 ready for review.
I'm adding the latest added story for this feature: 16512
This PR depends on PR #216


admin/services/pupil.service.js
```diff
@@ -25,6 +25,19 @@ const pupilService = {
     }
   },
   /**
+   * Fetch one pupil filtered by pupil id and school id
+   * @param pupilId
+   * @param schoolId
+   * @returns {Promise.<*>}
+   */
+  fetchOnePupil: async (pupilId, schoolId) => {
```
jon-shipley: We should discuss naming at some point.  I prefer to see the service in the call e.g.
`pupilService.fetchOne(id)`

But I would guess from this name that you prefer 
`fetchOnePupil` which I think is less clear as it hides the service.

pris54: @jon-shipley I prefer your suggestion, thanks. I'll have a refactoring PR next week with this and other improvements.

jon-shipley: This should really be moved to pupil.data.service module.

pris54: @jon-shipley  done in the refactoring branch

PR #219, opened 2017-10-09
Feature/16402 handle refresh during check

jon-shipley:
Handle page refreshes during the check by reading and initialising current state

* update inputService to save to storage immediately rather than on flush
* refreshing on the warmup or real questions cause the user to be moved to the next state, other views are allowed to refresh


pupil-spa/src/app/check/check.component.spec.ts
```diff
@@ -236,4 +248,140 @@ describe('CheckComponent', () => {
...
+      // set up test spy on Audit Service
+      component.refreshDetected()
+      // check the spy was called
+      expect(auditService.addEntry).toHaveBeenCalledTimes(1)
+      expect(auditEntryInserted instanceof RefreshDetected).toBeTruthy()
+    })
+
+    it('moves to the next state if a real test question was asked', () => {
+      // find the state for the first warmup question
+      let q1 = component['allowedStates'].indexOf('Q1');
```
GeorgeChatzigiannis: const

pupil-spa/src/app/check/check.component.spec.ts
```diff
@@ -236,4 +248,140 @@ describe('CheckComponent', () => {
...
+      // exercise the code
+      component.ngOnInit()
+
+      // test
+      expect(component['state']).toBe(q1 + 1)
+    })
+
+    it('state stays the same if a page refresh happens on a warmup loading screen', () => {
+      // find the state for the first warmup question
+      let state = component['allowedStates'].indexOf('LW1');
```
GeorgeChatzigiannis: state is not modified (const)

pupil-spa/src/app/check/check.component.ts
```diff
@@ -27,6 +27,14 @@ export class CheckComponent implements OnInit {
   public allowedStates: Array<string> = [];
   private totalNumberOfQuestions: number;
   public static readonly checkStateKey = 'checkstate';
+  private static warmupIntroRe = /^warmup-intro$/;
```
GeorgeChatzigiannis: Should we have a statics file for these?
Just so to keep purely the functional part for the service here.

jon-shipley: The state is internal to the component, so it makes sense to keep it together to me.

GeorgeChatzigiannis:
is this stilll WIP as stated in the description or is it 100% rdy for review?

jon-shipley:
It's WIP still

pupil-spa/src/app/check/check.component.ts
```diff
@@ -234,4 +261,81 @@ export class CheckComponent implements OnInit {
...
+    this.auditService.addEntry(new RefreshDetected());
+
+    // Lets say that handling reloads during the check should always show the current screen
+    // in which case handling the reload whilst a question was being shown is a special case.
+    if (CheckComponent.questionRe.test(stateDesc)) {
+      // the page was reloaded when a question was shown
+      // Store the answer as the empty string (as there was no input)
+      // we need to initialise the current question, with the one from the current state
+      const matches = CheckComponent.questionRe.exec(stateDesc);
+      const questionNum = parseInt(matches[ 1 ], 10);
```
GeorgeChatzigiannis: isn't radix (second arg) defaulting to 10 anyway?

jon-shipley: The recommendation is always use it, so I do :)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

pupil-spa/src/app/check/check.component.ts
```diff
@@ -80,10 +84,29 @@ export class CheckComponent implements OnInit {
...
-    this.state = 0;
-    this.isWarmUp = true;
-    this.viewState = 'warmup-intro';
-    this.totalNumberOfQuestions = this.warmupQuestionService.getNumberOfQuestions();
+    // set up the state
+    if (this.storageService.getItem(CheckComponent.checkStateKey)) {
+      // existing check state detected
+      // assume we are reloading during a check
+      const existingState = this.storageService.getItem(CheckComponent.checkStateKey);
+      if (!this.isValidState(existingState)) {
```
GeorgeChatzigiannis: @jon-shipley not a big fan of nested if as they oppose the principles of clean coding.
How about you approach it with a gateway method?
https://wpshout.com/unconditionally-refactoring-nested-statements-cleaner-code/

GeorgeChatzigiannis: if (!this.storageService.getItem(CheckComponent.checkStateKey)) {
..code 
return
}

jon-shipley: @GeorgeChatzigiannis  I have refactored the code block to make it clearer.

PR #221, opened 2017-10-11
Feature/pupils not taking check refactoring

pris54:
Refactoring and unit tests related to feature 14131 "Pupils not taking the check"


GuyHarwood:
@pris54 instead of just adding 'WIP' to your commits, can you write a brief description of the changes you made.  It's really helpful when in situations where we might need to comb back through commits, as it tells a meaningful story.  A wise man once said...

> "imagine that you are handing over your repo to a psychopath"

pris54:
@GuyHarwood HAHAHA, OK, I will, I'm lazy sometimes.

admin/spec/service/pupils-not-taking-check.service.spec.js
```diff
@@ -0,0 +1,130 @@
...
+      service = setupService(pupilsWithReasonsMock)
+    })
+
+    it('should return a list of pupils with reasons (happy path)', async (done) => {
+      try {
+        const pupilsWithReasons = await service.fetchPupilsWithReasons(9991001)
+        expect(pupilsWithReasons).toEqual(pupilsWithReasonsMock)
+        done()
+      } catch (error) {
+        done('Error found while testing fetchPupilsWithReasons')
```
jon-shipley: This style no longer works with jasmine.  Instead you have to do:
`expect(true).toBe(false)`
if you want to guard against it throwing.

admin/services/pupils-not-taking-check.service.js
```diff
@@ -0,0 +1,102 @@
...
+/** @namespace */
+
+const pupilsNotTakingCheckService = {
+
+  /**
+   * Fetch pupils with reasons to not take the check.
+   * @param schoolId
+   * @returns {Promise.<*>}
+   */
+  fetchPupilsWithReasons: async (schoolId) => {
```
jon-shipley: This should really be in the data service layer

admin/services/pupils-not-taking-check.service.js
```diff
@@ -0,0 +1,102 @@
...
+    .sort('lastName')
+    if (pupilsWithReasons.length > 0) {
+      return pupilsWithReasons
+    }
+  },
+  /**
+   * Sort columns by last name asc/desc.
+   * @param sortField
+   * @param sortDirection
+   * @returns {Promise.<{htmlSortDirection: Array, arrowSortDirection: Array}>}
```
jon-shipley: I don't see a Promise being returned

pris54: Indeed, that's wrong. Fixed in next commit.

admin/spec/service/pupils-not-taking-check.service.spec.js
pris54: Fixed in next commit.

PR #223, opened 2017-10-12
Fix failing docker build

jon-shipley:
Blind fix for failing build on azure

### DIFF
--- a/pupil-spa/Dockerfile
+++ b/pupil-spa/Dockerfile
@@ -14,6 +14,7 @@ WORKDIR /ng-app
 
 COPY . .
 
+RUN yarn run postinstall
 RUN yarn run sass
 ARG api_url=http://localhost:3001
 ENV API_URL=$api_url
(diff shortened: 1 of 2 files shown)


pupil-spa/Dockerfile
```diff
@@ -14,6 +14,7 @@ WORKDIR /ng-app
 
 COPY . .
 
+RUN yarn run postinstall
```
GuyHarwood: £1 says this isn't needed :)

jon-shipley: You owe me £1 😆

GuyHarwood: Goddamit!

Question is - why is post install not firing automatically?

jon-shipley: I reckon it does but it then copies as bunch of files into another directory `ng-app`.  We could clean that up by only running install once, in the new directory.

PR #224, opened 2017-10-12
Remove 'withdrawn', add new reasons. Edit to lower case.

pris54:


jon-shipley:
@pris54 - if the PR has not been merged yet (and therefore not deployed) it's okay to edit the migration.

pris54:
@jon-shipley I've updated the approach, would you like to review again?
This is how it works now:
1) Attendance code is unset from pupils documents (a sort of reset)
2) Attendance code collection is dropped and re-created with the updated list of reasons, including a sorting value to be used in the future (it's an AC that's been postponed for now)

admin/data/migrations/migrations/20171012121301-add-new-attendance-codes.js
```diff
@@ -0,0 +1,53 @@
...
+module.exports = {
+  up (db, next) {
+    mongoose.connect(config.mongodb.url, async error => {
+      if (error) {
+        next(new Error('Could not connect to mongodb: ' + error.message))
+      }
+
+      try {
+        db.collection('attendancecodes').drop()
+        await createAttendanceCode('Incorrect registration', 1, 0)
```
jon-shipley: To be absolutely on it, the drop() almost certainly takes a callback, so the awaits should sit inside the callback.  Otherwise fine :)

PR #227, opened 2017-10-12
Feature/16612 pupil content changes

jon-shipley:
Copy changes

But..

* api change on the spa client login / questions to return date of birth
* refactored pupil-authentication.service to move data access to data services
* jwtToken createToken() signature change
* add new warmup loading component to SPA as the html text is different for warmup questions
* and finally the copy changes requested
* and updated cucumber tests

### DIFF
--- a/admin/services/data-access/pupil.data.service.js
+++ b/admin/services/data-access/pupil.data.service.js
@@ -1,0 +1,31 @@
+'use strict'
+
+const Pupil = require('../../models/pupil')
+const pupilDataService = {}
+
+/**
+ * Find a single pupil by criteria in `options`
+ * @param options
+ * @return {Promise.<{Object}>}
+ */
+pupilDataService.findOne = async function (options) {
+  const p = await Pupil.findOne(options).populate('school').lean().exec()
+  return p
+}
+
+/**
+ * Update a single object with the new fields from `doc`
+ * @param {string} id Object._id
+ * @param {Object} doc Document to pass to mongo
+ * @return {Promise}
+ */
+pupilDataService.update = async function (id, doc) {
+  return new Promise((resolve, reject) => {
+    Pupil.updateOne({_id: id}, doc, (error) => {
+      if (error) { return reject(error) }
+      resolve(null)
+    })
+  })
+}
+
+module.exports = pupilDataService
--- a/admin/controllers/questions.js
+++ b/admin/controllers/questions.js
@@ -25,10 +25,7 @@ const getQuestions = async (req, res) => {
     return apiResponse.unauthorised(res)
   }
 
-  const pupilData = {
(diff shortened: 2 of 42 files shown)


jon-shipley:
@activemq @sinhaamit1980 - please can you approve the cucumber tests I've altered as per the copy changes requested in 16612

As this deals with the SPA authentication probably best if we get 2 approvals or more.

admin/services/data-access/pupil.data.service.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Update a single object with the new fields from `doc`
+ * @param {string} id Object._id
+ * @param {Object} doc Document to pass to mongo
+ * @return {Promise}
+ */
+pupilDataService.update = async function (id, doc) {
+  return new Promise((resolve, reject) => {
+    Pupil.updateOne({_id: id}, doc, (error) => {
```
GeorgeChatzigiannis: I am surprised this works with a callback for errors.
isn't there a promisified version of this?

GeorgeChatzigiannis: you could also promisify it with bluebird I suppose and it would be the same result(?)

jon-shipley: Not that I know of.

You can put `.exec()` on the end to get a Promise, but then it does not wait for the response from the server.  So not the same thing.

I don't see why you expect it not to work.  Care to elaborate?

GeorgeChatzigiannis: I am expecting it to work don't get me wrong but I was also expecting it to work under a promise just fine and wait for the server.

jon-shipley: The underlying mongo api is callback-only. I suspect `updateOne`  is a minimal wrapper around it.

admin/services/data-access/pupil.data.service.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
```
GeorgeChatzigiannis: oh boy we are gonna have conflicts on this one, gonna be fun!!!

admin/services/data-access/pupil.data.service.js
pris54: Mongoose updateOne returns a query:
http://mongoosejs.com/docs/api.html#model_Model.updateOne

GuyHarwood: the official mongoClient is promisified and much cleaner.  But then we lose the schemas...

jon-shipley: @pris54 - I'm not sure from your comment if there is a better way to handle this or not.  Can you clarify please?

PR #228, opened 2017-10-16
Update loading screen content

jon-shipley:
Update SPA copy based on UX review


pupil-spa/src/app/loading/loading.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="preload" id="js-preload-div">
   <div>
     <p id="js-loading" class="font-large text-center">
-      Question {{question}} out of {{total}}
+      Question {{question}}&hellip;
```
GeorgeChatzigiannis: @jon-shipley what's &hellip;

pupil-spa/src/app/login/login.component.html
```diff
@@ -8,8 +8,8 @@
 
 <div class="grid-row">
     <div class="column-two-thirds">
-        <header class="page-header">
-            <h1 class="heading-xlarge">Multiplication tables check</h1>
+        <header class="page-header" style="margin-top: 25px">
+            <h1 class="heading-xlarge" style="margin-top: 25px">Multiplication tables check</h1>
```
GeorgeChatzigiannis: @jon-shipley I'd rather avoin inline styles.
I don't think they are easily maintainable unless we are using them in common components.

jon-shipley: I usually agree with you.  I could create a new class just for this individual heading, but won't the css size increase? It seems better to acknowledge that this is essentially an override only for this page, and using the style attribute keeps the meaning clear.

GeorgeChatzigiannis: How about that login.component.css?
It seems empty and by default angular creates that for you right?
Personally I find the idea of component based css great.
What are your thoughts on this idea?

jon-shipley: @GeorgeChatzigiannis - obviously its a good idea, but as we aren't using that pattern I don't see that we should start.

GeorgeChatzigiannis: @jon-shipley I don't see us using the inline pattern either though apart from some 
'<br style="clear:both"/>' which are very wrong in the first place.

pris54: @jon-shipley I agree with you that we need to be careful about increasing the css size but inline css is not a good practice. If we write good CSS we would rarely end up with classes that are used only once. If too many of these one times cases are happening then it's time to review our SASS/CSS.

GuyHarwood: @jon-shipley @GeorgeChatzigiannis @pris54 be aware that i am aiming to disallow inline css & JS through content security policy restrictions.  Best we don't add any more going forward.

PR #231, opened 2017-10-18
Added test for Generate pin landing page

sinhaamit1980:


admin/test/features/step_definitions/generate_pupil_pins_steps.rb
```diff
@@ -0,0 +1,15 @@
+
+And(/^I am on the generate pupil pins page$/) do
+  school_landing_page.generate_pupil_pin.click
+  @page = generate_pupil_pins_page
```
activemq: what is the value of saving into a instance variable here? @sinhaamit1980

admin/test/features/step_definitions/generate_pupil_pins_steps.rb
```diff
@@ -0,0 +1,15 @@
...
+end
+
+Then(/^I should see a heading for the Generate Pupil Pins$/) do
+  expect(generate_pupil_pins_page).to have_heading
+end
+
+And(/^I can see the info message for generating the pupil pin$/) do
+  expect(@page.has_content?('Pupils will need a Personal Identification Number (PIN) and school password in order to start the check. These expire daily.')).to be_truthy, "Expected text: 'Pupils will need a Personal Identification Number (PIN) and school password in order to start the check. These expire daily.' ... Not found"
```
activemq: why are we using has content? Can we not identify be trying to locate the element, surely that is quicker than scanning the whole page for the text @sinhaamit1980

admin/test/features/step_definitions/generate_pupil_pins_steps.rb
sinhaamit1980: Done

admin/test/features/step_definitions/generate_pupil_pins_steps.rb
sinhaamit1980: Done

PR #232, opened 2017-10-18
[WIP] feature/15011 - Remove check windows

pris54:
WIP:
1) Modal popup
Next PR: controller simplified into services + unit tests.


admin/controllers/administrator.js
```diff
@@ -328,11 +352,39 @@ const saveCheckWindows = async (req, res, next) => {
...
+
+  try {
+    checkWindow = await fetchCheckWindow(req.params.checkWindowId)
+  } catch (err) {
+    return next(err)
+  }
+
+  if (checkWindow) {
+    if (Date.parse(checkWindow.checkStartDate) < Date.now()) {
+      return res.redirect('/administrator/check-windows/status/validationError')
```
jon-shipley: Usually you would write the flash message to session storage here - which happens automatically.   Then you do not need to to redirect to different routes.

admin/services/check-window.service.js
```diff
@@ -2,13 +2,31 @@
...
+
+  /**
+   * Get check window for the current time.
+   * @returns {Promise.<*>}
+   */
+  getCurrentCheckWindow: async () => {
     const now = new Date()
     const checkWindow = await CheckWindow.findOne({startDate: {$lte: now}, endDate: {$gte: now}}).exec()
     if (!checkWindow) {
       throw new Error('No checkwindow is currently available')
```
jon-shipley: I don't see a need to throw an error here.  Isn't returning null sufficient?  I don't think we can predict what the caller expects here.  Conversely, by throwing here the caller must always use a try / catch block - that seems sub-optimal to me.

admin/controllers/administrator.js
```diff
@@ -6,6 +6,11 @@ const settingsErrorMessages = require('../lib/errors/settings')
 const settingsValidator = require('../lib/validator/settings-validator')
 const checkWindowValidator = require('../lib/validator/check-window-validator')
 const checkWindowErrorMessages = require('../lib/errors/check-window')
+const {
+  fetchCheckWindow,
```
jon-shipley: I thought we had agreed not to destructure service calls?  It reads more clearly if we leave it in - the service clearly identifies the layer the code is working at.

admin/services/check-window.service.js
```diff
@@ -2,13 +2,31 @@
...
     }
     return checkWindow
+  },
+  /**
+   * @param url
+   * @returns {*}
+   */
+  addForwardSlashToUrl: (url) => {
```
jon-shipley: Seems out of place as a public method.

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,27 @@
...
+const CheckWindow = require('../../models/check-window')
+
+const checkWindowDataService = {
+
+  /**
+   * Fetch check window document by id.
+   * @param id
+   * @returns {Promise.<*>}
+   */
+  fetchCheckWindow: async (id) => {
```
jon-shipley: This is a specialised case of findOne - a generic findOne implementation would be more useful.  Also `.exec()` is missing.

```
async function findOne (criteria) {
   const checkWindow = await CheckWindow.findOne(criteria).exec()
   return checkWindow
}
```

```
myCheckWindow = checkWindowDataService.findOne({id: 1234})
```

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,27 @@
...
+    checkWindow = await CheckWindow.findOne({'_id': id})
+    return checkWindow
+  },
+  /**
+   * Delete check window by id.
+   * @param id
+   * @returns {Promise.<void>}
+   */
+  deleteCheckWindow: async (id) => {
+    await CheckWindow.deleteOne({'_id': id})
```
jon-shipley: A specialised form of delete seems just fine. 👍

jon-shipley:
@pris54 - please make sure the code is up to date with master before requesting a code review.

admin/controllers/administrator.js
```diff
@@ -328,11 +352,39 @@ const saveCheckWindows = async (req, res, next) => {
...
+  }
+
+  try {
+    checkWindow = await fetchCheckWindow(req.params.checkWindowId)
+  } catch (err) {
+    return next(err)
+  }
+
+  if (checkWindow) {
+    if (Date.parse(checkWindow.checkStartDate) < Date.now()) {
```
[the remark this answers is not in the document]
pris54: it comes up as '2017-10-22T00:00:00.000Z', I'm converting to timestamp.

admin/controllers/administrator.js
pris54: Removing is not rendering so I have two options:
1) Build the whole default page again in this POST request (not a great plan)
2) Redirect (not great but less bad)
Does it make sense?

admin/services/check-window.service.js
pris54: OK, I'll remove it

admin/services/check-window.service.js
pris54: In the end I'm not using it, it'll be removed.

admin/services/data-access/check-window.data.service.js
pris54: Generic? You mean directly written in the controller? I would still prefer to have all nicely grouped in services.

GuyHarwood:
@pris54 how come tests and simplification come after this work?  the TDD approach would cover both of these.  Also - deletes.  We need to ask for confirmation on whether they should be soft or hard deletes, as they may well have a preference.  Seek confirmation from PB/BL

admin/controllers/administrator.js
pris54: OK, will modify this.

pris54:
@GuyHarwood tests come together with refactoring, it's not a brand new feature but an existing one with some additions so I'm using the same approach as for 'pupils not taking the check'. Keen to take on a brand new feature to do proper TDD though.

pris54:
@GuyHarwood I'll confirm with BAs re' soft/hard delete

GuyHarwood:
@pris54 OK, thanks.

admin/data/migrations/migrations/20171019123302-add-check-windows.js
```diff
@@ -0,0 +1,50 @@
...
+        adminStartDate: '2017-09-09 11:33:00',
+        deleted: false
+      })
+
+      ch.push({
+        checkWindowName: 'Check Window Test B',
+        checkStartDate: '2017-10-22 12:00:00',
+        checkEndDate: '2017-10-28 12:00:00',
+        adminStartDate: '2017-10-19 11:44:00',
+        deleted: false
```
GuyHarwood: please ensure you update the schemas under `design/schemas` with any new fields

admin/services/data-access/check-window.data.service.js
pris54: @jon this is now a special case for check windows only (see changes).

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Fetch check windows by status, sort by, sort direction and date (current or past).
+   * @param deleted
+   * @param sortBy
+   * @param sortDirection
+   * @param current
+   * @returns {Promise.<void>}
+   */
+  fetchCheckWindows: async (sortBy, sortDirection, deleted, current) => {
+    return new Promise(async (resolve, reject) => {
```
jon-shipley: I don't think there is a need to return a Promise when in an async function.  Returns or thrown errors will be converted to resolve and reject calls automatically.

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,84 @@
...
+      } else {
+        query.checkEndDate = {$lt: currentTimestamp}
+      }
+
+      try {
+        checkWindows = await CheckWindow
+          .find(query)
+          .sort(sort)
+          .exec()
+      } catch (error) {
```
jon-shipley: Once you get rid of the promise you can just let this error escape - it will be converted to a reject() automatically

admin/controllers/administrator.js
```diff
@@ -183,13 +189,44 @@ const getCheckWindows = async (req, res, next) => {
...
+    checkWindows = await checkWindowDataService.fetchPastCheckWindows(sortField, sortDirection)
+    checkWindowsPast = checkWindows.map(cw => {
+      const adminStartDateMo = moment(cw.adminStartDate)
+      const checkStartDateMo = moment(cw.checkStartDate)
+      const checkEndDateMo = moment(cw.checkEndDate)
+
+      return {
+        id: cw._id,
+        checkWindowName: cw.checkWindowName,
+        adminStartDate: adminStartDateMo.format('DD MMM YYYY'),
```
jon-shipley: There is a DateService that offers a format very similar to this, except that the format used by DateService is GDS compliant.

admin/controllers/administrator.js
```diff
@@ -328,11 +365,48 @@ const saveCheckWindows = async (req, res, next) => {
...
+  try {
+    checkWindow = await checkWindowDataService.fetchCheckWindow(req.params.checkWindowId)
+  } catch (err) {
+    return next(err)
+  }
+
+  if (checkWindow) {
+    if (Date.parse(checkWindow.checkStartDate) < Date.now()) {
+      req.flash('error', 'Deleting an active check window is not allowed.')
+      return res.redirect('/administrator/check-windows')
```
jon-shipley: It looks to me like this redirect could isn't needed, if the second one is moved outside of the else condition so it always redirects.

admin/data/migrations/migrations/20171019113308-edit-check-windows-status.js
```diff
@@ -0,0 +1,37 @@
...
+const config = require('../config')
+
+module.exports = {
+  up (db, next) {
+    mongoose.connect(config.mongodb.url, async error => {
+      if (error) {
+        next(new Error('Could not connect to mongodb: ' + error.message))
+      }
+      try {
+        await db.collection('checkwindows').update({}, {$set: {'deleted': false}}, {multi: true})
```
jon-shipley: Booleans like 'deleted' should be called 'isDeleted' instead for clarity

admin/data/migrations/migrations/20171019113308-edit-check-windows-status.js
```diff
@@ -0,0 +1,37 @@
+'use strict'
+
+const mongoose = require('mongoose')
+mongoose.Promise = global.Promise
+const config = require('../config')
+
+module.exports = {
+  up (db, next) {
+    mongoose.connect(config.mongodb.url, async error => {
```
jon-shipley: You don't really needed mongoose here as you are using a mongo-style update.

admin/data/migrations/migrations/20171019113308-edit-check-windows-status.js
```diff
@@ -0,0 +1,37 @@
...
+        mongoose.disconnect(() => next())
+      } catch (error) {
+        console.log('ERROR', error)
+        next(error)
+      }
+    })
+  },
+
+  down (db, next) {
+    mongoose.connect(config.mongodb.url, async error => {
```
jon-shipley: Again, this mongoose.connect is not needed - you can just use the db var directly.  It's not wrong though.

admin/data/migrations/migrations/20171019123302-add-check-windows.js
```diff
@@ -0,0 +1,50 @@
...
+      ch.push({
+        checkWindowName: 'Check Window Test C',
+        checkStartDate: '2017-10-18 12:00:00',
+        checkEndDate: '2017-10-21 12:00:00',
+        adminStartDate: '2017-10-15 11:22:00',
+        deleted: false
+      })
+
+      try {
+        await CheckWindow.insertMany(ch)
```
jon-shipley: I had to read from line 15 to here to understand what `ch` meant.  `cw` would be a more obvious choice for me.

admin/models/check-window.js
```diff
@@ -21,6 +21,11 @@ const CheckWindow = new Schema({
     type: Date,
     required: true
   },
+  deleted: {
```
jon-shipley: All other booleans are named isX - please change to isDeleted

admin/services/data-access/check-window.data.service.js
pris54: OK. Then perhaps we should modify this pattern everywhere else it's been used too.

admin/controllers/administrator.js
pris54: 'DD MMM YYYY' is what UX wants...

admin/controllers/administrator.js
pris54: Done

admin/data/migrations/migrations/20171019113308-edit-check-windows-status.js
pris54: We have a mix of mongoose and plain mongo in the migrations, I rather stay with this version of the script please, it works just fine.

admin/data/migrations/migrations/20171019113308-edit-check-windows-status.js
pris54: Please, let's keep it or refactor all the migrations. From now on if you don't want mongoose in the migrations scripts I won't use it but please don't make me rewrite this.

PR #233, opened 2017-10-19
Feature/16811 update the psychometricians report

jon-shipley:
Prep work for the Psychometrician's report

* add marking service
* model updates
* data access services


admin/services/marking.service.js
```diff
@@ -0,0 +1,57 @@
...
+  const results = {
+    marks: 0,
+    maxMarks: completedCheck.data.answers.length,
+    processedAt: new Date()
+  }
+
+  for (let answer of completedCheck.data.answers) {
+    if (answer.factor1 * answer.factor2 === parseInt(answer.answer, 10)) {
+      answer.isCorrect = true
+      results.marks += 1
```
GuyHarwood: Does results.marks only count correct marks?

jon-shipley: @GuyHarwood - yes it is count of all marks applied

PR #236, opened 2017-10-20
Feature/check window refactoring [refactoring]

pris54:
Depends on PR 232 - 
Simplification of controllers. Logic moved to services. Implementation of sorting service for both features 'check windows' and 'pupils not taking the check'. Initial unit tests.


admin/controllers/administrator.js
```diff
@@ -328,11 +302,48 @@ const saveCheckWindows = async (req, res, next) => {
...
+  try {
+    checkWindow = await checkWindowDataService.fetchCheckWindow(req.params.checkWindowId)
+  } catch (err) {
+    return next(err)
+  }
+
+  if (checkWindow) {
+    if (Date.parse(checkWindow.checkStartDate) < Date.now()) {
+      req.flash('error', 'Deleting an active check window is not allowed.')
+      return res.redirect('/administrator/check-windows')
```
jon-shipley: The redirect in the if and the else parts is the same, so it can be extracted and moved to run at the end?

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,83 @@
...
+const checkWindowDataService = {
+
+  /**
+   * Fetch check window document by id.
+   * @param id
+   * @returns {Promise.<*>}
+   */
+  fetchCheckWindow: async (id) => {
+    let checkWindow
+    checkWindow = await CheckWindow.findOne({'_id': id, 'deleted': false}).exec()
```
jon-shipley: We should actually just do

```
return CheckWindow.findOne({'_id': id, 'deleted': false}).exec()
```

Anything calling an async function will will await'ing anyway: https://eslint.org/docs/rules/no-return-await  This also applies to other methods here.

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Fetch check windows by status, sort by, sort direction and date (current or past).
+   * @param deleted
+   * @param sortBy
+   * @param sortDirection
+   * @param current
+   * @returns {Promise.<void>}
+   */
+  fetchCheckWindows: async (sortBy, sortDirection, deleted, current) => {
+    return new Promise(async (resolve, reject) => {
```
jon-shipley: There is no need for this promise - please remove it.

admin/services/data-access/check-window.data.service.js
```diff
@@ -0,0 +1,83 @@
...
+
+      sort[sortBy] = sortDirection
+      query.deleted = !deleted ? false : deleted
+      if (current === true) {
+        query.checkEndDate = {$gte: currentTimestamp}
+      } else {
+        query.checkEndDate = {$lt: currentTimestamp}
+      }
+
+      try {
```
jon-shipley: This try / catch block is not doing anything.  Please remove it.

admin/models/check-window.js
```diff
@@ -21,6 +21,11 @@ const CheckWindow = new Schema({
     type: Date,
     required: true
   },
+  deleted: {
```
jon-shipley: Please name this isDeleted to be consistent with other booleans.

admin/services/check-window.service.js
```diff
@@ -1,14 +1,63 @@
...
     }
     return checkWindow
+  },
+  /**
+   * Format check period (start and end dates).
+   * @param startDate
+   * @param endDate
+   * @returns {string}
+   */
+  formatCheckPeriod: (startDate, endDate) => {
```
jon-shipley: Date formatting should go in the DateService.

pris54: @jon-shipley Did you notice how specific this formatting is? Do you still want it as a generic solution in the date service?

admin/services/data-access/check-window.data.service.js
pris54: @jon-shipley Done, however, do you realise we should modify this syntax all across the site? It's a common pattern we've been using so far.

admin/controllers/administrator.js
pris54: Done.

admin/controllers/administrator.js
```diff
@@ -6,6 +6,10 @@ const settingsErrorMessages = require('../lib/errors/settings')
 const settingsValidator = require('../lib/validator/settings-validator')
 const checkWindowValidator = require('../lib/validator/check-window-validator')
 const checkWindowErrorMessages = require('../lib/errors/check-window')
+const checkWindowService = require('../services/check-window.service')
+const dateService = require('../services/date.service')
+const checkWindowDataService = require('../services/data-access/check-window.data.service')
+const sortingService = require('../services/sorting-ui.service')
```
GeorgeChatzigiannis: that should be then:
`const sortingUIService = require('../services/sorting-ui.service')`

admin/services/check-window.service.js
pris54: @jon-shipley This custom date formatting is in the service.

admin/controllers/school.js
```diff
@@ -16,18 +15,14 @@ const {
...
-  fetchPupilsWithReasons,
-  getAttendanceCodes } = require('../services/data-access/pupils-not-taking-check.data.service')
+const pupilsNotTackingCheckService = require('../services/pupils-not-taking-check.service')
+const pupilsNotTackingCheckDataService = require('../services/data-access/pupils-not-taking-check.data.service')
 const dateService = require('../services/date.service')
 const pupilDataService = require('../services/data-access/pupil.data.service')
 const schoolDataService = require('../services/data-access/school.data.service')
 const generatePinsService = require('../services/generate-pins.service')
 const { sortRecords } = require('../utils')
+const sortingService = require('../services/sorting-ui.service')
```
GeorgeChatzigiannis: sortingUIService here too

admin/services/check-window.service.js
jon-shipley: Okay

PR #237, opened 2017-10-23
Feature/16921 sticky banner changes

pris54:
jQuery is not needed anymore for a bottom fixed positioning. Please note that more arguments are needed for the partial view.


pris54:
@sinhaamit1980 @activemq Could either of you look into these tests? Thanks!

PR #240, opened 2017-10-24
Feature/16343 generate pins list sorting & ui changes

GeorgeChatzigiannis:


admin/controllers/school.js
```diff
@@ -215,10 +214,17 @@ const getGeneratePinsList = async (req, res, next) => {
   } catch (error) {
     return next(error)
   }
-  const pupils = await generatePinsService.getPupils(school._id)
+  const sortingOptions = [ { 'key': 'lastName', 'value': 'asc' } ]
+  let sortField = req.params.sortField === undefined ? 'lastName' : req.params.sortField
+  const sortDirection = req.params.sortDirection === undefined ? 'asc' : req.params.sortDirection
+  const { htmlSortDirection, arrowSortDirection } = sortingUIService.getAttributes(sortingOptions, sortField, sortDirection)
+  console.log(htmlSortDirection, arrowSortDirection)
```
jon-shipley: debug?

admin/services/data-access/pupil.data.service.js
```diff
@@ -59,4 +59,23 @@ pupilDataService.update = async function (id, doc) {
...
+ * @param schoolId
+ * @param sortingField
+ * @param sortingDirection
+ * @returns {Array}
+ */
+pupilDataService.getSortedPupils = async (schoolId, sortingField, sortingDirection) => {
+  // TODO: Introduce integration tests
+  const sort = {}
+  sort[sortingField] = sortingDirection
+  const pupils = await Pupil
```
jon-shipley: As this is an async service, which means the caller *must* await on the service for the resolve(), you can just write it like this:

```
return Pupil
.find()
.lean()
.exec()
```

jon-shipley: There shorter form would be `return await ...` but you can lose the await as the caller is await'ing anyway.

admin/controllers/administrator.js
```diff
@@ -9,7 +9,7 @@ const checkWindowErrorMessages = require('../lib/errors/check-window')
 const checkWindowService = require('../services/check-window.service')
 const dateService = require('../services/date.service')
 const checkWindowDataService = require('../services/data-access/check-window.data.service')
-const sortingUiService = require('../services/sorting-ui.service')
+const sortingUIService = require('../services/sorting-ui.service')
```
pris54: @GeorgeChatzigiannis I believe the correct camel case is the as it was before 'sortingUiService' - I know it's not intuitive but that's the correct syntax. 

From Wikipedia:
"Initialisms of three or more letters are CamelCase instead of upper case (e.g., parseDbmXmlFromIPAddress instead of parseDBMXMLFromIPAddress")
https://en.wikipedia.org/wiki/Naming_convention_(programming)#Challenges

GuyHarwood: i would say sortingUiService is more readable.  can we just approve the damn PR already 😄

GeorgeChatzigiannis: changed it to sortingAttributesService

admin/controllers/school.js
GeorgeChatzigiannis: removed

admin/services/data-access/pupil.data.service.js
GeorgeChatzigiannis: changed 👍

admin/test/features/generate_pupil_pins.feature
```diff
@@ -33,3 +33,10 @@ Feature: Generate Pupil PINs
     And I am on the generate pupil pins page
     When I click Generate PINs button
     Then I cannot see this pupil in the list of Pupil on Generate Pin list page
+
+  Scenario: Sorting Pupil list on Generate Pins page
+    Given I am logged in
+    And I navigate to generate pupil pins page
+    And I click Generate PINs button
+    When I click on the Pupil heading
+    Then I should see on the Generate Pin Pupil List page that all pupils are displayed in descending order of lastname
```
activemq: This is a extremely long step name, there is an existing step for this already - 
'Then I should see a list of pupils sorted by surname on Generate Pins List Page'

activemq: Scenarios again have quite a few conjunctions please try summarise steps and make the scenarios less imperative.

sinhaamit1980: Done

PR #244, opened 2017-10-26
Feature/17000 pin submission new ui

GeorgeChatzigiannis:
Pupil pin submission based on the old implementation.

### DIFF
--- a/admin/controllers/school.js
+++ b/admin/controllers/school.js
@@ -227,6 +227,29 @@ const getGeneratePinsList = async (req, res, next) => {
   })
 }
 
+const postGeneratePins = async (req, res, next) => {
+  const { pupil: pupilsList } = req.body
+  if (!pupilsList) {
+    return res.redirect('/school/generate-pins-list')
+  }
+  let submittedPupils
+  try {
+    submittedPupils = await generatePinsService.generatePupilPins(pupilsList)
+    await pupilDataService.saveMultiple(submittedPupils)
+  } catch (error) {
+    return next(error)
+  }
+  return res.redirect('/school/generated-pins-list')
+}
+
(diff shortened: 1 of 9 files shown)


admin/test/features/step_definitions/generate_pupil_pins_steps.rb
```diff
@@ -59,4 +59,28 @@
 And(/^I am on Generate pins Pupil List page$/) do
   step 'I navigate to generate pupil pins page'
   step 'I click Generate PINs button'
+  @page = generate_pupil_pins_page
```
activemq: Why do we need an instance variable here ? @sinhaamit1980

admin/test/features/step_definitions/pupil_not_taking_check.rb
```diff
@@ -66,6 +66,7 @@
 Given(/^I am on the pupil reason page$/) do
   step 'I am on the pupil not taking check page'
   step 'I want to add a reason'
+  @page = pupil_reason_page
```
activemq: Why do we need an instance variable here? @sinhaamit1980

admin/test/features/step_definitions/generate_pupil_pins_steps.rb
sinhaamit1980: We need this for sticky banner steps which is common across Pupil check and Generate pin page

admin/test/features/step_definitions/pupil_not_taking_check.rb
sinhaamit1980: We need this for sticky banner steps which is common across Pupil check and Generate pin page

PR #249, opened 2017-10-31
feature/17000 Bug fixes

GeorgeChatzigiannis:
Validate pin for authentication and compare pinExpiresAt value with current date time

### DIFF
--- a/admin/services/generate-pins.service.js
+++ b/admin/services/generate-pins.service.js
@@ -2,6 +2,10 @@ const moment = require('moment')
 const pupilDataService = require('../services/data-access/pupil.data.service')
 const randomGenerator = require('../lib/random-generator')
 
+const fourPMToday = () => {
+  return moment().startOf('day').add(16, 'hours')
+}
+
 const generatePinService = {
   /**
    * Fetch pupils and filter required only pupil attributes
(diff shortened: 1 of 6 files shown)


admin/services/generate-pins.service.js
```diff
@@ -14,7 +18,7 @@ const generatePinService = {
     let pupils = await pupilDataService.getSortedPupils(schoolId, sortField, sortDirection)
     // filter pupils
     pupils = pupils
-      .filter(p => !p.pin && !p.attendanceCode && !p.result)
+      .filter(p => !generatePinService.isValidPin(p.pin, p.pinExpiresAt) && !p.attendanceCode && !p.result)
```
jon-shipley: This whole line is a bit much. I would suggest moving this to a function - removePupilsWithPin or something

admin/services/generate-pins.service.js
```diff
@@ -37,28 +41,28 @@ const generatePinService = {
...
     }
     return school
   },
 
-  isValidPin: (pinExpiresAt) => {
-    if (!pinExpiresAt) return false
-    return moment(pinExpiresAt).isBefore(moment().startOf('day').add(16, 'hours'))
+  isValidPin: (pin, pinExpiresAt) => {
+    if (!pinExpiresAt || !pin) return false
+    return moment(pinExpiresAt).isAfter(moment.utc())
```
GuyHarwood: I would just pass the pupil into the method

GeorgeChatzigiannis: @GuyHarwood all I need really is these two properties though.

admin/services/generate-pins.service.js
GeorgeChatzigiannis: @jon-shipley updated

PR #254, opened 2017-11-02
Feature/16348 Update pupil and school documents instead of save

GeorgeChatzigiannis:
Revert find methods to return lean PoJos and replaced save with update methods

### DIFF
--- a/admin/services/data-access/pupil.data.service.js
+++ b/admin/services/data-access/pupil.data.service.js
@@ -95,7 +95,6 @@ pupilDataService.updateMultiple = async function (pupils) {
     savedPupils = results
   }, error => { throw new Error(error) }
   )
-  console.log(savedPupils)
   return savedPupils
 }
 
--- a/admin/services/data-access/pupil.data.service.js
+++ b/admin/services/data-access/pupil.data.service.js
@@ -90,7 +90,7 @@ pupilDataService.update = async function (id, doc) {
 pupilDataService.updateMultiple = async function (pupils) {
   // returns Promise
   let savedPupils = []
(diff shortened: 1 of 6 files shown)


admin/services/data-access/pupil.data.service.js
```diff
@@ -87,10 +87,10 @@ pupilDataService.update = async function (id, doc) {
   })
 }
 
-pupilDataService.saveMultiple = async function (pupils) {
+pupilDataService.updateMultiple = async function (pupils) {
   // returns Promise
   let savedPupils = []
-  await Promise.all(pupils.map(p => p.save())).then(results => {
+  await Promise.all(pupils.map(p => Pupil.updateOne({'_id': { $eq: p._id }}, p))).then(results => {
```
GuyHarwood: @GeorgeChatzigiannis can this not be an updateMany call?

admin/spec/service/data-access/pupil.data.service.spec.js
```diff
@@ -117,21 +117,27 @@ describe('pupil.data.service', () => {
       expect(mockPupil.verify()).toBe(true)
     })
   })
-  describe('#saveMultiple', () => {
+  describe('#updateMultiple', () => {
```
GuyHarwood: what's the hash for?

GeorgeChatzigiannis: it's just a pattern we are using to signify that's a method.
Happy to remove it if it's confusing

admin/services/data-access/pupil.data.service.js
pris54: Why do you need $eq with updateOne?

PR #258, opened 2017-11-02
Feature/16347 generated pins ui

GeorgeChatzigiannis:


admin/services/date.service.js
```diff
@@ -4,6 +4,7 @@ const moment = require('moment')
 const gdsFullFormat = 'D MMMM YYYY'
 const gdsShortFormat = 'D MMM YYYY'
 const gdsLongFormat = 'DD MMM YYYY'
+const customFormat = 'dddd DD MMMM'
```
jon-shipley: If this is mean to be quite English like - then why specify DD rather than just D
Sure Friday 3 November is more natural than Friday 03 November?

Please consider changing this.

admin/spec/service/date.service.spec.js
```diff
@@ -17,4 +17,11 @@ describe('date service', () => {
       expect(dateService.formatShortGdsDate(date)).toBe('31 Dec 2010')
     })
   })
+
+  describe('#formatDayAndDate', () => {
+    it('correctly formats a date', () => {
+      const date = new Date(2010, 11, 31, 14, 10, 0, 0)
+      expect(dateService.formatDayAndDate(date)).toBe('Friday 31 December')
```
jon-shipley: This function takes a date but a date is not passed in from the school controller and there is no test for that usage.   Should there be?

admin/services/date.service.js
GeorgeChatzigiannis: @jon-shipley changed

admin/spec/service/date.service.spec.js
GeorgeChatzigiannis: moment is handling that for me though, which means that I will be essentially testing the library if I add a case where there's a null date and I would expect to receive the current one in the specified format.
@jon-shipley thoughts?

jon-shipley: Just pass new Date() from the controller as the arg?

jon-shipley: https://www.sitepoint.com/functional-programming-pure-functions/

GeorgeChatzigiannis: @jon-shipley done

admin/test/features/support/mongo_db_helper.rb
```diff
@@ -79,11 +79,16 @@ def self.expire_pin(forename, lastname, school_id, flag=true)
...
     CLIENT[:pupils].update_one({'foreName': forename, 'lastName': lastname, 'school': school_id.to_i},
-                               {'$set' => {'pin' => NIL}})
+                               {'$set' => {'pin' => flag}})
   end
 
+  # def self.reset_pin(forename, lastname, school_id)
+  #   CLIENT[:pupils].update_one({'foreName': forename, 'lastName': lastname, 'school': school_id.to_i},
+  #                              {'$set' => {'pin' => NIL}})
+  # end
+
```
activemq: @sinhaamit1980 delete if not needed

admin/test/features/support/page_objects/generate_pupil_pins_page.rb
```diff
@@ -5,11 +5,9 @@ class GeneratePupilPinsPage < SitePrism::Page
   element :generate_pin_message, '.column-two-thirds', text: 'Pupils will need a Personal Identification Number (PIN) and school password in order to start the check. These expire daily.'
   elements :info_message, '.list-number li'
   element :generate_pin_btn, 'input[value="Generate PINs"]'
-  element :generate_more_pin_btn, 'input[value="Generate more PINs"]'
-  element :school_password, '.column-two-thirds br'
 
-  element :select_all_pupils, '#selectAll'
-  element :unselect_all_pupils, '#selectAll'
+  element :select_all_pupils, '#tickAllCheckboxes'
+  element :unselect_all_pupils, '#tickAllCheckboxes'
```
activemq: @sinhaamit1980 as its the same element, try getting a unique id via the text also, ie 
element :select_all_pupils, '#tickAllCheckboxes', text: 'Select all'

sinhaamit1980: Done

admin/test/features/support/mongo_db_helper.rb
sinhaamit1980: Done

PR #265, opened 2017-11-06
Bug/17228 delete second check start

jon-shipley:
Remove a duplicated block of code that was causing checkStart() to be called twice when the pupil logs in.


jon-shipley:
Looking for 1 dev review please

@sinhaamit1980 @activemq - looking for 1 tester please

pupil-spa/src/app/question/question.component.ts
```diff
@@ -87,7 +87,7 @@ export class QuestionComponent implements OnInit, AfterViewInit {
    */
   @HostListener('document:keydown', ['$event'])
   handleKeyboardEvent(event: KeyboardEvent) {
-    console.log('question.component: handleKeyboardEvent(): event: ', event);
+    // console.log('question.component: handleKeyboardEvent(): event: ', event);
```
GeorgeChatzigiannis: will you need this console log?
can always revert from git history.
I'd rather keep it as clean and 'comment-less' as possible

jon-shipley: Yes - these will be replaced by a call to a logger Real Soon Now

PR #267, opened 2017-11-06
bug/16347 Display DoB for duplicate names in generated pins page

GeorgeChatzigiannis:
* If pupil has same first name and surname, then date of birth is also displayed 
* If pupil has same first name, surname name and same date of birth, then middle name is also displayed

### DIFF
--- a/admin/services/pupil.service.js
+++ b/admin/services/pupil.service.js
@@ -89,11 +89,11 @@ const pupilService = {
     return true
   },
   /**
-   * Adds show Date of Birth flag for pupils with equal full names
+   * Adds show Date of Birth flag for pupils that have been alphabetically sorted by last name and have equal full names
    * @param {Array} pupils
    * @returns {Array}
    */
-  requiresExtraFields: (pupils) => {
+  addIdentificationFlags: (pupils) => {
     pupils.forEach((p, i) => {
       const currentPupil = pupils[ i ]
       const nextPupil = pupils[ i + 1 ]
--- a/admin/services/pupil.service.js
+++ b/admin/services/pupil.service.js
@@ -87,6 +87,18 @@ const pupilService = {
       }
     }
     return true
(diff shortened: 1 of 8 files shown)


admin/spec/service/pupil.service.spec.js
```diff
@@ -65,4 +66,19 @@ describe('pupil service', () => {
...
     })
   })
+  describe('displayDoB', () => {
+    it('has showDoB property as true for pupils with same fullname', () => {
+      const pupil1 = Object.assign({}, pupilMock)
+      const pupil2 = Object.assign({}, pupilMock)
+      const pupils = pupilService.displayDoB([pupil1, pupil2])
+      expect(pupils[0].showDoB && pupils[1].showDoB).toBeTruthy()
+    })
+    it('has showDoB property as fa;se for pupils with different fullname', () => {
```
GuyHarwood: missing l in false :)

GeorgeChatzigiannis: @GuyHarwood updated

admin/services/pupil.service.js
```diff
@@ -87,6 +87,28 @@ const pupilService = {
       }
     }
     return true
+  },
+  /**
+   * Adds show Date of Birth flag for pupils with equal full names
```
jon-shipley: There should be a note to say this *assumes* the array of objects have been sorted by some criteria.

admin/services/pupil.service.js
```diff
@@ -87,6 +87,28 @@ const pupilService = {
...
+   * @param {Array} pupils
+   * @returns {Array}
+   */
+  requiresExtraFields: (pupils) => {
```
jon-shipley: This name seems a bit vague to me.  Perhaps something like `addIdentificationFlags` is more descriptive?

PR #268, opened 2017-11-07
Feature/16349 print pins

GeorgeChatzigiannis:
This PR also covers the isolation of pupil pin functionality into a separate router & controller


admin/controllers/pupil.js
```diff
@@ -257,6 +261,28 @@ const postEditPupil = async (req, res, next) => {
...
 
+const getPrintPins = async (req, res, next) => {
+  res.locals.pageTitle = 'Print pupils'
+  let pupils
+  let school
+  let qrDataURL
+  const date = dateService.formatDayAndDate(new Date())
+  try {
+    pupils = await generatePinsService.getPupilsWithActivePins(req.user.School)
+    school = await generatePinsService.getActiveSchool(req.user.School)
```
jon-shipley: Surprising.  Is their logic in this call particular to generating pins? I would expect the call to be to the SchoolService.

admin/spec/service/qr.service.spec.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+/* global describe, it, expect */
+
+const qrService = require('../../services/qr.service')
+const qrURL = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAIQAAACECAYAAABRRIOn' +
```
jon-shipley: This looks like it would be better as a mock

admin/controllers/pupil.js
```diff
@@ -257,6 +261,28 @@ const postEditPupil = async (req, res, next) => {
   res.redirect(`/school/pupil-register/lastName/true?hl=${pupilId}`)
 }
 
+const getPrintPins = async (req, res, next) => {
```
jon-shipley: I think Pupil is large enough.  Now might be a good time to start splitting functionality.  How about creating a PupilPrintPinController?

GeorgeChatzigiannis: @jon-shipley print pin sounds too dedicated. How about one for generate pins instead?
Thoughts?

admin/controllers/pupil.js
GeorgeChatzigiannis: I think I should rename this to getSchoolWithActivePass

admin/controllers/pupil.js
jon-shipley: How about pupil pin controller then?  To deal with generating, and printing for sure.  Maybe viewing?

admin/controllers/pupil.js
GeorgeChatzigiannis: well its relevant to fetching an active pass that's why I kept it in this service.

GeorgeChatzigiannis: Perhaps it's better for it to lie under school service yea

admin/controllers/pupil.js
GeorgeChatzigiannis: Yea I'll start refactoring

admin/config.js
```diff
@@ -16,5 +16,6 @@ module.exports = {
   ROLE_TEST_DEVELOPER: 'TEST-DEVELOPER',
   ROLE_TEACHER: 'TEACHER',
   HOME_TEST_DEVELOPER: '/administrator',
-  HOME_TEACHER: '/school/school-home'
+  HOME_TEACHER: '/school/school-home',
```
GuyHarwood: we don't want routes in the config file.  I asked Paula to make a similar change in another PR, please ask her about what she did to change this.  Think she made a routeConfig.js file?

admin/controllers/pupil.js
GuyHarwood: Good to see this happening

admin/controllers/pupil.js
GuyHarwood: if it's returning the school pin, then rename the method to something more appropriate like 'getSchoolPin'.

The current name doesn't make any sense in this context anyway

admin/config.js
jon-shipley: @GeorgeChatzigiannis - looks like these routes are still here.

GeorgeChatzigiannis: @GuyHarwood @jon-shipley I'll have to ask @pris54 about them since they were introduced from her on this commit bcf8766dcc3399f7e503cb93c4e9b4b5c001f210

GuyHarwood: 👍🏻 Please

admin/assets/stylesheets/elements/_printing.scss
```diff
@@ -32,14 +32,23 @@
...
-    max-height: 180px;
-    max-width: 200px;
-  }
   .pupil-content {
+    text-align: left;
     line-height: 2em;
     font-size: .8em;
   }
+  .qr {
+    width: 5em;
```
pris54: @GeorgeChatzigiannis Pixels remain the preferred unit to use as it makes it easier to keep our styles consistent with the ones provided by GOV.UK, also easier to use $gutter variables and keep a standardise use of the space.

admin/services/pin-generation.service.js
```diff
@@ -0,0 +1,92 @@
+const moment = require('moment')
+const pupilDataService = require('../services/data-access/pupil.data.service')
+const randomGenerator = require('../lib/random-generator')
+const mongoose = require('mongoose')
+const pinValidator = require('../lib/validator/pin-validator')
+const pupilIdentificationFlagService = require('../services/pupil-identification-flag.service')
+
+const fourPMToday = () => {
```
pris54: @GeorgeChatzigiannis As discussed in a previous PR, the correct camel case would be fourPmToday

admin/spec/service/qr.service.spec.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+/* global describe, it, expect */
+
+const qrService = require('../../services/qr.service')
+const qrURL = require('../mocks/qr-url')
```
pris54: @GeorgeChatzigiannis same issue re' camel. I won't add comments every time I see it =)

admin/views/pupil-pin/pin-print.ejs
```diff
@@ -0,0 +1,72 @@
...
+            </form>
+        </div>
+
+        <div class="heading-large">Multiplication Table Check</div>
+        <ul class="list list-bullet font-xsmall" role="listbox" tabindex="0">
+            <li role="listitem">Please cut out and distribute to pupils</li>
+            <li role="listitem">The school passwords and pupil PINs will expire 4pm on <%= date %></li>
+            <li role="listitem">New PINs will need to be generated if not used by 4pm on <%= date %></li>
+        </ul>
+        <div class='pupil-print-table'>
```
pris54: @GeorgeChatzigiannis We use double quotes for CSS classes and ids.

admin/services/pin-generation.service.js
GeorgeChatzigiannis: @pris54 https://msdn.microsoft.com/en-us/library/141e06ef(v=vs.71).aspx
When using acronyms, use Pascal case or camel case for acronyms more than two characters long. For example, use HtmlButton or htmlButton. However, you should capitalize acronyms that consist of only two characters, such as System.IO instead of System.Io.

pris54: @GeorgeChatzigiannis This is different from the convention I have been using for quite a while (linked to Wikipedia provided in previous PR). We should come up with an agreement on how to deal with these cases.

GeorgeChatzigiannis: Most of the links I found are supporting capitalization for two character acronyms. I believe we should keep it this way.

GuyHarwood: @GeorgeChatzigiannis i disagree.  This about readability.  It just looks like PMT is the acronym.  Change it to fourPmToday please.

admin/spec/service/qr.service.spec.js
GuyHarwood: also agree.  qrUrl please

admin/assets/stylesheets/elements/_printing.scss
GeorgeChatzigiannis: @pris54 i see we are using both em and px everywhere?
I am happy to change this to px but we should also take notice then and update the whole css codebase with px units.
I'll introduce a similar variable to gutter

admin/assets/stylesheets/elements/_top_navigation.scss
```diff
@@ -0,0 +1,62 @@
...
+  &:link,
+  &:active,
+  &:visited,
+  &:hover {
+    color: $text-colour;
+    text-decoration: none;
+  }
+}
+
+@media (max-width: 640px) {
```
pris54: @GeorgeChatzigiannis I believe from line 47 to 62 can be deleted as it's not being used.

GeorgeChatzigiannis: done

PR #269, opened 2017-11-07
bug/17325 Unique upn validation fix

GeorgeChatzigiannis:
UPN validation should only trigger when id is different under the same UPN

### DIFF
--- a/admin/lib/validator/pupil-validator.js
+++ b/admin/lib/validator/pupil-validator.js
@@ -122,7 +122,8 @@ module.exports.validate = async (pupilData) => {
   if (!(validationError.get('upn'))) {
     const existingPupil = await pupilDataService.findOne({ upn: pupilData.upn })
     // if pupil is not stored already under the same id and UPN
-    if (existingPupil && pupilData._id && existingPupil._id.toString() !== pupilData._id) {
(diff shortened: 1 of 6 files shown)


admin/lib/validator/pupil-validator.js
```diff
@@ -121,7 +121,8 @@ module.exports.validate = async (pupilData) => {
   // We need to check that the UPN is unique
   if (!(validationError.get('upn'))) {
     const existingPupil = await pupilDataService.findOne({ upn: pupilData.upn })
-    if (existingPupil) {
+    // if pupil is not stored already under the same id and UPN
```
jon-shipley: If I am reading this correctly UPN is not validated if pupil does not have an _id field.  Is there any chance of that happening, e.g. during create()

GeorgeChatzigiannis: The same validation is being used on both edit and create a pupil record.
Essentially this fixes the issue where editing a pupil would trigger validation to have an error due to getting the exact same user and thus disallowing the edit to go through.

GeorgeChatzigiannis: The change should only trigger validation if an existing upn is used from a pupil with different _id.

jon-shipley: Same issue?

PR #270, opened 2017-11-07
Feature/17380 add speech synth user

jon-shipley:


jon-shipley:
@GeorgeChatzigiannis - the randomProp will get removed from the document, and the speechSynthesis prop will get added with the default value of false.  That *is* the benefit of a tightly controlled sub-document.

GeorgeChatzigiannis:
@jon-shipley fair enough

PR #271, opened 2017-11-07
Add new report to show just the scores. 

jon-shipley:
This is a cut-down simple version of the full report.


jon-shipley:
Looking for 1 approval please

admin/services/psychometrician-report.service.js
```diff
@@ -159,4 +183,14 @@ psychometricianReportService.produceReportData = function (completedCheck) {
   return psData
 }
 
+/**
+ * Filter an Array of psreportcache object to a minimal set
```
GeorgeChatzigiannis: since you added it above might as well add here the @return part of this method

PR #272, opened 2017-11-08
Feature/17327 check forms new role

pris54:
PBIs related: 17327 and 17330


admin/config.js
```diff
@@ -13,8 +13,10 @@ module.exports = {
...
   QUESTION_TIME_LIMIT: 5,
   TIME_BETWEEN_QUESTIONS: 2,
+  ROLE_SERVICE_MANAGER: 'SERVICE-MANAGER',
   ROLE_TEST_DEVELOPER: 'TEST-DEVELOPER',
   ROLE_TEACHER: 'TEACHER',
-  HOME_TEST_DEVELOPER: '/administrator',
-  HOME_TEACHER: '/school/school-home'
+  HOME_TEACHER: '/school/school-home',
+  HOME_TEST_DEVELOPER: '/test-developer/home',
+  HOME_SERVICE_MANAGER: '/service-manager/home'
```
GuyHarwood: feels like we should have a dedicated roles file and a routes configuration separate from the main config

admin/routes/test-developer.js
```diff
@@ -8,9 +8,15 @@ const CheckForm = require('../models/check-form')
...
+const config = require('../config')
+const testDeveloperController = require('../controllers/test-developer')
 
+router.get('/', isAuthenticated(config.ROLE_TEST_DEVELOPER), (req, res, next) => testDeveloperController.getTestDeveloperHome(req, res, next))
+router.get('/home', isAuthenticated(config.ROLE_TEST_DEVELOPER), (req, res, next) => testDeveloperController.getTestDeveloperHome(req, res, next))
+
+/* @TODO: The code below is meant to be refactored */
 /* GET manage check forms page. */
-router.get('/manage-check-forms', isAuthenticated(), async function (req, res, next) {
+router.get('/manage-check-forms', isAuthenticated(config.ROLE_TEST_DEVELOPER), async function (req, res, next) {
```
GuyHarwood: looks like we are bundling up routes by job role into controllers.  This convention is ok until another role is given access to a specific route, and then it just doesn't make sense.  I would much rather have controllers grouped by feature, such as 'check-forms-controller'.

pris54: I agree. There is more work coming only to deal with roles so I expect that's when we will look more thoroughly into the current solution we've got in place (which we've always known that is a first step only). 
For check forms I will split the controller by feature, I'll do backwards refactoring for check windows too (as it's related), however this story is just to unblock me to carry on with the actual 'check forms' work.

GuyHarwood: Ok 👍

GuyHarwood: @pris54 does that mean you will break up the controller in this PR, or the actual 'check forms' work?

pris54: @GuyHarwood I'll break-up the controller in the next PR, that is, in the actual check forms work (this is just so that I can start working on it)

PR #275, opened 2017-11-08
SPA: create speech service

jon-shipley:
Add Speech service

Preparatory work at this stage.  Not hooked up.


GuyHarwood:
LGTM 👍

PR #276, opened 2017-11-08
Feature/15093 check window further tests

pris54:
Check windows was developed before test coverage was required, these are the initial unit tests for the feature.


admin/spec/lib/validator/check-window-validator.spec.js
```diff
@@ -34,95 +38,257 @@ describe('Check window validator', function () {
...
-    req.body.checkEndMonth = ''
-    let validationError = await checkWindowValidator.validate(req)
-    expect(validationError.hasError()).toBe(true)
-    done()
+  describe('When editing a check window', function () {
+    it('should require a check window name when editing', async function (done) {
+      req.body.checkWindowName = ''
+      req.body.checkWindowId = 'abc123'
+      const validationError = await checkWindowValidator.validate(req)
+      expect(validationError.hasError()).toBe(true)
```
GuyHarwood: how can we be sure that it's this exact field that has the error?  i would usually go for an explicit assertion such as `expect(validationError.message).toBe('check name is required')`

admin/spec/mocks/check-window.js
```diff
@@ -1,12 +1,15 @@
 const mongoose = require('mongoose')
 mongoose.Promise = global.Promise
 const moment = require('moment')
+const {Types: {ObjectId}} = require('mongoose')
```
GuyHarwood: makes sense in this context (only one usage) but please import entire objects by default

admin/spec/models/check-window.spec.js
```diff
@@ -35,7 +50,7 @@ describe('check-window schema', function () {
     })
   })
 
-  it('requires a adminStartDate property', function (done) {
+  it('should require a adminStartDate property', function (done) {
```
GuyHarwood: again, i would assert the error details

`expect(error.message).toBe('admin start date required')`

admin/spec/models/check-window.spec.js
```diff
@@ -53,12 +68,100 @@ describe('check-window schema', function () {
...
+        expect(cw.checkEndDate.toString()).toBe(checkEndDate.toString())
+        expect(cw.isDeleted).toBe(false)
+        done()
+      } catch (err) {
+        console.log('ERROR', err)
+      }
+      done()
+    })
+
+    it('should fails', async function (done) {
```
GuyHarwood: typo.  also, its not clear why it should fail, please add a condition to the description

admin/spec/models/check-window.spec.js
```diff
@@ -53,12 +68,100 @@ describe('check-window schema', function () {
...
+        expect(form5Id._id.toString()).toBe('59e8e10039dbad3cd033719a')
+        expect(form8Id._id.toString()).toBe('59e88622d38a9f2d1fcebbb3')
+        done()
+      } catch (err) {
+        console.log('ERROR', err)
+      }
+      done()
+    })
+
+    it('should fail', async function (done) {
```
GuyHarwood: again, what's the failure condition?

admin/spec/models/check-window.spec.js
```diff
@@ -53,12 +68,100 @@ describe('check-window schema', function () {
...
+  describe('Method getCheckWindowsAssignedToForms', function () {
+    it('should return the expected results', async function (done) {
+      expect(typeof CheckWindow.getCheckWindowsAssignedToForms).toBe('function')
+      sandbox
+        .mock(CheckWindow)
+        .expects('find')
+        .chain('exec')
+        .resolves(CheckWindowMocks)
+
+      try {
```
GeorgeChatzigiannis: is the try catch needed here?
i would assume you would only need it in case you want to add an expectation on the catch part

admin/spec/models/check-window.spec.js
```diff
@@ -2,21 +2,36 @@
 
 /* global beforeEach, describe, it, expect */
 
+const mongoose = require('mongoose')
+mongoose.Promomise = global.Promise
```
jon-shipley: typo

PR #280, opened 2017-11-09
Pass speech synth config to SPA

jon-shipley:
Pass the speech synthesis config to the SPA


jon-shipley:
Looking for 1 approval

PR #285, opened 2017-11-15
Feature/17402 check forms refactoring

pris54:
More refactoring is coming in the following PRs.


admin/spec/models/check-window.spec.js
```diff
@@ -80,82 +80,4 @@ describe('check-window schema', function () {
       done()
     })
   })
-
-  describe('Method getCheckWindow', function () {
```
GuyHarwood: would i be right in assuming that you are removing these as all they do is test the mongoose model (which is pointless IMO)?

do we have tests for the service?

admin/services/check-window.service.js
```diff
@@ -53,12 +52,101 @@ const checkWindowService = {
...
+   *  each form is assigned to.
+   * @return {Promise}
+   */
+  getCheckWindowsAssignedToForms: () => {
+    return new Promise(async (resolve, reject) => {
+      let checkWindows
+      let data = {}
+
+      try {
+        checkWindows = await CheckWindow.find({}).exec()
```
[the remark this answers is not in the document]
pris54: @GuyHarwood Methods such as 'getCheckWindowsAssignedToForms' have not being refactored yet, these are part of PBI 17403 (not in development yet). All the Mongoose usage in the services will be moved to data.services. In this PR routes and controllers were decoupled so you'll see this code as altered but only at a high level.

admin/spec/models/check-window.spec.js
pris54: @GuyHarwood Methods were moved from models to data.services, and those tests were deleted, not sure how you see this...

GuyHarwood: that's what I had hoped.  There is little value in testing the mongoose models

admin/services/check-window.service.js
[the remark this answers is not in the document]
GuyHarwood: 👍

admin/controllers/check-form.js
```diff
@@ -0,0 +1,269 @@
...
+ * @returns {Promise.<*>}
+ */
+const uploadAndViewForms = async (req, res, next) => {
+  res.locals.pageTitle = 'Upload and view forms'
+  try {
+    const forms = await checkFormDataService.getActiveForms().sort({createdAt: -1}).exec()
+    let formData = forms.map(e => { return e.toJSON() })
+    const checkWindows = await checkWindowService.getCheckWindowsAssignedToForms()
+
+    formData.forEach(f => {
```
GeorgeChatzigiannis: I see you are doing some data processing here, shouldn't that be included in a service in order to keep the controller more lean?

admin/controllers/check-form.js
```diff
@@ -0,0 +1,269 @@
...
+    const CheckWindow = await checkFormDataService.getActiveForm(id).exec()
+    if (!CheckWindow) {
+      return next(new Error(`Unable to find form.id [${id}]`))
+    }
+    // Un-assign check-form from any check-windows
+    const checkWindowsByForm = await checkWindowService.getCheckWindowsAssignedToForms()
+    if (checkWindowsByForm[CheckWindow._id]) {
+      // Array of CheckWindows models, each with a forms array
+      let modifiedCheckWindows = []
+      checkWindowsByForm[CheckWindow._id].forEach(cw => {
```
GeorgeChatzigiannis: I would expect these logic here to be added in a service as well.
It will make your tests way easier/simpler

admin/controllers/check-form.js
```diff
@@ -0,0 +1,269 @@
...
+const removeCheckForm = async (req, res, next) => {
+  const id = req.params.formId
+  try {
+    const CheckWindow = await checkFormDataService.getActiveForm(id).exec()
+    if (!CheckWindow) {
+      return next(new Error(`Unable to find form.id [${id}]`))
+    }
+    // Un-assign check-form from any check-windows
+    const checkWindowsByForm = await checkWindowService.getCheckWindowsAssignedToForms()
+    if (checkWindowsByForm[CheckWindow._id]) {
```
GeorgeChatzigiannis: I would expect these logic here to be added in a service as well.
It will make your tests way easier/simpler.

admin/controllers/check-form.js
```diff
@@ -0,0 +1,269 @@
...
+  // Various errors cause a page to be rendered instead, and it *needs* a title
+  res.locals.pageTitle = 'Upload check form'
+  req.breadcrumbs(res.locals.pageTitle)
+
+  // Pick up the list of forms in case we have an error and need to re-render the page
+  try {
+    forms = await checkFormDataService.getActiveForms().sort({createdAt: -1}).exec()
+    formData = forms.map(e => { return e.toJSON() })
+    const checkWindows = await checkWindowService.getCheckWindowsAssignedToForms()
+    formData.forEach(f => {
```
GeorgeChatzigiannis: same as above.

admin/routes/test-developer.js
```diff
@@ -2,152 +2,22 @@
...
+router.get('/', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.getTestDeveloperHome(req, res, next))
+router.get('/home', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.getTestDeveloperHome(req, res, next))
+router.get('/view-form/:formId', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.displayCheckForm(req, res, next))
+router.get('/upload-and-view-forms', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.uploadAndViewForms(req, res, next))
+router.get('/delete-form/:formId', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.removeCheckForm(req, res, next))
+router.get('/upload-new-form', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.uploadCheckForm(req, res, next))
+router.post('/upload-new-form', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.saveCheckForm(req, res, next))
```
GeorgeChatzigiannis: finally a cleaner test developer route 👍

GeorgeChatzigiannis:
The following files are missing corresponding tests:
- admin/controllers/check-form.js (whole spec)
- admin/services/data-access/check-form.data.service.js (whole spec)
- admin/services/data-access/check-window.data.service.js (whole spec)
- admin/services/check-window.service.js (getCheckWindowsAssignedToForms)

admin/controllers/check-form.js
pris54: @GeorgeChatzigiannis As explained in private and dev chats, "285 is a high level refactoring of check forms, it doesn’t include all the refactoring but the core modifications" - More refactoring of controllers, services, etc is coming next. This PR was getting too big.

admin/controllers/check-form.js
pris54: See previous comment

admin/controllers/check-form.js
pris54: Same reply, this PR is only the correct routing and basic pages, more code refactoring comes next -or else this PR will become way too big.

admin/test/features/upload_view_forms.feature
```diff
@@ -0,0 +1,28 @@
...
+    Then I should see a heading on Upload and View forms page
+
+  Scenario: Upload and view forms page has a option to upload new form
+    Then I should have the option to upload new form
+
+  Scenario: Upload new forms page has a title
+    And I am on the Upload new forms page
+    Then I should see a heading on Upload new form page
+    And I can see sub heading 'Download example' on Upload new form Page
+    And I can see sub heading 'Upload file(s)' on Upload new form Page
```
activemq: @sinhaamit1980 Please try and keep to a 1 or 2 assertions per scenario. This looks like it has 4, which is not good as if the first one fails we have no feedback on the others, split them in to their own scenarios

admin/controllers/check-form.js
```diff
@@ -0,0 +1,269 @@
...
+ * Remove check form.
+ * @param req
+ * @param res
+ * @param next
+ * @returns {Promise.<*>}
+ */
+const removeCheckForm = async (req, res, next) => {
+  const id = req.params.formId
+  try {
+    const CheckWindow = await checkFormDataService.getActiveForm(id).exec()
```
jon-shipley: I *think* that getActiveForm() should return a plain javascript object, so this should be `checkWindow` not `CheckWindow`, but I am confused by the .exec() on the end, which looks like mongoose is bleeding into the controller which definitely shouldn't happen.

admin/controllers/check-form.js
jon-shipley: We should not have mongoose models in the controller.

admin/routes/test-developer.js
```diff
@@ -166,12 +36,12 @@ router.get('/choose-check-window', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOP
...
     } else {
       console.error(new Error('No formIds selected'))
-      return res.redirect('manage-check-forms')
+      return res.redirect('upload-and-view-forms')
     }
   }
 
   try {
-    forms = await CheckForm.getActiveForms({_id: formIds}).sort({_id: 1}).exec()
+    forms = await checkFormService.getActiveForms({_id: formIds}).sort({_id: 1}).exec()
```
jon-shipley: The checkFormService should be returning JSON objects, but the line below makes me think otherwise...

admin/services/data-access/check-form.data.service.js
```diff
@@ -0,0 +1,26 @@
...
+
+const CheckForm = require('../../models/check-form')
+
+const checkFormDataService = {
+  /**
+   * Get form by id, when isDeleted is false.
+   * @param id
+   */
+  getActiveForm: (id) => {
+    return CheckForm.findOne({_id: id, isDeleted: false})
```
jon-shipley: This really really needs to return the JSON object.  So `.lean().exec()` is required here.

admin/services/data-access/check-form.data.service.js
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Get all forms, when isDeleted is false.
+   * @param q
+   */
+  getActiveForms: (q) => {
+    if (!q) {
+      q = {}
+    }
+    q.isDeleted = false
+    return CheckForm.find(q)
```
jon-shipley: This needs to return JSON objects, not mongoose models.  `.lean().exec` is required.  It would then need to be async.

admin/services/data-access/check-window.data.service.js
```diff
@@ -47,6 +47,18 @@ const checkWindowDataService = {
...
+   * Fetch (one) check window for present date.
+   * @returns {Promise.<*>}
+   */
+  fetchCurrentCheckWindow: async () => {
+    const now = new Date()
+    const checkWindow = await CheckWindow.findOne({startDate: {$lte: now}, endDate: {$gte: now}}).exec()
+    if (!checkWindow) {
+      throw new Error('No checkwindow is currently available')
+    }
+    return checkWindow
```
jon-shipley: This should be returning the `.lean()` plain object, not the mongoose model.

pris54:
@GuyHarwood @jon-shipley Thanks for reviewing. This first PR is mainly about routing and moving code to new files to create the correct structure. There is more refactoring coming next where 1) controllers are simplified by using services 2) services and data services are expanded 3) existing methods are improved - Test coverage was simplified whenever possible, then next comes further coverage as the refactoring is completed.

pris54:
@GeorgeChatzigiannis @GuyHarwood @jon-shipley Thank you for reviewing. I will in short close this PR, I will then create a new one with a massive refactoring, all in one PR, including all your recommendations -most of which were meant to be in a separate PR (as explained) for the sake of keeping PRs at a reasonable size.

GuyHarwood:
@pris54 rather than a 'massive refactoring', why not consider breaking the work up into bitesize chunks.  I think the wider issue with this PR is that it appears to make things worse than they already are.  Maybe focus on one service/controller/etc per PR and do it TDD?

Creating a single huge PR isn't going to make things any better.  I think we should go through this work tomorrow, after standup, and see what strategy we can come up with.

pris54:
@GuyHarwood That's why I tried to do by addressing first routing and basic structure, check forms feature is big and code is interconnected in a way that smaller PRs are going to be even more messy.

PR #287, opened 2017-11-17
Pupil PIN expiration and check start timestamp addition once pupil starts the check

GeorgeChatzigiannis:


admin/services/pin.service.js
```diff
@@ -33,4 +36,17 @@ pinService.getActiveSchool = async (schoolId) => {
...
+ * Expire pupil's pin
+ * @param token
+ * @param checkCode
+ */
+pinService.expirePupilPin = async (token, checkCode) => {
+  const decoded = jwtService.decode(token)
+  const pupil = await pupilDataService.findOne({_id: ObjectId(decoded.sub)})
+  const currentTimeStamp = moment.utc()
+  await checkDataService.update({checkCode: checkCode}, { '$set': { checkStartedAt: currentTimeStamp } })
+  await pupilDataService.update(pupil._id, { pinExpiresAt: currentTimeStamp })
```
jon-shipley: Should the pin be reset to null as well?

admin/spec/controllers/check-started.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      const req = getReq(goodReqParams)
+      const controller = require('../../controllers/check-started').preCheck
+      spyOn(jwtService, 'verify').and.returnValue((Promise.reject(new Error('error'))))
+      await controller(req, res)
+      const data = JSON.parse(res._getData())
+      expect(res.statusCode).toBe(400)
+      expect(data.error).toBe('Bad request')
+      done()
+    })
+    it('verifies a pupil as unauthorized ', async (done) => {
```
jon-shipley: I think this test should say something about why the login will be unauthorised.

admin/spec/service/pin.service.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+describe('pin.service', () => {
+  let sandbox
+
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create()
+  })
+
+  afterEach(() => sandbox.restore())
+  describe('getPupilsWithActivePins', () => {
```
jon-shipley: how about a newline to separate functions?

admin/spec/service/pin.service.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    let sandbox
+    let pupil1
+    let pupil2
+    beforeEach(() => {
+      sandbox = sinon.sandbox.create()
```
jon-shipley: sandbox has already been created - this is duplication

admin/spec/service/pin.service.spec.js
```diff
@@ -0,0 +1,116 @@
...
+      pupil1 = Object.assign({}, pupilMock)
+      pupil1.pin = 'f55sg'
+      pupil1.pinExpiresAt = moment().startOf('day').add(16, 'hours')
+      pupil2 = Object.assign({}, pupilMock)
+      pupil2._id = '595cd5416e5ca13e48ed2520'
+      pupil2.pinExpiresAt = moment().startOf('day').add(16, 'hours')
+    })
+    afterEach(() => sandbox.restore())
```
jon-shipley: duplication - not needed?

pupil-spa/src/app/services/audit/auditEntry.ts
```diff
@@ -1,6 +1,6 @@
-export type AuditEntryType = 'WarmupIntroRendered' | 'WarmupCompleteRendered' | 'CheckComplete' |
-'CheckStarted' | 'QuestionRendered' | 'QuestionAnswered' | 'PauseRendered' | 'RefreshDetected' |
-  'UtteranceStarted' | 'UtteranceEnded';
+export type AuditEntryType = 'WarmupIntroRendered' | 'WarmupCompleteRendered' | 'CheckStartedAPICallSucceed' |
```
jon-shipley: CheckStartedAPICallSucceed is in the present tense, but the other verbs are all past tense.  Suggest rename to `CheckStartedAPICallSucceeded`

pupil-spa/src/app/services/audit/auditEntry.ts
```diff
@@ -1,6 +1,6 @@
...
+'CheckStartedAPICallFailed' | 'CheckComplete' |'CheckStarted' | 'QuestionRendered' | 'QuestionAnswered' |
```
jon-shipley: I am not sure about auditing CheckStartedApiCalledFailed and CheckStartedApiCallSucceed.  Why not just log CheckStarted?  We don't log any other call returns in the AuditLog, and it seems to be a lower level concern.

pupil-spa/src/app/services/submission/submission.service.ts
```diff
@@ -9,6 +9,25 @@ export class SubmissionService {
...
+      { checkCode, accessToken },
+      requestArgs)
+      .toPromise()
+      .then((response) => {
+          if (response.status !== 201) {
+            return new Error('Submit Error:' + response.status + ':' + response.statusText);
+          }
+        },
+        (err) => { throw new Error(err); }
+      ).catch((err) => { throw new Error(err); });
```
jon-shipley: Will this api call make any retries if the server is busy? E.g. 504, or 504?  It probably should back off and retry.

admin/app.js
```diff
@@ -218,6 +219,7 @@ app.use('/pupil-pin', pupilPin)
 app.use('/api/questions', questions)
```
GuyHarwood: i desperately want to change this to `/api/pupil-login` but i think we should leave it until we extract out the endpoints into separate app(s)

admin/controllers/check-started.js
```diff
@@ -0,0 +1,33 @@
...
+const apiResponse = require('./api-response')
+
+/**
+ * Expires the pupil's pin
+ * @param req
+ * @param res
+ * @returns { object }
+ */
+
+const preCheck = async (req, res) => {
```
GuyHarwood: can we call this checkStarted or something similar?  preCheck doesnt feel accurate.

admin/models/check.js
```diff
@@ -34,6 +34,9 @@ const Check = new Schema({
     type: Date,
     required: true
   },
+  checkStartedAt: {
```
GuyHarwood: this needs adding to the db schemas too

admin/services/check-start.service.js
```diff
@@ -29,14 +29,14 @@ const checkStartService = {
...
     }
-
     // Save the details to the `Check` collection
     const check = new Check({
       pupilId,
       checkCode,
       checkWindowId: checkWindow._id,
       checkFormId: checkForm._id,
-      pupilLoginDate: new Date()
+      pupilLoginDate: new Date(),
```
GuyHarwood: elsewhere we are using moment.  I would like to see consistency across the entire app

pretty sure we have a date service?

admin/services/pin.service.js
```diff
@@ -33,4 +36,17 @@ pinService.getActiveSchool = async (schoolId) => {
...
 
+/**
+ * Expire pupil's pin
+ * @param token
+ * @param checkCode
+ */
+pinService.expirePupilPin = async (token, checkCode) => {
+  const decoded = jwtService.decode(token)
+  const pupil = await pupilDataService.findOne({_id: ObjectId(decoded.sub)})
+  const currentTimeStamp = moment.utc()
```
GuyHarwood: does this work OK in unit tests?  would it not be better to use a date service (which i think we have) so you can mock this?

pupil-spa/src/app/services/audit/auditEntry.ts
GuyHarwood: i think we should audit the attempt of the call `CheckStartedApiCalled' and the result.  So i would go with...

`CheckStartedApiCalled`
`CheckStartedApiCallSucceeded`
`CheckStartedApiCallFailed`

pupil-spa/src/app/services/submission/submission.service.spec.ts
```diff
@@ -32,6 +32,20 @@ describe('SubmissionService', () => {
...
+    mockBackend.connections.subscribe((connection) => {
+      connection.mockRespond(new Response(new ResponseOptions({
+        body: JSON.stringify({isSaved: true}),
+        status: 200
+      })));
+    });
+    spyOn(storageService , 'getItem').and.returnValues({});
+    submissionService.submitCheckStartData().then(() => {
+      expect(storageService.getItem).toHaveBeenCalledWith('pupil');
+      expect(storageService.getItem).toHaveBeenCalledWith('access_token');
```
GuyHarwood: FYI: we will be moving the JWT token to http-only very soon.  This means it will be sent with each request by the browser, not as part of the body.

pupil-spa/src/app/services/submission/submission.service.ts
GuyHarwood: remember that we do not want this error to disrupt the check.  I think we should be handling this elegantly here, logging the failure and not throwing it to the calling component/service.

admin/services/pin.service.js
GeorgeChatzigiannis: @jon-shipley introduced as part of the update

admin/spec/service/pin.service.spec.js
GeorgeChatzigiannis: @jon-shipley added

admin/spec/service/pin.service.spec.js
GeorgeChatzigiannis: @jon-shipley thanks I've updated that

admin/spec/service/pin.service.spec.js
GeorgeChatzigiannis: @jon-shipley thanks I've updated that

admin/services/pin.service.js
GeorgeChatzigiannis: @GuyHarwood Generally I noticed it works fine even if I set the date/time through sinon faketimers.
Besides I only need to verify that the pupil update call is fired for my current tests in case the user account is non test one.
Also so far we currently only use the date service for formatting purposes.

GuyHarwood: If moment is working correctly with sinon then that’s ok 👍

pupil-spa/src/app/services/audit/auditEntry.ts
GeorgeChatzigiannis: @jon-shipley that was in the ticket's AC from Guy.

admin/app.js
GeorgeChatzigiannis: @GuyHarwood yea it is outside of the scope of the ticket anyway. we'll correct this on the new app.

admin/controllers/check-started.js
GeorgeChatzigiannis: @GuyHarwood changed

admin/models/check.js
GeorgeChatzigiannis: On the commit 6753b94df112d59e2f8c74810e3c513853d7fc5a you can see the change under design/schemas/database/checks.json
Did I miss something?

admin/services/check-start.service.js
GeorgeChatzigiannis: @updated to use moment. As stated earlier date service is used for formatting so it shouldn't be required.

pupil-spa/src/app/services/audit/auditEntry.ts
GeorgeChatzigiannis: @GuyHarwood updated

pupil-spa/src/app/services/submission/submission.service.spec.ts
GeorgeChatzigiannis: @GuyHarwood ok that's better, we'll definitely tackle this at that stage.

admin/spec/controllers/check-started.spec.js
GeorgeChatzigiannis: @jon-shipley updated

pupil-spa/src/app/services/submission/submission.service.ts
GeorgeChatzigiannis: @GuyHarwood error should not be thrown on the UI as per this commit 11cb144

GeorgeChatzigiannis:
@jon-shipley @GuyHarwood as per this commit e294156 it will now attempt the call 100 times before finally continuing running. We can make this number configurable and change it obviously if needed.

pupil-spa/src/app/services/submission/submission.service.ts
```diff
@@ -25,7 +25,13 @@ export class SubmissionService {
           }
         },
         (err) => { throw new Error(err); }
-      ).catch((err) => { throw new Error(err); });
+      ).catch((err) => {
```
GuyHarwood: imagine this code being executed by a class of 30 pupils.  It's like a mini Denial Of Service attack.

Please familiarise with the circuit breaker pattern, and implement it here.

You need to take into consideration how long you should wait before trying, and how long you should keep trying for, as they may reach the end of the check before it succeeds.  If they get to the end of the check, then its pointless to continue trying this.

jon-shipley: I think whats needed here is a service that act as a wrapper around a remote call.  Then it will be trivial to re-use it for other calls: authentication / check submission too.

GeorgeChatzigiannis: @GuyHarwood @jon-shipley there was an agreement to treat this as a standalone feature for all the api calls within the app.

GuyHarwood: @GeorgeChatzigiannis No.  This code cannot be added as-is.  It's not a separate feature, its a requirement of the check start notification.  Yes, it would be good to re-use elsewhere, but that's just an added benefit of a good implementation.

pupil-spa/src/app/services/submission/submission.service.ts
```diff
@@ -21,7 +24,11 @@ export class SubmissionService {
     await this.http.post(`${environment.apiURL}/api/check-started`,
       { checkCode, accessToken },
       requestArgs)
-      .retryWhen(errors => errors.delay(apiErrorDelay).take(apiErrorMaxAttempts))
+      .retryWhen((errors) => {
+        return errors.mergeMap((error) =>
+          (error.status !== 503 && error.status !== 504) ? Observable.throw(error) : Observable.of(error))
```
GuyHarwood: looks good.  Please remove status code filtering, and send each error to the audit log, if any occur.

jon-shipley: @GuyHarwood - why would you retry in all cases?  Please explain.

GuyHarwood: @jon-shipley because such a narrow range of status codes assumes these are the only situations in which we would want to try again.  Let's say that we miss a required env var on deployment, and the server is throwing 500s.  We correct it, and seconds later the app is up and running.  There should be opportunity to recover from the majority of unanticipated situations, not the few.

PR #288, opened 2017-11-20
Security cleanup

jon-shipley:
- Remove gulp-standard: it's advised to remove this by snyk.io and we are not using it anyway.  Use `standardjs` itself.


admin/package.json
```diff
@@ -4,13 +4,14 @@
   "private": false,
   "license": "GPL-3.0",
   "scripts": {
+    "coverage": "nyc -r text -r html ./run-jasmine.js",
+    "lint": "standard",
+    "postinstall": "gulp build",
     "prestart": "cd data/migrations && ../../node_modules/.bin/migrate-mongo up",
+    "sasswatch": "gulp watch",
     "start": "node ./server.js",
```
GeorgeChatzigiannis: @jon-shipley since you are updating these can we make start script to run both server js and the gulp watch command. I don't think there's a point of having them separately really.
We are already doing this on the pupil-spa so I'd rather remove the sasswatch command here as well.

jon-shipley: @GeorgeChatzigiannis  - would changing the start script have any knock-on effects on the docker containers or production environment though?

All I have done here is to add the lint task and sort the lines.  It looks like I have done more than I have.

GeorgeChatzigiannis: @jon-shipley well the script command remains the same really. only the functionality behind the command is populated.

PR #290, opened 2017-11-21
Feature/security update mongoose

jon-shipley:
Update out of date packages.


admin/app.js
```diff
@@ -45,17 +41,26 @@ if (unsetVars.length > 0) {
...
+  keepAlive: true,
+  reconnectTries: 120,
+  // set the delay between every retry (milliseconds)
+  reconnectInterval: 1000,
+  useMongoClient: true
 })
+
 autoIncrement.initialize(mongoose.connection)
 
+if (process.env.NODE_ENV !== 'production') {
```
GuyHarwood: we already have this on line 45!

admin/spec/models/completed-checks.spec.js
```diff
@@ -24,7 +24,6 @@ describe('CompletedChecks model', () => {
     check.receivedByServerAt = undefined
     check.validate(err => {
       expect(err).toBeDefined()
-      expect(err.message).toBe('completedChecks validation failed')
```
GuyHarwood: how come this was removed?  seems out of context 🤔

GuyHarwood:
@jon-shipley are we able to update the handlebars dependency in the SPA?  as we are on a low version and github keeps reporting a vulnerability...

https://github.com/DFEAGILEDEVOPS/MTC/network/dependencies#606634

admin/spec/models/completed-checks.spec.js
jon-shipley: The expectation was brittle.  That is not an error message out app produces - rather it is one produced from the mongoose library, which changed when I upgraded the mongoose package to latest.

It seems sufficient that it checks that there is an error caused by the `receivedByServerAt` field on the next line.

jon-shipley:
The handlebars vulnerability is fixed in commit 6fa110c.

admin/app.js
jon-shipley: Oops! Now fixed.

admin/spec/models/completed-checks.spec.js
GuyHarwood: makes sense to me 👍

PR #291, opened 2017-11-21
Feature/17605 check forms refactoring

pris54:
Core refactoring of check forms feature.
Further test coverage in the next PR
Further refactoring (check form page and assign forms to check windows) in the following PRs.


GuyHarwood:
This work is an intermediate step towards refactoring the existing check forms work, which was largely untested.  The following PRs will increase coverage.

In future, any existing features that need refactoring will be started from scratch - test first.  That way any 'old' code that is valid for re-use can be introduced once the tests are already in place.

PR #292, opened 2017-11-21
update rules and guidance

GuyHarwood:
Please read the updated notes as they are critical going forward.

All PRs should be created when you first start writing your tests for the new/refactored feature.

The updated document outlines how to approach this when refactoring legacy parts of the system.


contributing.md
```diff
@@ -7,6 +7,22 @@ The MTC team is following a set of patterns and practices throughout the develop
...
 - Bug branch naming onventions are similar to feature branches: `bug/(ticket number)-(description)`
 
+## Workflow expectations
+
+All work has to be done TDD style - tests first.  Create a PR once you have written a few tests, and invite team members to review.  This gives a clear indiciation of what the feature is expected to do, and shows you are practising TDD.  This is easy for new features, but a little more involved when refactoring existing work that has no coverage.
+
+### Refactoring approach
+
+As an example, lets say you are refactoring an existing controller and the services it depends on.
+1.  Create a new spec for the controller and write some tests that assert the expectation of the controller (returns 200 OK, has x,y in the body etc)
```
jon-shipley: I think the point of TDD is to keep the cycle time short between writing code, writing tests and running the test suite.  Writing more than one test at a time makes the cycle time longer, which I don't think is good advice.

Useful tools (new section) 

It may be useful to use [madge](https://www.npmjs.com/package/madge) for existing modules to view the dependency graph.  This may help in quickly identifying older code, that should be reviewed to see if needs to be refactored.

I would also point out that as our dependencies follow the flow-of-control the easiest way to create small Pull Requests when refactoring existing code is to start any refactoring at the lowest level first and work up to the controller.

GuyHarwood: @jon-shipley each test should pass before moving onto the next one, otherwise it creates all sorts of problems.  I will make that clear, as it's obviously reading differently to the intention.

Will add the madge section too.

GuyHarwood: @jon-shipley regarding the flow of control and bottom-up approach...

top-down works best on a blank canvas, as your dependencies emerge through the tests you write against the SUT.  I have suggested that we leave the old code alone, and start from scratch with replacement components, pulling in reusable code as and when necessary.

If the refactoring is just modifications or enhancements to code that's already tested, then your suggestion works.  But when replacing an older untested version (such as the recent pin generation work), it's proven to be hard work to try and refactor the feature in-place.  Hence why I suggest the blank canvas approach and replacement components.

jon-shipley: @GuyHarwood - yes.  Perhaps I didn't make myself clear - the bottom-up approach was suggested for refactoring existing code.  For now, the data-access layer is still being completed so any new work will inevitably require checking the data access layer.

This shouldn't  affect the controllers though.

PR #300, opened 2017-11-22
17592  / Update confirmation message and refactoring

jon-shipley:
feature/17592-errors-in-confirmation-message

* update copy in the flash message after adding a new pupil
* refactoring 
   - pupil controller refactored to be more testable
   - getAddPupil postAddPupil refactored
   - ValidationError made easier to test
   - new pupilDataService.save() method
  - new DateService. createFromDayMonthYear() method that is useful when creating dates from GDS input boxes (day, month, year from user input)
  - new pupilAddService


admin/controllers/pupil.js
```diff
@@ -76,15 +76,42 @@ const postAddPupil = async (req, res, next) => {
...
+    req.flash('info', '1 new pupil has been added')
   } catch (error) {
     next(error)
   }
   const pupilId = JSON.stringify([pupil._id])
   res.redirect(`/school/pupil-register/lastName/true?hl=${pupilId}`)
 }
 
-const getAddMultiplePupils = (req, res, next) => {
+controller.postAddPupil2 = async (req, res, next) => {
```
GeorgeChatzigiannis: is that a temporary naming?

admin/spec/controllers/pupil.spec.js
```diff
@@ -445,4 +456,68 @@ describe('pupil controller:', () => {
...
+
+      it('calls pupilAddService to add a new pupil to the database', async (done) => {
+        await controller(req, res, nextSpy)
+        expect(pupilAddServiceSpy.callCount).toBe(1)
+        done()
+      })
+
+      it('redirects to the pupil register page', async (done) => {
+        await controller(req, res, nextSpy)
+        expect(res.statusCode).toBe(302)
```
GeorgeChatzigiannis: that's just a 302 code where's the redirection call?
Perhaps also check for the page title to be what you expect?

admin/controllers/pupil.js
jon-shipley: yes

admin/controllers/pupil.js
```diff
@@ -10,81 +12,64 @@ const pupilService = require('../services/pupil.service')
 const pupilDataService = require('../services/data-access/pupil.data.service')
 const azureFileDataService = require('../services/data-access/azure-file.data.service')
 const pupilUploadService = require('../services/pupil-upload.service')
+const schoolDataService = require('../services/data-access/school.data.service')
+const pupilAddService = require('../services/pupil-add-service')
+
+const controller = {}
 
-const getAddPupil = async (req, res, next) => {
+controller.getAddPupil = async (req, res, next, error = null) => {
```
pris54: Are we changing the naming convention? This will require some refactoring of existing controllers.

jon-shipley: It's still the same name.  It was necessary to change the the exports for unit testing - as it was it was impossible to stub the getAddPupil from the postAddPupil.

I'd suggest this is a more testable pattern.

jon-shipley: The call signature was updated to allow the error page to displayed, as well as the default initial page.

PR #301, opened 2017-11-23
Feature/17605 check forms unit tests

pris54:
Further test coverage for check forms.

NOTE! Some of the tests for the check-form controller are commented out, they fail due to an error caused by mongoose-auto-increment, this plugin will soon be replaced with a different solution.


admin/controllers/check-form.js
```diff
@@ -68,15 +68,15 @@ const uploadAndViewForms = async (req, res, next) => {
 const removeCheckForm = async (req, res, next) => {
   const id = req.params.formId
   try {
-    const CheckWindow = await checkFormDataService.getActiveForm(id)
-    if (!CheckWindow) {
+    const checkForm = await checkFormDataService.getActiveForm(id)
```
pris54: CheckWindow was misleading, it's a form that we are getting here.

admin/controllers/check-form.js
```diff
@@ -68,15 +68,15 @@ const uploadAndViewForms = async (req, res, next) => {
...
+    if (!checkForm) {
       return next(new Error(`Unable to find form.id [${id}]`))
```
GuyHarwood: could you please change this to `Unable to find check form with id [${id}]`

i appreciate you may not have written it, but it will aid us with potential debugging

admin/models/check-form.js
```diff
@@ -37,21 +37,16 @@ const CheckForm = new Schema({
...
 }, {timestamps: true})
 
-// Make the `id` an incrementing number
-CheckForm.plugin(autoIncrement.plugin, { model: 'CheckForm', startAt: 1 })
-
-CheckForm.pre('save', function (next) {
+CheckForm.pre('save', async function (next) {
   if (!this._id) {
-    throw new Error('Missing _id field')
+    // @TODO: There may be a more efficient way to solve the auto-increment.
```
GuyHarwood: this could easily be the victim of a race condition.  It's quite hard to solve without a table lock.

i suggest we leave the old code as-is until we move to SQL, as i have already solved this by using auto increment for the id field in the checkForm table.

admin/controllers/check-form.js
pris54: Done, it will be in the next push.

admin/models/check-form.js
pris54: OK, I'll leave it as it is for now. I'll update the comment.

PR #306, opened 2017-11-23
Feature/17707 Restart form list

GeorgeChatzigiannis:
UI elements for both reason and pupil restart list


admin/services/restart.service.js
```diff
@@ -0,0 +1,39 @@
...
+ * @returns {Array}
+ */
+restartService.getPupils = async (schoolId) => {
+  const school = await schoolDataService.findOne({_id: schoolId})
+  if (!school) throw new Error(`School [${schoolId}] not found`)
+  let pupils = await pupilDataService.getSortedPupils(schoolId, 'lastName', 'asc')
+  pupils = pupils
+    .filter(p => restartService.isEligiblePupil(p))
+    .map(({ _id, pin, dob, foreName, middleNames, lastName }) =>
+      ({ _id, pin, dob: moment(dob).format('DD MMM YYYY'), foreName, middleNames, lastName }))
```
jon-shipley: Could you use the dateService to format the date of birth?

admin/services/restart.service.js
```diff
@@ -0,0 +1,39 @@
...
+  pupils = pupilIdentificationFlagService.addIdentificationFlags(pupils)
+  return pupils
+}
+
+/**
+ * Determine if pupil is eligible for restart
+ * @param p
+ * @returns {Boolean}
+ */
+restartService.isEligiblePupil = (p) => {
```
jon-shipley: I think `canRestart` is clearer english.  Or if you must how about `isPupilEligible()`?

admin/services/restart.service.js
```diff
@@ -0,0 +1,39 @@
...
+  // This is undergoing changes as the count won't be attached to the pupil collection
+  let restartCount = p.restartCount || 0
+  if (restartCount >= 2 || p.attendanceCode) return false
+  const hasExpiredToday = p.pinExpiresAt && moment(p.pinExpiresAt).isSame(moment.now(), 'day')
```
jon-shipley: If the pupil pin expires today, and this is called before the pin expiry time, then won't `hasExpiredToday` be true even though the pupil pin is still valid?

admin/views/restart/select-restart-list.ejs
```diff
@@ -6,11 +6,164 @@
...
+                        <div class="panel panel-border-wide">
+                            Please specify:
+                            <textarea class="form-control form-control-3-4" name="did-not-complete-textarea" id="did-not-complete-textarea" rows="1" maxlength="100"></textarea>
+                        </div>
+                    </div>
+                </div>
+            </li>
+        </ul>
+
+        <!--<br />-->
```
jon-shipley: still required?  Or can this be deleted?

admin/services/restart.service.js
GeorgeChatzigiannis: But this boolean is only there to verify that the of the pinExpiration is today.
The check on the line below will determine based on today the result from the validation of the pin.

admin/views/restart/select-restart-list.ejs
GeorgeChatzigiannis: removed

admin/services/restart.service.js
GeorgeChatzigiannis: @jon-shipley Updated

admin/services/restart.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/assets/javascripts/global-scripts.js
```diff
@@ -195,6 +221,14 @@ $(function () {
     inputStatus.radioStatus('attendanceCode', pupilsNotTakingCheck.validateForm)
   }
 
+  if ($('#pupilsRestartList').length > 0) {
+    inputStatus.toggleAllCheckboxes('#pupilsRestartList', restarts.validateForm)
+    inputStatus.selectAll('.multiple-choice-mtc')
+    inputStatus.deselectAll('.multiple-choice-mtc')
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley as per @pris54 this is required to expose these methods for a specific page.

admin/spec/service/restart.service.spec.js
```diff
@@ -0,0 +1,67 @@
...
+        result = await restartService.getPupils(schoolMock._id)
+      } catch (error) {
+        expect(error).toBeUndefined()
+      }
+      expect(result.length).toBe(2)
+    })
+  })
+
+  describe('isPupilEligible', () => {
+    it('it should return false if pupil has been restarted exactly 2 times', () => {
```
jon-shipley: What happens if the pupil has restarted a weirdly large number of times, like 5?

GeorgeChatzigiannis: @jon-shipley I have temporary disabled this test and removed the logic for restart counts on pupil as this is changing significantly as part of the restart submission work that is currently undergoing.

GeorgeChatzigiannis: There should never be a case where the pupil restart collection will have more than 2 restart occurances for a pupil Id based on that work.

admin/services/restart.service.js
```diff
@@ -0,0 +1,39 @@
...
+    .map(({ _id, pin, dob, foreName, middleNames, lastName }) =>
+      ({ _id, pin, dob: () => dateService.formatLongGdsDate(dob), foreName, middleNames, lastName }))
+  pupils = pupilIdentificationFlagService.addIdentificationFlags(pupils)
+  return pupils
+}
+
+/**
+ * Determine if pupil is eligible for restart
+ * @param p
+ * @returns {Boolean}
```
jon-shipley: {boolean}

PR #312, opened 2017-11-29
Feature/17604 check forms refactor form page

pris54:


admin/controllers/check-form.js
```diff
@@ -194,47 +194,36 @@ const saveCheckForm = async (req, res, next) => {
...
-//     next(error)
-//   }
-// }
+const displayCheckForm = async (req, res, next) => {
+  req.breadcrumbs('Upload and view forms', '/test-developer/upload-and-view-forms')
+  res.locals.pageTitle = 'View form'
+
+  let formData
+
+  formData = await checkFormDataService.getActiveFormPlain(req.params.formId)
```
jon-shipley: As this block has two async methods should we wrap it in a try/catch so that if the promise is rejected it won't go unhandled?

Node keeps nagging me that in the future unhandled promises will halt execution.

admin/services/check-form.service.js
```diff
@@ -126,6 +126,24 @@ const checkFormService = {
...
+   * @returns {Promise.<void>}
+   */
+  objectFormattingForSinglePage: async (checkWindows) => {
+    let data = {
+      checkWindowsName: [],
+      canDelete: true
+    }
+    checkWindows.forEach(cw => {
+      data.checkWindowsName.push(' ' + cw.checkWindowName)
+      if (cw.checkStartDate <= Date.now()) {
```
jon-shipley: It seems a little weird to have business logic in a formatting function.

admin/spec/controllers/check-form.spec.js
```diff
@@ -139,17 +139,17 @@ describe('check-form controller:', () => {
...
-        checkForm.save = () => {}
-
         spyOn(checkFormService, 'populateFromFile').and.returnValue(checkFormMock)
         spyOn(CheckForm, 'create').and.returnValue(Promise.resolve(checkFormMock))
         spyOn(fs, 'remove').and.returnValue(checkFormMock)
         controller = proxyquire('../../controllers/check-form', {}).saveCheckForm
       })
 
       it('should save the form and redirect the user', async (done) => {
+        const checkForm = new CheckForm()
```
jon-shipley: If this has been refactored we should not have mongoose models in controllers.  They should be strictly in the data-access layer.

admin/spec/service/check-form.service.spec.js
```diff
@@ -129,4 +129,21 @@ describe('check-form.service', () => {
       done()
     })
   })
+
+  describe('#unassignedCheckFormsFromCheckWindows()', () => {
+    beforeEach(() => {
+      service = proxyquire('../../services/check-form.service', {
+        '../../services/check-form.service': checkFormService
```
jon-shipley: This looks ... odd.  It has itself as a dependency?

admin/views/test-developer/view-check-form.ejs
```diff
@@ -18,51 +7,61 @@ let num = 1;
...
-            </dd>
-            <% }); %>
-        </dl>
+    <div class="column-one-third add-title-margin">
+        <aside role="complementary" class="support-column">
+            <h3 class="heading-medium">Related</h3>
+            <nav role="navigation">
+                <ul class="list">
+                    <li>
+                        <a href="/PDFs/MTC_administration_guidance_June-2017-trial.pdf" target="_blank">Guidance</a>
```
jon-shipley: Add a rel="noopener noreferrer"  attribute to any links that use  target="_blank" . This protects users against security vulnerabilities associated with the target="_blank"  attribute.

https://paper.dropbox.com/doc/Links-xiWNcJk5Wpb8Sj6G4Prtl

admin/services/check-form.service.js
```diff
@@ -126,6 +126,24 @@ const checkFormService = {
       const promises = modifiedCheckWindows.map(cw => { cw.save() })
       return Promise.all(promises)
     }
+  },
+  /**
+   * Append check windows name(s) and canDelete to check form object.
+   * @param checkWindows
+   * @returns {Promise.<void>}
+   */
+  objectFormattingForSinglePage: async (checkWindows) => {
```
jon-shipley: The name of this method still leaves me wondering what it does.

admin/spec/service/check-form.service.spec.js
GeorgeChatzigiannis: That doesn't seem like a case where you would use proxyquire indeed, simple require would do.

admin/services/check-form.service.js
GeorgeChatzigiannis: You would also want use moment.utc() here instead

admin/spec/controllers/check-form.spec.js
GuyHarwood: yes.  we are about to start the data layer refactoring and it will be much harder if there are mongoose dependencies in controllers and services

admin/services/check-form.service.js
pris54: @jon-shipley It deals with the logic to produce two items of this object, I don't know how to call it! I'll think of something better

admin/views/test-developer/view-check-form.ejs
pris54: :thumbsup:

admin/services/check-form.service.js
jon-shipley: @pris54 - there is a code smell here.  You have 'and' in the jsdoc description...

I feel that there should be a canDelete() method that only determines whether the checkForm can be deleted.  If you refactored this maybe the remainder could be moved elsewhere.

jon-shipley: Also - this method is marked as `async` but I think it is synchronous isn't it?

pris54: @jon-shipley I'm splitting the method into two for clarity, one will produce the window name(s), the other one outputs a true or false for 'canDelete'

admin/controllers/check-form.js
```diff
@@ -194,47 +194,48 @@ const saveCheckForm = async (req, res, next) => {
...
+  }
+
+  try {
+    checkWindows = await checkWindowService.getCheckWindowsAssignedToForms()
+  } catch (error) {
+    console.log(`Unable to find check window(s) for active check form: ${error.message}`)
+  }
+
+  if (checkWindows[req.params.formId]) {
+    formData.checkWindowsName = checkFormService.checkWindowNames(checkWindows[req.params.formId])
```
jon-shipley: Is this singular or plural?

singular: checkWindowName
plural: checkWindowNames

pris54: Yeah that's a bad wording. Fixed.

admin/controllers/check-form.js
```diff
@@ -194,47 +194,48 @@ const saveCheckForm = async (req, res, next) => {
...
+  res.locals.pageTitle = 'View form'
+
+  let formData
+  let checkWindows
+
+  try {
+    formData = await checkFormDataService.getActiveFormPlain(req.params.formId)
+    formData.checkWindowsName = []
+    formData.canDelete = true
+  } catch (error) {
```
GeorgeChatzigiannis: error is being passed down as an argument but is not really used in here

admin/services/check-form.service.js
```diff
@@ -126,6 +126,32 @@ const checkFormService = {
...
+      checkWindowsName.push(' ' + cw.checkWindowName)
+    })
+    return checkWindowsName
+  },
+  /**
+   * Append check windows name(s) and canDelete to check form object.
+   * @param checkWindows
+   * @returns {Promise.<void>}
+   */
+  canDelete: (checkWindows) => {
```
GeorgeChatzigiannis: is the jsdoc description right here?

admin/controllers/check-form.js
jon-shipley: If we are console logging we should also log the actual error.message too.  We will convert these logs to use winston later.

admin/services/check-form.service.js
jon-shipley: @pris54 please update the jsdoc description

admin/services/check-form.service.js
```diff
@@ -126,6 +126,32 @@ const checkFormService = {
...
+    let canDelete
```
jon-shipley: I'd prefer to default to false, and change it to true on the basis of a check.  This looks like it might be returning, false, or undefined.

Also it does not return a Promise, rather a boolean

pris54:
@jon-shipley @guy Can I please get an approval? All the recommendation have been done.

PR #319, opened 2017-11-30
Feature/17711 restart submission

GeorgeChatzigiannis:
- Pupil restart eligibility check
- pin expiration for pupils with existing pin or pinExpiresAt fields
- new pupil restarts collection which stores each restart record
- redirect to overview page on success with message highlight


admin/controllers/restart.js
```diff
@@ -20,11 +24,47 @@ const getSelectRestartList = async (req, res, next) => {
...
-  getRestartOverview,
-  getSelectRestartList
+controller.postSubmitRestartList = async (req, res, next) => {
+  const { pupil: pupilsList, restartReason, didNotCompleteInfo, restartFurtherInfo } = req.body
+  if (!pupilsList) {
+    return res.redirect('/restart/select-restart-list')
+  }
+  const validationError = restartValidator.validateReason(restartReason, didNotCompleteInfo)
+  if (validationError.hasError()) {
+    res.locals.pageTitle = 'Select pupils for restart'
```
pris54: 'Error: ' needs to be a prefix of the page title - we've been missing this AC but should start implementing it ASAP.
"Additionally, you should reflect that there's been an error in the <title> by prefixing the existing title with 'Error: '".
https://govuk-elements.herokuapp.com/errors/

admin/routes/restart.js
```diff
@@ -5,10 +5,12 @@ const router = express.Router()
 
 const {
   getRestartOverview,
-  getSelectRestartList
+  getSelectRestartList,
+  postSubmitRestartList
 } = require('../controllers/restart')
```
pris54: I thought we agreed we would require the whole of the controller and then go for a naming convention [controller].[method]

GeorgeChatzigiannis: For the purpose of routes I don't see it making any difference, we are not testing them after all.

GeorgeChatzigiannis: The controller however has changed.

design/schemas/database/pupilrestarts.json
```diff
@@ -0,0 +1,12 @@
...
+  "pupilId": "59cd18a731cb2f71e0bcc64f",
+  "recordedByUser": "Name",
+  "createdAt": "2017-09-28T15:44:00.550Z",
+  "reason": "Did not complete",
+  "didNotCompleteInformation": "The reason",
+  "furtherInformation": "Further reason",
+  "isDeleted": true,
+  "deletedByUser": "Name",
+  "deletedAt": "2017-09-28T15:44:00.550Z"
+}
```
GuyHarwood: newline

admin/models/pupil-restart.js
```diff
@@ -0,0 +1,41 @@
...
+mongoose.Promise = global.Promise
+const Schema = mongoose.Schema
+
+const PupilRestart = new Schema({
+  pupilId: {
+    type: Schema.Types.ObjectId,
+    required: true,
+    ref: 'Pupil'
+  },
+  recordedByUser: {
```
GuyHarwood: just so the others are aware - this will store the username.  Its a temporary fixture until we move to SQL and we will then use formal foreign keys by unique id

admin/models/pupil-restart.js
```diff
@@ -0,0 +1,41 @@
...
+  createdAt: {
+    type: Date
+  },
+  isDeleted: {
+    type: Boolean,
+    default: false,
+    required: true
+  },
+  deletedBy: {
+    type: String
```
GuyHarwood: just so the others are aware - this will store the username. Its a temporary fixture until we move to SQL and we will then use formal foreign keys by unique id

admin/routes/restart.js
```diff
@@ -5,10 +5,12 @@ const router = express.Router()
...
 
 router.get('/overview', isAuthenticated(config.ROLE_TEACHER), (req, res) => getRestartOverview(req, res))
```
GuyHarwood: overview seems a bit vague - can we go with /restarts/overview or /restarts-overview please

admin/routes/restart.js
```diff
@@ -5,10 +5,12 @@ const router = express.Router()
...
 router.get('/select-restart-list', isAuthenticated(config.ROLE_TEACHER), (req, res, next) => getSelectRestartList(req, res, next))
+router.post('/submit-restart-list', isAuthenticated(config.ROLE_TEACHER), (req, res, next) => postSubmitRestartList(req, res, next))
```
GuyHarwood: would it be better convention to prefix with the feature, so it would be /restart-submit-list etc? 🤔

admin/services/pin.service.js
```diff
@@ -51,4 +51,25 @@ pinService.expirePupilPin = async (token, checkCode) => {
...
+    const pupil = await pupilDataService.findOne({_id: ObjectId(id)})
+    if (pupil.pin || pupil.pinExpiresAt) pupils.push(pupil)
+  }
+  if (pupils.length === 0) return
+  pupils = pupils.map(p => {
+    p.pin = null
+    p.pinExpiresAt = null
+    return p
+  })
+  await pupilDataService.updateMultiple(pupils)
```
pris54: Should this not be simply ```return pupilDataService.updateMultiple(pupils)``` - as we always want a function to return something? - I'm looking at ESLint for guidance:
https://eslint.org/docs/rules/no-return-await

admin/routes/restart.js
pris54: It's a naming convention we are implementing everywhere else, so I believe it would be beneficial also for the sake of consistency, plus you will most likely need all the controller methods for the router anyway.

admin/services/pin.service.js
GeorgeChatzigiannis: The difference here is that the caller does not need to capture anything from this method.
The rule you linked is relevant to `return await` which I don't need to do anyway as the method will be behaving like a void.

admin/routes/restart.js
GeorgeChatzigiannis: Personally I think it's extra work across all routes (it's a little though I know) for no benefit.
Don't all routes at the moment use the exact same pattern?
I would agree that destructuring makes testing harder in cases like the controllers but overall we shouldn't exclude the assignment as it improves readability in some cases.
What's the opinion of the rest on this? @GuyHarwood @jon-shipley

admin/controllers/restart.js
GeorgeChatzigiannis: The problem with GDS here is that the don't really provide an example on this.
Do we need to prefix the page title or the error title?

admin/routes/restart.js
GuyHarwood: I would rather it be consistent, everywhere.

GeorgeChatzigiannis: @GuyHarwood updated.
We should also create a subtask in tech debt PBI to update all the routes

admin/config.js
```diff
@@ -8,10 +8,11 @@ module.exports = {
   NCA_TOOLS_AUTH_URL: process.env.NCA_TOOLS_AUTH_URL,
   NEW_RELIC_LICENSE_KEY: process.env.NEW_RELIC_LICENSE_KEY,
   PORT: process.env.PORT || '3001',
+  PUPIL_APP_URL: process.env.PUPIL_APP_URL,
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @GuyHarwood I put them in alphabetical order.

admin/controllers/restart.js
```diff
@@ -1,14 +1,18 @@
 const restartService = require('../services/restart.service')
+const restartValidator = require('../lib/validator/restart-validator')
+const ValidationError = require('../lib/validation-error')
 
-const getRestartOverview = async (req, res) => {
+const controller = {}
+
+controller.getRestartOverview = async (req, res) => {
```
jon-shipley: Is this really async?

admin/models/pupil-restart.js
GeorgeChatzigiannis: @GuyHarwood that property has changed to 'deletedByUser'

admin/lib/validator/restart-validator.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Validate restart reason
+ * @param reason
+ * @param specifyReason
+ * @returns {Boolean}
+ */
+restartValidator.validateReason = (reason, specifyReason) => {
+  const validationError = new ValidationError()
+  if (reason === 'Did not complete' && specifyReason.length === 0) {
+    validationError.addError('didNotCompleteInfo', 'Error: Please specify further information when "Did not complete" option is selected')
```
jon-shipley: We have usually put error messages in their own file.  This is useful, should a content review be required.  Suggest you consider it.

admin/routes/restart.js
GeorgeChatzigiannis: @GuyHarwood on app.js I prefix all these routes with /restart
line 221: app.use('/restart', restart)
So essentially this is going to be /restart/overview

admin/routes/restart.js
GeorgeChatzigiannis: @GuyHarwood as pointed on the previous comment it's going to be:
/restart/submit-restart-list

design/schemas/database/pupilrestarts.json
GeorgeChatzigiannis: @GuyHarwood added

admin/models/pupil-restart.js
GuyHarwood: 👍

admin/services/pin.service.js
jon-shipley: Yes `return pupilDataService.updateMultiple(pupils)` is correct

admin/config.js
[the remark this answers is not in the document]
GuyHarwood: 👍 good move

admin/routes/restart.js
GuyHarwood: @GeorgeChatzigiannis PBI created under tech debt.

admin/services/restart.service.js
```diff
@@ -16,24 +24,81 @@ restartService.getPupils = async (schoolId) => {
...
+ * @param didNotCompleteInfo
+ * @param restartFurtherInfo
+ * @param userId
+ * @returns {Promise.<void>}
+ */
+
+restartService.restart = async (pupilsList, restartReason, didNotCompleteInfo, restartFurtherInfo, userName) => {
+  await pinService.expireMultiplePins(pupilsList)
+  // All pupils should be eligible for restart before proceeding with creating a restart record for each one
+  const eligibilityList = await Promise.all(pupilsList.map(async pupilId => {
```
jon-shipley: It might be useful if this functionality was extracted into it's own function: `canAllPupilsRestart(pupilList)` returning bool.

admin/controllers/restart.js
GeorgeChatzigiannis: @jon-shipley nope bad paste from me and good catch from you

admin/services/pin.service.js
GeorgeChatzigiannis: @jon-shipley @pris54 even though as I said the caller does not need it, I will change the await to return to avoid confusion

admin/services/restart.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/lib/validator/restart-validator.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Validate restart reason
+ * @param reason
+ * @param specifyReason
+ * @returns {Boolean}
+ */
+restartValidator.validateReason = (reason, specifyReason) => {
+  const validationError = new ValidationError()
+  if (reason === 'Did not complete' && specifyReason.length === 0) {
+    validationError.addError('didNotCompleteInfo', 'Please specify further information when "Did not complete" option is selected')
```
jon-shipley: This error message should be coming from an external file please.

admin/services/data-access/pupil-restart.data.service.js
```diff
@@ -0,0 +1,25 @@
...
+const pupilRestartDataService = {}
+
+/**
+ * Create a new Pupil restart
+ * @param data
+ * @return {Promise}
+ */
+pupilRestartDataService.create = async function (data) {
+  const pupilRestart = new PupilRestart(data)
+  return pupilRestart.save()
```
jon-shipley: Actually won't this return the mongoose model?

Should it actually be:
```
await pupilRestart.save()
return pupilRestart.toObject()
```

admin/services/restart.service.js
```diff
@@ -16,24 +24,91 @@ restartService.getPupils = async (schoolId) => {
...
+
+/**
+ * Checks if all selected pupils are eligible for restarts
+ * @param pupilsList
+ * @return {boolean}
+ */
+restartService.canAllPupilsRestart = async (pupilsList) => {
+  const eligibilityList = await Promise.all(pupilsList.map(async pupilId => {
+    const canRestart = await restartService.canRestart(pupilId)
+    if (!canRestart) return { pupilId, notEligible: true }
```
jon-shipley: I wonder if a the property name `notEligible` (for a restart) should actually be positive: e.g. calling it `eligible: false`, or simply {canRestart: <bool>}?

The `!e.notEligible` a few lines down is not particularly clear to me.

admin/services/restart.service.js
```diff
@@ -60,14 +60,10 @@ restartService.isPupilEligible = async (p) => {
...
   // All pupils should be eligible for restart before proceeding with creating a restart record for each one
-  const eligibilityList = await Promise.all(pupilsList.map(async pupilId => {
-    const canRestart = await restartService.canRestart(pupilId)
-    if (!canRestart) return { pupilId, notEligible: true }
-    return { pupilId }
-  }))
-  eligibilityList.forEach(e => {
-    if (e.notEligible) throw new Error(`Pupil ${e.pupilId} is not eligible for a restart`)
-  })
+  const canAllPupilsRestart = restartService.canAllPupilsRestart(pupilsList)
```
jon-shipley: 👍 nice

admin/services/data-access/pupil-restart.data.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/lib/validator/restart-validator.js
GeorgeChatzigiannis: @jon-shipley it is now

admin/services/restart.service.js
```diff
@@ -16,24 +24,90 @@ restartService.getPupils = async (schoolId) => {
...
+ */
+
+restartService.restart = async (pupilsList, restartReason, didNotCompleteInfo, restartFurtherInfo, userName) => {
+  await pinService.expireMultiplePins(pupilsList)
+  // All pupils should be eligible for restart before proceeding with creating a restart record for each one
+  const canAllPupilsRestart = restartService.canAllPupilsRestart(pupilsList)
+  if (!canAllPupilsRestart) {
+    throw new Error(`One of the pupils is not eligible for a restart`)
+  }
+  const submitted = await Promise.all(pupilsList.map(async pupilId => {
```
jon-shipley: This inner function does not seem to need to be async perhaps?

admin/controllers/restart.js
```diff
@@ -60,7 +60,8 @@ controller.postSubmitRestartList = async (req, res, next) => {
   }
   res.locals.pageTitle = 'Restarts'
   req.breadcrumbs(res.locals.pageTitle)
-  req.flash('info', `${submittedPupils.length} new pupils have been submitted for restart`)
+  const restartInfo = submittedPupils.length < 2 ? '1 pupil' : `${submittedPupils.length} pupils`
```
jon-shipley: I think it makes sense to test for equality here: `submittedPupils.length === 1`

0 pupils
1 pupil
2 pupils
...

PR #322, opened 2017-12-01
Data layer refactor of controllers/service-manager

jon-shipley:
Data layer refactor only


admin/controllers/service-manager.js
```diff
@@ -103,17 +102,17 @@ const setUpdateTiming = async (req, res, next) => {
   }
 
   try {
-    await settings.save()
-    let settingsLog = new SettingsLog()
+    await settingDataService.createOrUpdate(settings)
 
+    let settingsLog = {}
```
GeorgeChatzigiannis: does it matter if settingslong is an empty object?

admin/controllers/service-manager.js
```diff
@@ -81,11 +80,11 @@ const setUpdateTiming = async (req, res, next) => {
   res.locals.pageTitle = 'Check settings'
   let settings
 
-  const settingsRecord = await Settings.findOne().exec()
+  const settingsRecord = await settingDataService.findOne()
   if (settingsRecord) {
     settings = settingsRecord
   } else {
-    settings = new Settings()
+    settings = {}
```
GeorgeChatzigiannis: same as the settingsLog is this going to matter if it's an empty object vs undefined?

GeorgeChatzigiannis: I guess it does cause the query will thrown the exception.
Nevermind this comment

jon-shipley: Well it gets set later on...

admin/controllers/service-manager.js
jon-shipley: Not at this point

admin/services/data-access/setting-log.data.service.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
```
GeorgeChatzigiannis: jsdoc missing

PR #326, opened 2017-12-04
Feature/17962 sql driver implementation

GuyHarwood:
# Work in Progress

Abstraction of SQL Server connection pooling into app level object.  
Surfacing of data retrieval and modification methods via sql.service.js

Integration tests to follow.


admin/services/data-access/sql.service.js
```diff
@@ -0,0 +1,57 @@
+'use strict'
+
+const Request = require('tedious').Request
+const connectionService = require('./sql.connection.service')
+
+const sqlService = {}
+
+sqlService.query = (sql, params) => {
```
jon-shipley: Needs jsDoc

admin/services/data-access/sql.service.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    request.on('row', function (cols) {
+      results.push(cols)
+    })
+    con.execSql(request)
+  })
+}
+
+sqlService.modify = async (sql, params) => {
```
jon-shipley: Needs jsDoc. I don't think that it needs to be async as all this function does it to return a Promise.  That is certainly `await`able though.

admin/sql.usage.example.js
```diff
@@ -0,0 +1,27 @@
...
+const sqlConnectionService = require('./services/data-access/sql.connection.service')
+const sqlService = require('./services/data-access/sql.service')
+
+sqlConnectionService.init()
+
+const examples = {}
+
+examples.getData = () => {
+  sqlService.query('SELECT * FROM settings')
+    .then(result => console.log('query settings table result:', result))
```
jon-shipley: I think we need an example using `await` rather than `Promise.then` please.  That's our usual idiom.

admin/sql.usage.example.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+examples.modifyData = () => {
+  const id = { name: 'id', type: TYPES.Int, value: 2 }
+  const updatedAt = { name: 'updatedAt', type: TYPES.DateTime2, value: TYPES.Null }
+  const loadingTimeLimit = { name: 'loadingTimeLimit', type: TYPES.TinyInt, value: 1 }
+  const questionTimeLimit = { name: 'questionTimeLimit', type: TYPES.TinyInt, value: 2 }
+
+  sqlService.modify('INSERT settings (id, updatedAt, loadingTimeLimit, questionTimeLimit) VALUES (@id, @updatedAt, @loadingTimeLimit, @questionTimeLimit)',
```
jon-shipley: This example does nothing with the `rowCount` return value.   This would be a nice addition to make the example more complete.

admin/sql.usage.example.js
```diff
@@ -0,0 +1,27 @@
...
+    id, updatedAt, loadingTimeLimit, questionTimeLimit)
+}
+
+examples.modifyData()
```
jon-shipley: ?? spare

admin/services/data-access/sql.service.js
GuyHarwood: good catch!

PR #328, opened 2017-12-04
Remove controller dependency on Pupil mongoose model

jon-shipley:
Remove Pupil and School dependencies on mongoose models from pupil controller.

Before:
pupil js before

After:
pupil js after

### DIFF
--- a/admin/controllers/pupil.js
+++ b/admin/controllers/pupil.js
@@ -1,8 +1,8 @@
 'use strict'
 
 const moment = require('moment')
+const R = require('ramda')
 const School = require('../models/school')
-const Pupil = require('../models/pupil')
 const errorConverter = require('../lib/error-converter')
 const ValidationError = require('../lib/validation-error')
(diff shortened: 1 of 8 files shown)


admin/controllers/pupil.js
```diff
@@ -88,7 +85,7 @@ controller.getAddMultiplePupils = (req, res, next) => {
 controller.postAddMultiplePupils = async (req, res, next) => {
   let school
   try {
-    school = await School.findOne({_id: req.user.School}).exec()
+    school = await schoolDataService.findOne({_id: req.user.School})
```
GeorgeChatzigiannis: how about adding a method to pupil service for now that all it does is call school and pupil data service while being wrapped in a try catch?

jon-shipley: That's a good call but outside the intended scope of this PR - and there are 8 PRs to tackle just to move all mongoose models into the data layer.

PR #329, opened 2017-12-04
Feature/17904 check forms - form name

pris54:


admin/controllers/check-form.js
```diff
@@ -172,6 +172,18 @@ const saveCheckForm = async (req, res, next) => {
     })
   }
 
+  try {
+    const fileName = await checkFormService.buildFormName(uploadFile.filename)
+    if (!fileName) {
+      req.flash('error', `${fileName} already exists. Rename and upload again.`)
+      return res.redirect('/test-developer/upload-new-form')
+    } else {
```
jon-shipley: We don't need the else part here... can just use the return in the `if` part above a a guard.  This simplifies and clarifies the intent. E.g.

```
if (!fileName) {
      req.flash('error', `${fileName} already exists. Rename and upload again.`)
      return res.redirect('/test-developer/upload-new-form')
} 
checkForm.name = fileName
```

admin/services/check-form.service.js
```diff
@@ -176,6 +176,18 @@ const checkFormService = {
       }
     })
     return canDelete
+  },
+  /**
+   * Build a form name based on the file name.
+   * @param fileName
+   */
+  buildFormName: async (fileName) => {
+    if (!fileName || fileName.length < 5) {
```
jon-shipley: What is this magic number 5? Better to define it as a constant.

admin/services/data-access/check-form.data.service.js
```diff
@@ -48,10 +48,24 @@ const checkFormDataService = {
...
     return checkForm.toObject()
+  },
+
+  /**
+   * Find check form by name.
+   * @param formName
+   */
+  findCheckFormByName: (formName) => {
+    let query = { 'isDeleted': false, 'name': formName }
+    return CheckForm.findOne(query).exec()
```
jon-shipley: This must return a plain javascript object please:
`CheckForm.findOne(query).lean().exec()` would work

pris54:
Thank you for reviewing @jon-shipley - I've done the suggested changes, plus I've fixed a little bug.

admin/services/check-form.service.js
```diff
@@ -176,6 +176,19 @@ const checkFormService = {
       }
     })
     return canDelete
+  },
+  /**
+   * Build a form name based on the file name.
+   * @param fileName
+   */
+  buildFormName: async (fileName) => {
+    const minFileNameSize = 4 // for example: a.csv
```
jon-shipley: As this is a `length` the example filename will have a length of 5, but code disagrees and says 4. Shouldn't they agree?

admin/services/check-form.service.js
```diff
@@ -176,6 +176,19 @@ const checkFormService = {
       }
     })
     return canDelete
+  },
+  /**
+   * Build a form name based on the file name.
+   * @param fileName
```
jon-shipley: This param returns a string, undefined, and false.  Whilst it would be better to have a single return type, at least please document the return types in use.

PR #330, opened 2017-12-04
Feature/17713 submitted restart list

GeorgeChatzigiannis:
- List of pupils with restart status and reason
- Highlight upon successful submission both on page top and row specific


admin/services/restart.service.js
```diff
@@ -110,4 +111,47 @@ restartService.canRestart = async pupilId => {
...
+ * @param schoolId
+ * @returns {Array}
+ */
+
+restartService.getSubmittedRestarts = async schoolId => {
+  let pupils = await pupilDataService.getSortedPupils(schoolId, 'lastName', 'asc')
+  if (!pupils || pupils.length === 0) return []
+  let restarts = []
+  await Promise.all(pupils.map(async p => {
+    const r = await pupilRestartDataService.findOne({ pupilId: p._id, isDeleted: false })
```
jon-shipley: It's worth adding a comment that the findOne is inside the loop because of a bug in the Cosmos MongoDB API.  We will want to fix this later. So a TODO would be correct...

admin/services/restart.service.js
```diff
@@ -110,4 +111,47 @@ restartService.canRestart = async pupilId => {
...
+  return restarts
+}
+
+/**
+ * Get pupil's restart status
+ * @param pupilId
+ * @returns {String}
+ */
+
+restartService.getStatus = async pupilId => {
```
jon-shipley: The service here seems to know a lot about the UI - which is where I presume the message are going.  I would suggest that if we must provide UI elements from a business service (policy) then we should provide two elements:  first a status code, and second the user-friendly message.

@GuyHarwood - what do you think?

admin/controllers/restart.js
```diff
@@ -4,11 +4,25 @@ const ValidationError = require('../lib/validation-error')
...
+  let restarts
+  try {
+    restarts = await restartService.getSubmittedRestarts(req.user.School)
+  } catch (error) {
+    return next(error)
+  }
+  let { hl } = req.query
+  if (hl) {
+    hl = JSON.parse(hl)
+    hl = typeof hl === 'string' ? JSON.parse(hl) : hl
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: it's happening on the add pupil case as well, there's was case where double json parse is required to expose the array.
It had something to do with express but I am not sure if it is required anymore.

GuyHarwood: why JSON instead of a comma delimited query string?
(i know this isn't the first occurrence of this :) )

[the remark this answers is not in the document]
GuyHarwood: the whole 'JSON in query string' seems a little overkill for a set of ids.

Do we really need it?

admin/services/restart.service.js
GeorgeChatzigiannis: updated

admin/services/restart.service.js
GuyHarwood: it does feel like a UI requirement is added at the logic layer. 

although I wonder how much value-add we will get from refactoring it 🤔

on one side it's 'just enough to meet the requirements' and on the other its fulfilling UI reqs at the logic layer.

GeorgeChatzigiannis:
@GuyHarwood @jon-shipley As per our conversation statuses now are fetched from a new collection.

admin/models/restart-code.js
```diff
@@ -0,0 +1,19 @@
...
+mongoose.Promise = global.Promise
+const Schema = mongoose.Schema
+
+const RestartCode = new Schema({
+  status: {
+    type: String,
+    required: true
+  },
+  code: {
+    type: Number,
```
jon-shipley: I was expecting the restart code to be a short 2 or 3 letter code, rather than a sequence.  I'm happy if @GuyHarwood is happy.

The dev benefit to a short code is that is has more meaning, so makes code understandable.  The integer is effectively meaningless, making on-going development that much harder.

e.g.

```
Code: REM      Description: Remove restart
Code: TKN      Description: Restart taken
Code: MAX     Description: Maximum number of restart already taken
```

I'm sure someone can suggest an improvement here.

GuyHarwood: would rather use codes tbh.  

reasons for not taking check are numbers, but ive made some notes about potentially changing that once we move to SQL.

GuyHarwood: @GeorgeChatzigiannis given that the Ids are numeric, lets go with short alpha codes.  they feel superfluous as numbers.

admin/controllers/restart.js
```diff
@@ -14,10 +14,7 @@ controller.getRestartOverview = async (req, res, next) => {
     return next(error)
   }
   let { hl } = req.query
-  if (hl) {
-    hl = JSON.parse(hl)
-    hl = typeof hl === 'string' ? JSON.parse(hl) : hl
-  }
+  hl = hl && hl.split(',')
```
GuyHarwood: as discussed, can you please parse the UUIDs before trusting them in a query

jon-shipley: They get parsed by express as they are standard qs parameters.   From this point on they only get used by the display to match against properties to see if the green bar needs to be shown.

We could positively sanitise them against hex chars but is this necessary?

GuyHarwood: as long as any inputs are sanitised, then i'm happy

GeorgeChatzigiannis: @GuyHarwood updated

admin/models/restart-code.js
GeorgeChatzigiannis: @GuyHarwood @jon-shipley updated

PR #333, opened 2017-12-05
Feature/17979 sql integration tests & migrations

GuyHarwood:
# Work in Progress

- basic integration test
- migration engine and migrations


admin/data/sql/migrations/001.undo.baseline-tables.sql
```diff
@@ -0,0 +1,34 @@
...
+DROP TABLE dbo.adminLogonEvent
+DROP TABLE dbo.[check]
+DROP TABLE dbo.[user]
+DROP TABLE dbo.pupil
+DROP TABLE dbo.checkForm
+DROP TABLE dbo.school
+DROP TABLE dbo.role
+DROP TABLE dbo.checkWindow
+DROP TABLE dbo.attendanceCode
+DROP TABLE dbo.settings
```
GeorgeChatzigiannis: extra line

admin/data/sql/migrations/003.undo.admin-app-user-permissions.sql
```diff
@@ -0,0 +1 @@
+REVOKE SELECT,UPDATE,DELETE,INSERT,EXECUTE,REFERENCES ON schema::DBO to mtcAdminUser;
```
GeorgeChatzigiannis: extra line here too

admin/tests-integration/sql.service.spec.js
```diff
@@ -0,0 +1,87 @@
+'use strict'
+/* global describe beforeEach beforeAll afterEach it expect jasmine spyOn fail */
+
+require('dotenv').config()
+const sql = require('../services/data-access/sql.service')
+const sqlPool = require('../services/data-access/sql.pool.service')
+const Connection = require('tedious').Connection
+const Request = require('tedious').Request
+
+const executeAdminRequest = async (adminConnection, sql) => {
```
jon-shipley: This might not need to be async.  As it returns a Promise it can still be await'ed on.

admin/tests-integration/create-user.sql
```diff
@@ -0,0 +1,10 @@
+-- create db
```
jon-shipley: It is worth marking files as `.ddl.sql` and `.dml.sql` - do we want to keep them separate?

GuyHarwood: one file is a transaction, and naming may start to get over complicated if both types appear in one file.  

I am comfortable with it as-is

admin/tests-integration/sql.service.spec.js
GuyHarwood: good spot 👏 

golden rule - only mark as async if you need to await

admin/tests-integration/sql.service.spec.js
```diff
@@ -0,0 +1,87 @@
...
+  return new Promise(async (resolve, reject) => {
```
GuyHarwood: @jon-shipley i don't think the Promise needs to be async either

admin/tests-integration/create-user.sql
jon-shipley: I'm not sure I was clear.

If we need to migrate a `null` col to `not null` then we (usually) end up with two transactions:

1. Data manipulation to ensure that there is a value present in the db col (INSERT, UPDATE)
2. Data definition change where we run ALTER TABLE

In my mind this would appear as two files, perhaps with a naming scheme that identified them as belonging together.

We can run it all together in one file, but that seems messier to me.

GuyHarwood: I get what you initially meant :)

given your example, I would have them in one file, as an atomic operation.

Happy to go with the wider vote if this makes things clearer for everyone

jon-shipley: ✅ then my question is answered

admin/tests-integration/sql.service.spec.js
```diff
@@ -1,35 +1,46 @@
...
+  beforeAll(() => {
     sqlPool.init()
-    done()
   })
 
-  it('should permit select query with no parameters', async (done) => {
+  it('should permit select query with no parameters', async () => {
+    let settingsRows = await sql.query('SELECT * FROM Settings')
+    expect(settingsRows).toBeDefined()
+    expect(settingsRows.length).toBe(1)
```
GuyHarwood: @jon-shipley @GeorgeChatzigiannis if the expectation fails (for example when `settingsRows.length` equals 2, it hangs until Jasmine times out.

I get the predicted warning: `(node:45871) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.` but nothing else until the jasmine time out.

What am I missing/overlooking here?

admin/tests-integration/sql.service.spec.js
```diff
@@ -0,0 +1,84 @@
...
+
+  it('should omit the version column from returned set', async () => {
+    await sql.query('SELECT * FROM Settings')
+    const actual = await sql.query('SELECT * FROM Settings')
+    expect(actual).toBeDefined()
+    const row = actual[0]
+    expect(row.version).toBeUndefined()
+  })
+
+  fit('should store the timezone offset with the datetime value', async () => {
```
GuyHarwood: @jon-shipley can we have a chat about this on Monday.  Getting some odd results, likely from the way i am using moment.

GuyHarwood:
Mergin as Travis issues are rvm related

PR #344, opened 2017-12-13
bug/18120 - Correct counting. Counting label as variable.

pris54:


admin/test/features/support/page_objects/generate_pupil_pins_page.rb
```diff
@@ -1,3 +1,4 @@
+require_relative 'sticky_banner_module'
```
activemq: There is no need to require the file like this, please rename the sticky_banner_module.rb to sticky_banner_section.rb & also rename the class to StickyBanner. This then allows you to use the section in any page object like: 
section :sticky_banner, SticktBanner, 'identifier'

PR #346, opened 2017-12-15
Feature/18110 pupil outcomes

GeorgeChatzigiannis:
- Refactor score percentage calculation into a separate service
- Created service method to determine pupil's status


admin/services/pupil.service.js
```diff
@@ -73,4 +58,46 @@ pupilService.validatePupil = async (pupil, pupilData) => {
   return true
 }
 
+/**
+ * Fetch pupil's status
+ * @param pupil
+ * @returns {String}
+ */
+pupilService.getStatus = async (pupil) => {
```
GuyHarwood: this (along with the school controller) need breaking up.  This could have gone in a separate service

admin/services/score.service.js
```diff
@@ -0,0 +1,44 @@
+const checkDataService = require('./data-access/check.data.service')
+
+const scoreService = {}
+
```
GuyHarwood: is this a completely new file?  as i added something similar a while back...

admin/data/migrations/migrations/20171215131500-add-pupil-status-code.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+const mongoose = require('mongoose')
```
jon-shipley: @GuyHarwood  - Do we also want to add this migration to the SQL branch?

admin/models/pupil-status-code.js
```diff
@@ -0,0 +1,16 @@
+const mongoose = require('mongoose')
+mongoose.Promise = global.Promise
+const Schema = mongoose.Schema
+
+const PupilStatusCode = new Schema({
+  status: {
```
jon-shipley: This is actually a description.  Do you think the name reflects this?   When we move to SQL I'm more used to fields like this being called `statusDesc`...

admin/services/pupil-identification-flag.service.js
```diff
@@ -10,17 +10,20 @@ pupilIdentificationFlag.addIdentificationFlags = (pupils) => {
     const currentPupil = pupils[ i ]
     const nextPupil = pupils[ i + 1 ]
     if (nextPupil === undefined) return
-    if (currentPupil.foreName === nextPupil.foreName && currentPupil.lastName === nextPupil.lastName &&
+    if (pupilIdentificationFlag.haveEqualFullNames(currentPupil, nextPupil) &&
```
jon-shipley: I like this.  Reads very nicely. 👍

admin/data/migrations/migrations/20171215131500-add-pupil-status-code.js
GuyHarwood: yes.  We discussed this in the friday dev session - this piece of work should be the last with any mongo work, and will require a set of migrations.  You or I can do them with George this week.

admin/services/score.service.js
GeorgeChatzigiannis: let me know which one it is please.

admin/services/pupil.service.js
GeorgeChatzigiannis: I thought about it as well. I'll generate a pupil.status.service and extract it completely.

admin/test/features/pupil_status.feature
```diff
@@ -0,0 +1,51 @@
...
+  Scenario: Pupil status is Pin Generated when PIN is generated and active
+    Given I have generated a pin for a pupil
+    When I am on the Pupil Register page
+    Then I can see the status for the pupil is 'PIN generated'
+
+  @manual
+  Scenario: Pupil status is IN Progress when a pupil is logged in
+    Given I have a pupil who has started a warm up question
+    When I am on the Pupil Register page
+    Then I can see the status for the pupil is 'In Progress'
```
activemq: Can we not automate this?

PR #350, opened 2017-12-19
settings data service: refactor to SQL Server

GuyHarwood:
# Work In Progress
Refactor setting.data.service to use SQL Server

### DIFF
--- a/admin/services/data-access/setting.data.service.js
+++ b/admin/services/data-access/setting.data.service.js
@@ -1,11 +1,24 @@
 'use strict'
 
 const Setting = require('../../models/setting')
+const sqlService = require('./sql.service')
+const TYPES = require('tedious').TYPES
 
 const settingDataService = {}
 
-settingDataService.findOne = async function (options) {
-  return Setting.findOne(options).lean().exec()
+settingDataService.findOne = async function () {
+  const sql = 'SELECT * FROM Settings WHERE id=@id'
+  const params = [ {
+    name: 'id',
+    value: 1,
+    type: TYPES.Int
+  }]
+
+  const result = await sqlService.query(sql, params)
+  if (result && result.length === 1) {
+    return result[0]
+  }
+  return {}
 }
 
 /**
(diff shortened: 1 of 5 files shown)


admin/services/data-access/setting.data.service.js
```diff
@@ -1,11 +1,24 @@
...
 
 const Setting = require('../../models/setting')
+const sqlService = require('./sql.service')
+const TYPES = require('tedious').TYPES
 
 const settingDataService = {}
 
-settingDataService.findOne = async function (options) {
-  return Setting.findOne(options).lean().exec()
+settingDataService.findOne = async function () {
```
jon-shipley: Can we call new the sql methods s`sqlFindOne()`?  Then we can build them up side by side and then delete the existing mongo methods.

We can rename from sqlFindOne to findOne at the end.

admin/services/data-access/setting.data.service.js
```diff
@@ -1,11 +1,24 @@
...
+  const sql = 'SELECT * FROM Settings WHERE id=@id'
+  const params = [ {
+    name: 'id',
+    value: 1,
+    type: TYPES.Int
+  }]
+
+  const result = await sqlService.query(sql, params)
+  if (result && result.length === 1) {
```
jon-shipley: The expectation is that findOne returns the first row, no matter how many results there are, or what it's id is.

Something more like `SELECT TOP 1 from $TABLE`

GuyHarwood: @jon-shipley true.  But the settings table is an exception due to the fact that there should only ever be one row.  This is why i omitted the auto increment on the primary key.
Happy to change it, as it doesnt necessarily have to row with `id=1`
Thoughts?

admin/services/data-access/setting.data.service.js
```diff
@@ -1,11 +1,18 @@
...
 
 const settingDataService = {}
 
-settingDataService.findOne = async function (options) {
-  return Setting.findOne(options).lean().exec()
+settingDataService.findOne = async function () {
+  const sql = 'SELECT TOP 1 * FROM Settings'
+
+  const result = await sqlService.query(sql)
+  if (result && result.length === 1) {
```
jon-shipley: Why test for exactly one?  It could be `result.length > 0` I would understand it if we threw an error if there was more than one.

GuyHarwood:
wrong target branch

PR #354, opened 2017-12-20
Feature/sql refactor admin authentication

jon-shipley:
Work in Progress

ToDo: more tests, integration tests need to be enabled

Refactor the admin authentication strategy to use SQL to login


admin/authentication/local-strategy.js
```diff
@@ -64,23 +74,19 @@ async function saveInvalidLogonEvent (logonEvent, message) {
   try {
     logonEvent.errorMsg = message
     logonEvent.isAuthenticated = false
-    await logonEvent.save()
+    await adminLogonEventDataService.sqlCreate(logonEvent)
   } catch (error) {
-    console.log('Failed to save logon event: ' + error.message)
+    winston.log('Failed to save logon event: ' + error.message)
```
GuyHarwood: IIRC you are missing the log level in this call.  Either use `winston.log('info|warn|error|debug', 'message')`
or one of the convenience methods such as `winston.info`

admin/authentication/local-strategy.js
```diff
@@ -64,23 +74,19 @@ async function saveInvalidLogonEvent (logonEvent, message) {
...
-    logonEvent.ncaUserType = session.UserType
     logonEvent.ncaUserName = session.UserName
-    logonEvent.ncaSessionToken = session.SessionToken
-    logonEvent.school = session.School
-    logonEvent.role = session.role
-    logonEvent.save()
+    await adminLogonEventDataService.sqlCreate(logonEvent)
   } catch (error) {
-    console.log('Failed to save logon event: ' + error.message)
+    winston.log('Failed to save logon event: ' + error.message)
```
GuyHarwood: same as above

admin/authentication/local-strategy.js
```diff
@@ -1,61 +1,71 @@
...
-    remoteIp: (req.headers['x-forwarded-for'] || req.connection.remoteAddress),
+    body: JSON.stringify(R.omit(['password'], R.prop('body', req))),
+    remoteIp: req.headers['x-forwarded-for'] || req.connection.remoteAddress,
     userAgent: req.headers['user-agent'],
     loginMethod: 'local'
-  })
+  }
+
+  try {
+    const user = await userDataService.sqlFindOneByIdentifier(email)
```
GuyHarwood: these prefixes are a very good idea! 👍

admin/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "check-development-app",
+  "name": "mtc-admin",
```
GuyHarwood: 👍

admin/package.json
```diff
@@ -63,6 +63,7 @@
     "tedious-connection-pool": "^1.0.5",
     "uuid": "^3.1.0",
     "validator": "^9.2.0",
+    "winston": "^2.4.0",
```
GuyHarwood: this will likely clash with my addition when it hits master.  shouldn't be an issue

admin/services/data-access/sql.service.js
```diff
@@ -11,16 +55,19 @@ function parseResults (results) {
     // const metadata = []
     row.forEach(col => {
       if (col.metadata.colName !== 'version') {
-        json[col.metadata.colName] = col.value
+        if (col.metadata.type.type === 'DATETIMEOFFSETN' && col.value) {
```
GuyHarwood: do we need to capture `'DATETIMEOFFSET'` too?

admin/services/data-access/sql.service.js
```diff
@@ -40,13 +87,15 @@ sqlService.query = (sql, params) => {
     }
     let results = []
     // http://tediousjs.github.io/tedious/api-request.html
+    winston.debug(`sql.service: SQL: ${sql}`)
```
GuyHarwood: we can add a new category in winston.  Would it be better to do this now? 🤔

jon-shipley: I think `debug` is fine, but I don't object to a new category.

PR #359, opened 2017-12-20
Include missing chain from mock call

GeorgeChatzigiannis:


admin/spec/service/data-access/pupil-restart.data.service.spec.js
```diff
@@ -85,7 +85,7 @@ describe('pupil-restart.data.service', () => {
   describe('#update', () => {
     let mock
     beforeEach(() => {
-      mock = sandbox.mock(PupilRestart).expects('updateOne').returns(pupilRestartMock)
+      mock = sandbox.mock(PupilRestart).expects('updateOne').chain('exec').returns(pupilRestartMock)
```
jon-shipley: should be `resolve` rather than return.

GeorgeChatzigiannis: @jon-shipley yes, updated.

PR #364, opened 2017-12-21
Feature/18300 settings log data refactor

GuyHarwood:
userId added to express session.  This represents the `id` field in the `user` table
moved data service usage out of controller and into dedicated service.
this abstracts the addition of settingsLog entries into one place.
data services now use SQL Server

### DIFF
--- a/admin/authentication/local-strategy.js
+++ b/admin/authentication/local-strategy.js
@@ -41,19 +41,20 @@ module.exports = async function (req, email, password, done) {
       return done(null, false)
     }
 
-    const mock = {
+    const sessionData = {
       EmailAddress: email,
       UserName: email,
       UserType: 'SchoolNom',
       School: school.dfeNumber,
       role: role.title,
-      logonAt: Date.now()
+      logonAt: Date.now(),
+      userId: user.id
     }
 
     // Success - valid login
     logonEvent.user_id = user.id
-    await saveValidLogonEvent(logonEvent, mock)
-    return done(null, mock)
+    await saveValidLogonEvent(logonEvent, sessionData)
+    return done(null, sessionData)
   } catch (error) {
     winston.warn(error)
     await saveInvalidLogonEvent(logonEvent, 'Server error: ' + error.message)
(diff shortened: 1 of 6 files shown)


admin/authentication/local-strategy.js
```diff
@@ -41,19 +41,21 @@ module.exports = async function (req, email, password, done) {
...
-      logonAt: Date.now()
+      logonAt: Date.now(),
+      userId: user.id
     }
 
     // Success - valid login
     logonEvent.user_id = user.id
-    await saveValidLogonEvent(logonEvent, mock)
-    return done(null, mock)
+    req.session.userId = user.id
```
jon-shipley: Seems odd that this is required here.  Am I missing something?

GuyHarwood: this is basically adding the userId to the session object, so it can be used to pass to services that need to record it (such as the settingService).

jon-shipley: It's already added in to the sessionData above - I don't think this line `req.session.userId = user.id` is needed.

GuyHarwood: my discovery in manual testing suggests otherwise.  Maybe we have an issue here?
i'll do some further testing with the other values

GuyHarwood: found it! its under `req.session.passport.user`

admin/controllers/service-manager.js
```diff
@@ -104,7 +104,7 @@ const setUpdateTiming = async (req, res, next) => {
   }
 
   try {
-    await settingService.update(settings.loadingTimeLimit, settings.questionTimeLimit, req.session.userId)
+    await settingService.update(settings.loadingTimeLimit, settings.questionTimeLimit, req.session.passport.user.id)
```
jon-shipley: We would usually write this as `req.user.id` rather than access the session / passport data directly.

admin/controllers/service-manager.js
```diff
@@ -44,7 +44,7 @@ const getUpdateTiming = async (req, res, next) => {
   res.locals.pageTitle = 'Check settings'
   let settings
   const successfulPost = req.params.status || false
-
+  console.log('session:', req.session)
```
jon-shipley: debug?

admin/controllers/service-manager.js
GuyHarwood: done 👍

PR #365, opened 2017-12-21
Feature/18190 add edit groups

pris54:
- Add and edit groups
- Assigning and unassigning pupils to groups


admin/assets/javascripts/global-scripts.js
```diff
@@ -18,7 +18,6 @@ $(function () {
         if (validation) {
           validationStatus = validation()
         }
-        console.log('VALIDATE', validationStatus)
```
GuyHarwood: just a reminder that all server side console.log statements should now be `winston.info|debug|warn|error` :)

admin/models/group.js
```diff
@@ -9,7 +9,7 @@ const Group = new Schema({
     type: String,
     required: true
   },
-  pupils: [],
+  pupils: {},
```
GuyHarwood: will this work for a set?

admin/services/data-access/group.data.service.js
```diff
@@ -11,7 +11,57 @@ const groupDataService = {}
 groupDataService.fetchGroups = async function (query) {
   const q = query || {}
   q.isDeleted = false
-  return Group.find(q).sort({ name: 'asc' }).lean().exec()
+  return Group
+    .find(q)
+    .collation({ locale: 'en', strength: 2 })
```
GuyHarwood: this will likely fail in cosmos :(

admin/services/group.service.js
```diff
@@ -1,11 +1,27 @@
...
 }
 
+groupService.getPupils = async function (schoolId) {
+  if (!schoolId) { return false }
+  return pupilDataService.getSortedPupils(schoolId)
+}
+
+groupService.getGroupById = async function (groupId) {
+  if (!groupId) { return false }
+  return groupDataService.fetchGroup({'_id': groupId})
```
GuyHarwood: pupilDataService uses `getXXX` 
groupDataService uses `fetchXXX`
is that intentional?

admin/spec/service/data-access/group.data.service.spec.js
```diff
@@ -19,11 +20,11 @@ describe('GroupDataService', () => {
 
   afterEach(() => sandbox.restore())
 
-  describe('#groupDataService', () => {
+  describe('#fetchGroups', () => {
     let mock
 
     beforeEach(() => {
-      mock = sandbox.mock(Group).expects('find').chain('sort').chain('lean').chain('exec').resolves(groupsMock)
+      mock = sandbox.mock(Group).expects('find').chain('collation').chain('sort').chain('lean').chain('exec').resolves(groupsMock)
```
GuyHarwood: just a reminder to modify this if you remove the collation call from the data service 🎗

admin/assets/javascripts/global-scripts.js
pris54: @GuyHarwood Thanks for the reminder! That was WIP done before winston was in place.

GuyHarwood: @pris54 this particular one is client side, which will still use console.  just server side ones that can rely on winston :)

admin/services/data-access/group.data.service.js
pris54: Arghh...! It's just to make it case insensitive. I'll look into other ways to sort this out.

admin/services/group.service.js
pris54: Yes, it's to not make it confusing (I hope?)

GuyHarwood: i would go with the standard get, otherwise its inconsistent with other data services

admin/services/data-access/group.data.service.js
GuyHarwood: @pris54 speak to george, he had this issue too

jon-shipley:
Nothing more to add other than the comments already made.

admin/services/group.service.js
```diff
@@ -6,22 +6,44 @@ const pupilDataService = require('../services/data-access/pupil.data.service')
...
 groupService.getGroups = async function (query) {
-  return groupDataService.fetchGroups(query)
+  return groupDataService.getGroups(query)
 }
 
-groupService.getPupils = async function (schoolId) {
+groupService.getPupils = async function (schoolId, groupIdToExclude) {
   if (!schoolId) { return false }
-  return pupilDataService.getSortedPupils(schoolId)
+  let query = {}
```
GuyHarwood: this should be done in the data service when we move it to SQL.
Can you just add a `// HACK` comment to this effect.
Once we have got near the end of the data layer refactor i will scan the sources for all `WARN`, `HACK` and `TODO`

admin/services/data-access/group.data.service.js
GuyHarwood: @pris54 this will definitely fail in Cosmos - just remove the collation and add a HACK or TODO comment and we will sort it with SQL Server

admin/services/group.service.js
pris54: @GuyHarwood I've commented out the 'collation' line and added a TODO comment. I've also explained in the PBI what this means in terms of functionality.

PR #366, opened 2017-12-21
Feature/18199 refactor school data service to user sql server

jon-shipley:
Major: new generic sqlService.update() method
Minor: 2 calls to schoolDataService replaced by new sql calls
More calls yet to do - but submitting this PR as the update method may be useful

### DIFF
--- a/admin/controllers/pupil-pin.js
+++ b/admin/controllers/pupil-pin.js
@@ -36,11 +36,11 @@ const getGeneratePinsList = async (req, res, next) => {
   const { htmlSortDirection, arrowSortDirection } = sortingAttributesService.getAttributes(sortingOptions, sortField, sortDirection)
   // TODO: data service call should be moved to a service
   try {
-    school = await schoolDataService.findOne({ _id: req.user.School })
+    school = await schoolDataService.sqlFindOneByDfeNumber(req.user.School)
     if (!school) {
       return next(Error(`School [${req.user.school}] not found`))
     }
-    pupils = await pinGenerationService.getPupils(school._id, sortField, sortDirection)
+    pupils = await pinGenerationService.getPupils(school.dfeNumber, sortField, sortDirection)
   } catch (error) {
     return next(error)
   }
--- a/admin/controllers/pupil-pin.js
+++ b/admin/controllers/pupil-pin.js
@@ -1,3 +1,4 @@
+const R = require('ramda')
 const config = require('../config')
 const schoolDataService = require('../services/data-access/school.data.service')
 const pupilDataService = require('../services/data-access/pupil.data.service')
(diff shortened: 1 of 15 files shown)


admin/controllers/pupil-pin.js
```diff
@@ -36,11 +37,11 @@ const getGeneratePinsList = async (req, res, next) => {
   const { htmlSortDirection, arrowSortDirection } = sortingAttributesService.getAttributes(sortingOptions, sortField, sortDirection)
   // TODO: data service call should be moved to a service
   try {
-    school = await schoolDataService.findOne({ _id: req.user.School })
+    school = await schoolDataService.sqlFindOneByDfeNumber(req.user.School)
     if (!school) {
       return next(Error(`School [${req.user.school}] not found`))
     }
-    pupils = await pinGenerationService.getPupils(school._id, sortField, sortDirection)
+    pupils = await pinGenerationService.getPupils(school.dfeNumber, sortField, sortDirection)
```
GeorgeChatzigiannis: Just so that I understand we are not quering schools based on Id anymore?

admin/controllers/pupil-pin.js
```diff
@@ -62,12 +63,14 @@ const postGeneratePins = async (req, res, next) => {
   try {
     submittedPupils = await pinGenerationService.generatePupilPins(pupilsList)
     await pupilDataService.updateMultiple(submittedPupils)
-    school = await schoolDataService.findOne({_id: req.user.School})
+    school = await schoolDataService.sqlFindOneByDfeNumber(req.user.School)
```
GeorgeChatzigiannis: I was under the impression we are keeping the naming conventions for the methods for consistency. Aren't we developing this in parallel with the mongo layer?

admin/services/data-access/sql.service.js
```diff
@@ -207,6 +283,26 @@ sqlService.generateInsertStatement = async (table, data) => {
   return { sql, params }
 }
 
+/**
+ * Utility function for internal sqlServivce use.  Generate the SQL UPDATE statement and list of parameters
```
GeorgeChatzigiannis: typo 'sqlServivce'

admin/controllers/pupil-pin.js
jon-shipley: Yes, we can, but the database has changed a little for schools, as id is now a new synthetic id, rather than the dfeNumber from before.  So `dfeNumber` is what was the Cosmos `id` field.

Initially we were forced to make the mdfenumber the PK as Cosmos does not support secondary unique indexes, so it was the only way to guarantee the dfeNumber was unique.

For the database having an `id` col as PK is mandatory (for us) - I have seen some guidelines that @GuyHarwood is writing that enforces this.

admin/controllers/pupil-pin.js
jon-shipley: Almost, however we are not going to write a mongo -> sql translation layer, so generic calls like `findOne` will become as many individual queries as needed.

At the moment all new calls are being prefixed by `sql` so we build them up alongside the Mongo calls.

GuyHarwood: im finding the distinction very useful when knee deep in coding.  You know what you've changed and what you havent

admin/services/data-access/sql.service.js
```diff
@@ -207,6 +283,26 @@ sqlService.generateInsertStatement = async (table, data) => {
...
+  const sql = R.join(' ', [
+    'UPDATE',
+    table,
+    'SET ',
+    generateSetStatements(R.omit(['id'], data)),
+    'WHERE id=@id'
+  ])
+  winston.debug('sql.service: SQL ', sql)
+  return { sql, params }
+}
```
GuyHarwood: 👍

PR #367, opened 2017-12-21
Ruby Framework for SQL implementation

sinhaamit1980:


admin/test/config/database.yml
```diff
@@ -0,0 +1,18 @@
+localhost:
+  username: 'sa'
+  password: 'Mtc-D3v.5ql_S3rv3r'
+  host: 'localhost'
+  port: '1433'
+  database: 'mtc'
+dev:
```
GuyHarwood: i would just have one set that default to the local dev instance values, but can be overridden by env vars.
see `./admin/config.js` for inspiration

admin/test/features/support/env.rb
```diff
@@ -56,4 +57,13 @@
   CLIENT = Mongo::Client.new('mongodb://localhost/mtc')
 end
 
+DATABASE_CONFIG = YAML::load(File.read("#{File.dirname(__FILE__)}/../../config/database.yml"))
```
GuyHarwood: i think this would be a good place to do your env var overrides

admin/test/features/support/sql_db_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def self.pupil_details_using_names(firstname, lastname)
+    sql = "SELECT * FROM [mtc_admin].[pupil] WHERE foreName='#{firstname}' AND lastName='#{lastname}'"
+    result = SQL_CLIENT.execute(sql)
+    result.first
+  end
+
+  def self.find_teacher(name)
+    @array_of_users = []
+    sql = "SELECT * FROM [mtc_admin].[user] WHERE identifier='#{name}'"
```
GuyHarwood: identifier is unique, so you will only ever get one result

admin/test/features/support/sql_db_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def self.pupil_details(upn)
+    sql = "SELECT * FROM [mtc_admin].[pupil] WHERE upn='#{upn}'"
+    result = SQL_CLIENT.execute(sql)
+    result.first
+  end
+
+  def self.pupil_details_using_names(firstname, lastname)
+    sql = "SELECT * FROM [mtc_admin].[pupil] WHERE foreName='#{firstname}' AND lastName='#{lastname}'"
+    result = SQL_CLIENT.execute(sql)
```
GuyHarwood: its highly likely that users will have same forename and surname.  maybe return set?

admin/test/features/support/sql_db_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+    @array_of_users = []
+    sql = "SELECT * FROM [mtc_admin].[user] WHERE identifier='#{name}'"
+    result = SQL_CLIENT.execute(sql)
+    @array_of_users = result.each{|row| row.map}
+  end
+
+  def self.find_school(school_id)
+    @array_of_schools = []
+    sql = "SELECT * FROM [mtc_admin].[school] WHERE id='#{school_id}'"
+    result = SQL_CLIENT.execute(sql)
```
GuyHarwood: you will only ever get one result at most

admin/test/features/support/sql_db_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+    @array_of_schools = result.each{|row| row.map}
+  end
+
+  def self.find_school_by_dfeNumber(school_dfeNumber)
+    @array_of_schools = []
+    sql = "SELECT * FROM [mtc_admin].[school] WHERE dfeNumber='#{school_dfeNumber}'"
+    result = SQL_CLIENT.execute(sql)
```
GuyHarwood: same as above

GuyHarwood:
looking good, a few suggestions 👍

admin/test/features/support/sql_db_helper.rb
jon-shipley: Sometimes this is used to dereference the container array.

GuyHarwood: 👍

admin/test/config/database.yml
sinhaamit1980: Done

admin/test/features/support/sql_db_helper.rb
sinhaamit1980: We will refactor it when we will start replacing the database method in our page objets.
I think we need this method for the newly created pupil which have unique name

admin/test/config/database.yml
```diff
@@ -0,0 +1,18 @@
...
+  password: 'Mtc-D3v.5ql_S3rv3r'
+  host: ''
+  port: '1433'
+  database: 'mtc'
+test:
+  username: 'sa'
+  password: 'Mtc-D3v.5ql_S3rv3r'
+  host: ''
+  port: '1433'
+  database: 'mtc'
```
activemq: @sinhaamit1980 Is there a need for this file now? As your setting vars in the env.rb now?

admin/test/features/support/env.rb
```diff
@@ -56,4 +57,26 @@
   CLIENT = Mongo::Client.new('mongodb://localhost/mtc')
 end
 
+Database = ENV['SQL_DATABASE'] || 'mtc'
+Server = ENV['SQL_SERVER'] || 'localhost'
+Port =  ENV['SQL_PORT'] || 1433
+Admin_User = ENV['SQL_ADMIN_USER'] || 'sa'
+Admin_Password = ENV['SQL_ADMIN_USER_PASSWORD'] || 'Mtc-D3v.5ql_S3rv3r'
```
activemq: please use ruby variable style ie. admin_password etc Lowercase and spaces replaced with underscores.

admin/test/features/support/env.rb
```diff
@@ -57,13 +57,26 @@
...
 end
 
-DATABASE_CONFIG = YAML::load(File.read("#{File.dirname(__FILE__)}/../../config/database.yml"))
-ENV['SQL_ENV'] ||='localhost'
-SQL_CLIENT = TinyTds::Client.new(username: DATABASE_CONFIG[ENV['SQL_ENV']]['username'],
-                                 password: DATABASE_CONFIG[ENV['SQL_ENV']]['password'],
-                                 host: DATABASE_CONFIG[ENV['SQL_ENV']]['host'],
-                                 port: DATABASE_CONFIG[ENV['SQL_ENV']]['port'],
-                                 database: DATABASE_CONFIG[ENV['SQL_ENV']]['database'])
+database = ENV['SQL_DATABASE'] || 'mtc'
```
GuyHarwood: 👍

PR #368, opened 2017-12-21
Feature/18316 make user.school_id nullable with conditions

GuyHarwood:
`user.school_id` should be optional for service_manager and test-developer roles.

Enforce condition via a trigger


admin/data/sql/migrations/017.do.user-school-trigger.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TRIGGER user_school_id_null_check
```
jon-shipley: Question: should we include the trigger conditions in the name?  e.g. user_school_id_null_check_insert_update ?

PR #370, opened 2018-01-06
Feature/18191 remove group

pris54:
Soft-deleting groups.
Depends on branch 'feature/18190-add-edit-groups' PR #365


admin/services/data-access/group.data.service.js
```diff
@@ -8,10 +8,64 @@ const groupDataService = {}
...
  * @returns {Promise<Promise|*>}
  */
-groupDataService.fetchGroups = async function (query) {
+groupDataService.getGroups = async function (query) {
   const q = query || {}
   q.isDeleted = false
-  return Group.find(q).sort({ name: 'asc' }).lean().exec()
+  return Group
+    .find(q)
+    .collation({ locale: 'en', strength: 2 })
```
GeorgeChatzigiannis: this will not work with cosmosDB and will throw an error in the frontend.
it was the reason why I had to block pupils case insensitive sorting.

admin/lib/validator/group-validator.js
```diff
@@ -0,0 +1,40 @@
...
+module.exports.validate = async (groupData, oldName) => {
+  let validationError = new ValidationError()
+
+  // Group name
+  if (!groupData.name || isEmpty(groupData.name.trim())) {
+    validationError.addError('name', groupErrorMessages.nameIsRequired)
+  }
+
+  if (!isEmpty(groupData.name.trim())) {
+    if (!XRegExp('^[\\p{Latin}-\' 0-9]+$').test(groupData.name)) {
```
GeorgeChatzigiannis: might be useful to add a comment to explain this regexp

admin/services/data-access/group.data.service.js
GuyHarwood: @GeorgeChatzigiannis @pris54 I wonder if we should just park this subtle difference until we tackle it in the data layer refactor?  seems a waste for Paula to spend time on it when it will soon be refactored anyway...

GeorgeChatzigiannis: I was under the impression that every new PR from this point and forward should have data layer work done in SQL?

pris54: @GuyHarwood @GeorgeChatzigiannis Thanks for the comments on this issue. My understanding after seeing Guy's first comment was that this issue will be tackled as part of the SQL refactoring. I've added a comment to the PBI explaining this.

GuyHarwood: I don't see how it would have been feasible for Paula to do this in SQL, given the xmas break and the fact that nobody was around to provide guidance on the new services etc.
Now that we are back we should draw a line under CosmosDB.

GeorgeChatzigiannis: Ok that makes sense

admin/services/group.service.js
```diff
@@ -1,11 +1,49 @@
...
+  if (!schoolId) { return false }
+  let query = {}
+  let filteredPupil = []
+  let groupedPupils = []
+  let pupils = await pupilDataService.getSortedPupils(schoolId)
+
+  if (groupIdToExclude) {
+    query = { '_id': { $ne: groupIdToExclude } }
+  }
+  const groups = await groupDataService.getGroups(query)
```
GuyHarwood: most of this can be done in SQL when we do the refactor

pris54: Moving SQL will simplify our code 👍

admin/services/data-access/group.data.service.js
pris54: @GeorgeChatzigiannis @GuyHarwood This is a branch of the add/edit one ( PR #365 ) which means that this line of code has been commented out with a @TODO comment on top (see PR 365)

PR #371, opened 2018-01-09
bug/18328 - Check forms - correct sample CSV file to download

pris54:
Bug fix: linking to the correct CSV file.


admin/assets/CSVs/check-form-sample.csv
```diff
@@ -0,0 +1,25 @@
...
+11,2
+2,4
+5,7
+3,2
+8,11
+12,8
+7,4
+4,7
+3,9
+12,12
```
jon-shipley: repeated 12 x 12

pris54: Fixed!

PR #372, opened 2018-01-09
Feature/18290 check form data service refactor

GuyHarwood:
# Work in Progress


admin/services/check-form.service.js
```diff
@@ -144,6 +144,7 @@ const checkFormService = {
    * @param CheckWindowsByForm
    * @returns {Promise.<void>}
    */
+  // WARN this expects a CheckWindow but is passed a check form in controllers/check-form.js
```
GuyHarwood: @pris54 @GeorgeChatzigiannis @jon-shipley This method is called from `controllers/check-form.js` and expects a `checkWindow` as the first parameter.  It is passed a checkForm mongoose object.
Anyone remember this code or provide some clarity?

admin/services/data-access/check-form.data.service.js
```diff
@@ -34,38 +63,197 @@ const checkFormDataService = {
...
+   */
+  sqlFetchSortedActiveFormsByName: (windowId, sortDescending) => {
+    let sortOrder = 'ASC'
+    if (sortDescending) {
+      sortOrder = 'DESC'
+    }
+    const params = []
+    let sql = ''
+    if (windowId) {
+      sql = `SELECT * FROM [mtc_admin].[checkForm] WHERE isDeleted=0
```
jon-shipley: I'm thinking that we should not hard-code the database name `[mtc_admin]`.  In my latest PR (soon forthcoming) I export this from sqlService.  We can leave as-is for now though.

GuyHarwood: @jon-shipley this is the schema name, which is important

GuyHarwood: however, given that the user only has access to this schema it may be irrelevant.  We should test this.

PR #373, opened 2018-01-10
Pupil pin generate changes

GeorgeChatzigiannis:
Pupil PIN:
-  consists of 4 characters
- randomly generated from numbers (2-9)


admin/services/pin-generation.service.js
```diff
@@ -98,6 +97,15 @@ pinGenerationService.generateSchoolPassword = (school) => {
   return school
 }
 
+/**
+ * Generate Pupil Pin
+ * @returns {String}
+ */
+pinGenerationService.generatePupilPin = () => {
+  const chars = '23456789'
```
jon-shipley: If it is only numeric, why exclude the number '1'?

admin/services/pin-generation.service.js
```diff
@@ -98,6 +97,15 @@ pinGenerationService.generateSchoolPassword = (school) => {
...
+  return randomGenerator.getRandom(4, chars)
```
pris54: Would it not be better to have this argument (the length) in the config file? Same for number 8 in generateSchoolPassword?

GeorgeChatzigiannis: @pris54 This is a business rule that will apply to all cases. Having it in the config would mean that we would want to differentiate between cases which based on the AC won't really happen.

admin/services/pin-generation.service.js
GeorgeChatzigiannis: @jon-shipley based on the AC they want to exclude number 1

jon-shipley: It makes sense to exclude the number 1 when mixed alphanumerics are being used to prevent ambiguous characters being interpreted incorrectly, but I see little benefit here.

PR #374, opened 2018-01-10
feature/17425 - Sticky banner dynamic positioning.

pris54:
With this modification the sticky banner remains always above the footer. 
Tested on Chrome, Firefox and Safari for Mac. IE 9, 10, 11. Chrome, Safari and Firefox for Windows. iPad(s).


admin/assets/javascripts/global-scripts.js
```diff
@@ -154,6 +153,7 @@ $(function () {
      * @param status
      */
     toggle: function (status) {
+      stickyBannerPositioning()
```
GeorgeChatzigiannis: was thinking alternative naming for this convention.
how about adjustStickyBanner?
From what I noticed in this file the general approach is verb first.

activemq:
Tested across different browsers, a few issues with mobiles, however paula has agreed with sinead for these not to be a priority for the trial.

admin/assets/javascripts/global-scripts.js
jon-shipley: I think noun-verb naming is correct

admin/assets/javascripts/global-scripts.js
```diff
@@ -253,6 +253,26 @@ $(function () {
     }
   }
 
+  /**
+   * Sticky banner positioning.
+   */
+  function stickyBannerPositioning () {
+    var windowHeight = $(window).height()
```
jon-shipley: The admin app should be functional without javascript.  What happens here?

pris54: @jon-shipley We are going for JS enabled only, I've confirmed this three times.

admin/assets/javascripts/global-scripts.js
pris54: @jon-shipley @GeorgeChatzigiannis happy if I rename it adjustStickyBanner? Please confirm.

admin/assets/javascripts/global-scripts.js
jon-shipley: I see. That's fine then.

PR #375, opened 2018-01-10
School password generate changes

GeorgeChatzigiannis:
New pattern consists of:
- a random 3 letter lower case word generated from a list
- two random digits(2-9 only allowed) that are placed at a random index within the word being used


admin/services/pin-generation.service.js
```diff
@@ -90,8 +92,11 @@ pinGenerationService.generatePupilPins = async (pupilsList) => {
 pinGenerationService.generateSchoolPassword = (school) => {
   let { schoolPin, pinExpiresAt } = school
   if (!pinValidator.isActivePin(schoolPin, pinExpiresAt)) {
-    const length = 8
-    school.schoolPin = pinGenerationService.generateRandomPin(length)
+    const allowedWords = (config.Data.allowedWords && config.Data.allowedWords.split(',')) || []
+    const firstRandomWord = allowedWords.length > 0 ? allowedWords[Math.floor(Math.random() * allowedWords.length)] : ''
```
jon-shipley: Math.random() does not have enough unpredictability to be used for generating passwords.  Please use the crypto library instead.  https://nodejs.org/api/crypto.html

I suspect that reading and splitting the csv word list at startup It would also be better to parse the allowedWord list at startup and then keep it as as a data structure (e..g a Set) during operation.  

There are also limits to size of data that can be kept in environment variables - let's discuss this in private chat.

admin/spec/service/pin-generation.service.spec.js
```diff
@@ -190,6 +191,8 @@ describe('pin-generation.service', () => {
         const result = pinGenerationService.generateSchoolPassword(school)
         expect(result.pinExpiresAt).toBeDefined()
         expect(result.schoolPin.length).toBe(8)
+        expect(/[a-z]/i.test(result.schoolPin)).toBe(true)
+        expect(/[1-9]/i.test(result.schoolPin)).toBe(true)
```
jon-shipley: This doesn't test that you get word-number-word

GeorgeChatzigiannis: @jon-shipley updated

admin/services/pin-generation.service.js
GeorgeChatzigiannis: @jon-shipley updated with randomBytes method usage from crypto lib

admin/services/pin-generation.service.js
```diff
@@ -90,30 +94,39 @@ pinGenerationService.generatePupilPins = async (pupilsList) => {
 pinGenerationService.generateSchoolPassword = (school) => {
   let { schoolPin, pinExpiresAt } = school
   if (!pinValidator.isActivePin(schoolPin, pinExpiresAt)) {
-    const length = 8
-    school.schoolPin = pinGenerationService.generateRandomPin(length)
+    const allowedWords = (config.Data.allowedWords && config.Data.allowedWords.split(',')) || []
+    const firstRandomWord = allowedWords.length > 0
```
jon-shipley: These ternary expressions are duplicated.  It make more sense, IMHO, to palace a guard condition at the start of the function to ensure that allowedWords is populated.

admin/services/pin-generation.service.js
```diff
@@ -90,30 +94,39 @@ pinGenerationService.generatePupilPins = async (pupilsList) => {
...
- * Generate Pupil Pin
- * @returns {String}
- */
-pinGenerationService.generatePupilPin = () => {
-  const chars = '23456789'
-  return randomGenerator.getRandom(4, chars)
+ * Generating random numbers in specific range using crypto.randomBytes from crypto library
+ * Maximum available range is 281474976710655 or 256^6-1
+ * Maximum number for range must be equal or less than Number.MAX_SAFE_INTEGER (usually 9007199254740991)
+ * @param minimum
```
jon-shipley: The jsdoc needs to specify whether the max and min values in the range are inclusive or exclusive.    E.g. will they be in the output or not?

admin/services/pin-generation.service.js
```diff
@@ -90,30 +94,39 @@ pinGenerationService.generatePupilPins = async (pupilsList) => {
...
- * Generate Random Pin
- * @param length
+ * Generate Pupil Pin
  * @returns {String}
  */
-pinGenerationService.generateRandomPin = (length) => {
-  const chars = '23456789bcdfghjkmnpqrstvwxyz'
-  return randomGenerator.getRandom(length, chars)
+pinGenerationService.generatePupilPin = () => {
+  return randomGenerator.getRandom(4, chars)
```
jon-shipley: I think we should remove this magic number '4' - please define it as pupilPinLength or something similar.

admin/spec/service/pin-generation.service.spec.js
```diff
@@ -190,6 +190,10 @@ describe('pin-generation.service', () => {
         const result = pinGenerationService.generateSchoolPassword(school)
         expect(result.pinExpiresAt).toBeDefined()
         expect(result.schoolPin.length).toBe(8)
+        const schoolPin = result.schoolPin
+        expect(/[a-z]{3}/.test(schoolPin.slice(0, 3))).toBe(true)
```
jon-shipley: A single regex would cover all 3 cases: /[a-z][3}-[2-9]{2}-[a-z]{3}/

Note the numbers 0 and 1 should not be in the output.

admin/spec/service/pin-generation.service.spec.js
```diff
@@ -190,6 +190,7 @@ describe('pin-generation.service', () => {
         const result = pinGenerationService.generateSchoolPassword(school)
         expect(result.pinExpiresAt).toBeDefined()
         expect(result.schoolPin.length).toBe(8)
+        expect(/[a-z]{3}[2-9]{2}[a-z]{3}/.test(result.schoolPin)).toBe(true)
```
jon-shipley: I think this would benefit from anchors: `/^[a-z]{3}[2-9]{2}[a-z]{3}$/ otherwise it would match nonsense like `222abc99def111`

admin/spec/service/pin-generation.service.spec.js
```diff
@@ -220,4 +221,31 @@ describe('pin-generation.service', () => {
...
+          acc[ curr ] = 1
+        } else {
+          acc[ curr ] += 1
+        }
+        return acc
+      }, [])
+      counters = Object.values(counters)
+      counters.forEach((c, i) => {
+        const sample = Math.floor((counters[ i ] = c / length) * 100)
+        expect(sample >= 0 && sample <= 25).toBeTruthy()
```
jon-shipley: I don't understand what error tolerance you have applied to the ideal mathematical probability here.   Here is my rewrite of this test.  In practise, allowing a 15% tolerance seems to work fine.

```
it('should approximately generate numbers with similar probability chance', () => {
      const numbersArr = []
      const length = 1000
      const min = 1
      const max = 5

      for (let index = 0; index < length; index++) {
        numbersArr.push(pinGenerationService.generateCryptoRandomNumber(min, max))
      }

      const count = {}
      for (let num of numbersArr) {
        count[num] = count[num] ? count[num] + 1 : 1
      }

      const expectedResult = numbersArr.length / ((max - min) + 1)
      const errorPercentTolerance = 0.15 // 15%
      const tolerance = expectedResult * errorPercentTolerance

      for (let [number, freq] of Object.entries(count)) {
        expect(number >= min && number <= max).toBeTruthy()
        expect(freq).toBeGreaterThanOrEqual(expectedResult - tolerance)
        expect(freq).toBeLessThanOrEqual(expectedResult + tolerance)
      }
    })
```

.travis.yml
```diff
@@ -19,12 +19,12 @@ env:
...
   - secure: fXLqXS0ziuJxyIf5rCtBJ+3/YFdyVYu8sgSmFecCs4wATN8w6PSgP2t36JOlAPbm96LiDTNGlyIh8eQvu2FcKDp0S0R1thrTfjH/EVvA/Gq46w1MdTv/21R1KoeJhzIuNkLnRLG3EdYN/PSWdmLALJADeG7aAtDpfdV7HUN658ur9Rr5vvWoR9R7rK/iOqk34k4tF7ZRIe2ilKNj4HzsqlolQfNzIiwWMibUNRBzrHEEO+jW3u+ByF7TAEar6evolfC8luZfPjlIhRTHkmSkbXFjxKmr6ZzU/1UqCNye939POzwGQ7CFI2+danmsRBZhfBzXHuhRkcBhYpR2F0+UuxXA8exA/d820dMN0WApXHvau6IvnccSdZtE15TeOEAVqQKyumIIQClWZvL1RusZMAGtDOFEUzg2o4yGzbDQS7cSzROutdsHuMg6GwL6im4QgeZQHW9Cw3GazDWycpqe42+NSJ3g9NbdV8/9vlbARsNUDtcXm6VfWdQi+/qtmkM9SCFb3OWIg3Nmnt1DKtiEJfib/o4eaUwDQJ29toTTmq4zX9fNPi3nL6vDFtUVGQozGBV4fuLxNH8TsbKSa/8/Hjd1C80TO9uOZJHuNuunmZyjXxvS6/9D+CU/mzXUdC6Be6sSpppWQQsIQjgIILYLxyC2yNOl5X1imgGtqXwFnvQ=
+  - ALLOWED_WORDS="ant,bag,ben,bid,bin,box"
```
GeorgeChatzigiannis: @jon-shipley I tried this it had no impact on travis.

admin/config.js
```diff
@@ -18,6 +18,9 @@ module.exports = {
   TSO_AUTH_PUBLIC_KEY: process.env.TSO_AUTH_PUBLIC_KEY,
   TIME_BETWEEN_QUESTIONS: 2,
   LINES_PER_CHECK_FORM: 25,
+  Data: {
+    allowedWords: process.env.ALLOWED_WORDS || 'ant,bag,ben,bid,bin,box'
```
GeorgeChatzigiannis: @jon-shipley I wanted to provide some default values in case the env is not definedas well but I was under the impression we wanted to find a solution that includes travis.
As far as I tried there was no way for travis to expose node env variables.

admin/spec/service/pin-generation.service.spec.js
```diff
@@ -220,4 +221,37 @@ describe('pin-generation.service', () => {
...
+        count[num] = count[num] ? count[num] + 1 : 1
+      }
+
+      const expectedResult = numbersArr.length / ((max - min) + 1)
+      const errorPercentTolerance = 0.15 // 15%
+      const tolerance = expectedResult * errorPercentTolerance
+
+      for (let [number, freq] of Object.entries(count)) {
+        expect(number >= min && number <= max).toBeTruthy()
+        expect(freq).toBeGreaterThanOrEqual(expectedResult - tolerance)
```
GeorgeChatzigiannis: this failed on travis already once when tests rerun with my last commit, perhaps the errorPercentTolerance should be changed

jon-shipley: Upped from 15 - 18%

admin/services/pin-generation.service.js
```diff
@@ -88,32 +94,43 @@ pinGenerationService.generatePupilPins = async (pupilsList) => {
...
- * Generate Pupil Pin
- * @returns {String}
- */
-pinGenerationService.generatePupilPin = () => {
-  const chars = '23456789'
-  return randomGenerator.getRandom(4, chars)
+ * Generating random numbers in specific range using crypto.randomBytes from crypto library
+ * Maximum available range is 281474976710655 or 256^6-1
+ * Maximum number for range must be equal or less than Number.MAX_SAFE_INTEGER (usually 9007199254740991)
+ * @param minimum - inclusive
```
jon-shipley: Thanks for documenting this! 👍

PR #377, opened 2018-01-11
set isTestAccount flag for speech synth user

GuyHarwood:
the speech synthesis enabled test account does not have the `isTestAccount` flag set, so the pin is being wiped every night when the pin purge job runs.


jon-shipley:
Thanks!

PR #379, opened 2018-01-11
Upgrade the version of marked from `0.3.6` which has CVEs against it …

jon-shipley:
…to 0.3.12


checkapi/package.json
```diff
@@ -26,5 +26,9 @@
   "devDependencies": {
     "jsdoc": "^3.5.5",
     "standard": "^10.0.3"
+  },
+  "resolutions": {
+    "marked": "marked 0.3.6 has security vulns. jsdoc depends on marked",
```
jon-shipley: This is a poor mans comment

PR #380, opened 2018-01-12
Feature/18291 check window data service refactor

GuyHarwood:


admin/controllers/check-form.js
```diff
@@ -231,15 +227,15 @@ const displayCheckForm = async (req, res) => {
   }
 
   try {
-    checkWindows = await checkWindowService.getCheckWindowsAssignedToForms()
+    checkWindows = await checkWindowService.getCheckWindowsAssignedToFormsV2([formId])
```
jon-shipley: I agree with the added warning here.  I don't think it is good if the low level data access functions throw errors is something isn't found.  They simply do not have enough context to do that (if that is indeed what is happening)

admin/controllers/check-form.js
```diff
@@ -297,22 +293,23 @@ const assignCheckFormToWindowPage = async (req, res, next) => {
   let checkWindow
 
   try {
-    checkWindow = await checkWindowDataService.fetchCheckWindow(checkWindowId)
+    checkWindow = await checkWindowDataService.sqlFetchCheckWindow(checkWindowId)
```
jon-shipley: It might be better to call it `sqlFindOneById` (assuming checkWindow.id is passed over) and should return an object not an array.  The CheckWindow part is duplicated as this is the Check Window data service.

`find()` - returns an array
`findOne()` - returns an object

admin/controllers/check-form.js
```diff
@@ -379,21 +368,25 @@ const unassignCheckFormsFromWindowPage = async (req, res, next) => {
   let checkWindow
 
   try {
-    checkWindow = await checkWindowDataService.fetchCheckWindow(checkWindowId)
-    checkFormsList = await checkFormService.getAssignedFormsForCheckWindow(checkWindow.forms)
+    checkWindow = await checkWindowDataService.sqlFetchCheckWindow(checkWindowId)
```
jon-shipley: `sqlFindOneById()`?

admin/controllers/service-manager.js
```diff
@@ -139,15 +139,15 @@ const getCheckWindows = async (req, res, next) => {
 
   // Get current check windows
   try {
-    checkWindows = await checkWindowDataService.fetchCurrentCheckWindows(sortField, sortDirection)
-    checkWindowsCurrent = checkWindowService.formatCheckWindowDocuments(checkWindows, true)
+    checkWindows = await checkWindowDataService.sqlFetchCurrentCheckWindows(sortField, sortDirection)
```
jon-shipley: I thought we agreed to use 'find' methods in data services?

If so, this might be called `sqlFindCurrent()` or perhaps `sqlFindOneCurrent` or something similar - I am unsure how there could be more than one check window open at any one time.   It sounds like a bug.

admin/controllers/service-manager.js
```diff
@@ -139,15 +139,15 @@ const getCheckWindows = async (req, res, next) => {
...
+    checkWindowsCurrent = checkWindowService.formatCheckWindowDocuments(checkWindows, true, false)
   } catch (error) {
     return next(error)
   }
 
   // Get past check windows
   try {
-    checkWindows = await checkWindowDataService.fetchPastCheckWindows(sortField, sortDirection)
+    checkWindows = await checkWindowDataService.sqlFetchPastCheckWindows(sortField, sortDirection)
```
jon-shipley: s/Fetch/Find/ ?  Again CheckWindow is not really needed in the function name, although some may prefer it.

admin/controllers/service-manager.js
```diff
@@ -350,7 +350,7 @@ const removeCheckWindow = async (req, res, next) => {
       req.flash('error', 'Deleting an active check window is not allowed.')
     } else {
       try {
-        await checkWindowDataService.setDeletedCheckWindow(req.params.checkWindowId)
+        await checkWindowDataService.sqlSetDeletedCheckWindow(req.params.checkWindowId)
```
jon-shipley: `sqlDeleteCheckWindow()` might be clearer.  The implementation can still be a soft-delete.

admin/data/sql/migrations/024.do.create-vewCheckWindowsWithFormCount.sql
```diff
@@ -0,0 +1,8 @@
+CREATE VIEW [mtc_admin].[vewCheckWindowsWithFormCount] AS
```
jon-shipley: s/vew/view/ ?

admin/services/data-access/check-form.data.service.js
```diff
@@ -129,18 +129,28 @@ const checkFormDataService = {
...
-    WHERE cf.isDeleted=0 AND fw.checkWindow_id=@windowId
-    ORDER BY cw.[name] ${sortOrder}`
-    params.push({
-      name: 'windowId',
-      value: windowId,
-      type: TYPES.Int
-    })
+    let sql
+    if (windowId) {
+      sql = `SELECT cf.*, cw.[name] FROM [mtc_admin].[checkForm] cf
```
jon-shipley: There is no need to specify the schema here '[mtc_admin]'. Suggest we omit the schema.

admin/services/data-access/check-form.data.service.js
```diff
@@ -28,7 +28,7 @@ const checkFormDataService = {
    * @param id
    * @returns {Promise<*>}
    */
-  sqlGetActiveForm: (id) => {
+  sqlGetActiveForm: (id = undefined) => {
```
jon-shipley: It might be better to be called `sqlFindActiveForm`

admin/services/data-access/check-window.data.service.js
```diff
@@ -2,31 +2,76 @@
...
   fetchCheckWindow: async (id) => {
+    winston.warn('check-window.data.service.fetchCheckWindow is deprecated')
     return CheckWindow.findOne({'_id': id, 'isDeleted': false}).exec()
+  },
+    /**
+   * Fetch check window document by id.
+   * @param id
+   * @returns {Promise.<*>}
+   */
+  sqlFetchCheckWindow: async (id) => {
```
jon-shipley: it might be better to be called `sqlFindCheckWindow` for consistency with other data access services.

admin/services/data-access/check-window.data.service.js
```diff
@@ -2,31 +2,76 @@
...
+    const sql = 'SELECT * FROM [mtc_admin].[checkWindow] WHERE isDeleted=0 AND id=@id'
+    const params = [
+      {
+        name: 'id',
+        value: id,
+        type: TYPES.Int
+      }
+    ]
+    return sqlService.query(sql, params)
```
jon-shipley: If this can only return at most 1 item, then the object should be returned rather than the array

admin/services/data-access/check-window.data.service.js
```diff
@@ -49,47 +94,151 @@ const checkWindowDataService = {
...
+  },
+    /**
+   * Fetch check windows by status, sort by, sort direction and date (current or past).
+   * @param isDeleted
+   * @param sortBy valid values are [checkWindowName|adminStartDate|checkStartDate]
+   * @param sortDirection valid values are [asc|desc]
+   * @param isCurrent
+   * @returns {Promise.<void>}
+   */
+  sqlFetchCheckWindows: async (sortBy, sortDirection, isDeleted, isCurrent) => {
```
jon-shipley: It might be better named `sqlFindCheckWindows`

admin/services/data-access/check-window.data.service.js
```diff
@@ -49,47 +94,151 @@ const checkWindowDataService = {
...
     if (!checkWindow) {
       throw new Error('No check-window is currently available')
     }
     return checkWindow
+  },
+    /**
+   * Fetch (one) check window for present date.
+   * @returns {Promise.<*>}
+   */
+  sqlFetchCurrentCheckWindow: async () => {
```
jon-shipley: It might be better named sqlFindCurrent() and return an object

admin/services/data-access/check-window.data.service.js
```diff
@@ -49,47 +94,151 @@ const checkWindowDataService = {
...
   },
   /**
    * Fetch (non-deleted) current check windows by sort by, sort direction
    * @param sortBy
    * @param sortDirection
    * @returns {Promise.<*|Promise.<void>>}
    */
-  fetchCurrentCheckWindows: async (sortBy, sortDirection) => {
-    return checkWindowDataService.fetchCheckWindows(sortBy, sortDirection, false, true)
+  sqlFetchCurrentCheckWindows: async (sortBy, sortDirection) => {
```
jon-shipley: Whilst I am not sure why we need this method, it might be better names `sqlFindCurrentCheckWindows()`

admin/services/data-access/check-window.data.service.js
```diff
@@ -49,47 +94,151 @@ const checkWindowDataService = {
...
+  },
+    /**
+   * Create a new check window
+   * @param data
+   * @return {Promise.<*>}
+   */
+  sqlCreate: async (data) => {
+    return sqlService.create('[checkWindow]', data)
+  },
+  sqlFetchCheckWindowsAssignedToForms: async (formIds) => {
```
jon-shipley: It might be better named `sqlFindCheckWindowsAssignedToForms`

admin/.vscode/launch.json
```diff
@@ -7,8 +7,20 @@
...
       "program": "${workspaceFolder}/server.js"
-    }
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Jasmine Individual Test",
+      "program": "${workspaceRoot}/node_modules/jasmine/bin/jasmine.js",
+      "args": [
+          "spec/service/check-form.service.populateFormFile.spec.js"
```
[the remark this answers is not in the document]
GuyHarwood: i use it across branches, and if you use vscode, its very useful for debugging.
however, i will remove the jasmine entry, as its very specific and didnt work anyway.

admin/controllers/service-manager.js
GuyHarwood: yep.  this is what I pointed out the other day - this data service follows no convention at all.  nor do the controller methods etc.  its been an absolute pig to refactor.  Hence why i just prefixed the existing methods with `sql`

admin/data/sql/migrations/024.do.create-vewCheckWindowsWithFormCount.sql
GuyHarwood: that's pure habit from using `tbl`, `vew` and `spr` for the last 20 years 😆

admin/services/data-access/check-window.data.service.js
GuyHarwood: yes.  this caught me out a few times in testing

admin/services/data-access/check-form.data.service.js
GuyHarwood: 🤔  its fully qualified.  if there were other schemas in the same database, and they also contained a table of the same name it would fail

jon-shipley: Perhaps I don't understand what's going on here.  I assumed that it was connecting to the `mtc_admin` schema during setup, but looking at the code that doesn't seem to be the case.

Does that mean that ms-sql will look up the table name and choose some matching table from the information schema??  Not good!  So there is a performance hit in the lookup and the table may not actually even come from the correct schema...

GuyHarwood: the connection connects to the database.  The database can contain many schemas.  Its standard practice to include the schema prefix in queries.  You *can* use the default `dbo` schema, but this is not good practice.  Regarding the performance hit - i can't find anything to clarify this either way, but given the user only has one schema it will be negligible.  Also, this is standard MSSQL stuff - it's no doubt heavily optimised from the perspective of object metadata etc.

admin/data/sql/migrations/024.do.create-vewCheckWindowsWithFormCount.sql
jon-shipley: Okay, np

admin/services/data-access/check-form.data.service.js
jon-shipley: I agree (now) we should include the schema name.   I assumed the schema was part of the connection, as using an unspecified schema can't be good.

I would argue that the schema name is ostensibly  under the control of the DBA, so we should be able to change it with minimal effort - so we should use the variable exported from sqlService.schemaName instead of the hardcoded string.

I'm not arguing that strongly for this, as it's still fairly easy to do a search and replace, but the first solution is more elegant.

GuyHarwood: agreed.

PR #393, opened 2018-01-18
Feature/18294 group data layer refactor

pris54:
Data layer refactoring for 'group' feature.


admin/data/sql/migrations/026.undo.create-group.sql
```diff
@@ -0,0 +1 @@
+DROP VIEW [mtc_admin].[group]
```
GuyHarwood: should be `DROP TABLE`

admin/data/sql/migrations/027.undo.create-pupilGroup.sql
```diff
@@ -0,0 +1 @@
+DROP VIEW [mtc_admin].[pupilGroup]
```
GuyHarwood: should be `DROP TABLE`

admin/data/sql/migrations/027.do.create-pupilGroup.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE [mtc_admin].[pupilGroup]
+(
+  [id] [int] IDENTITY(1,1) NOT NULL,
+  [group_id] INT NOT NULL,
```
GuyHarwood: you will need a `pupil_id` reference and column in here too

admin/services/data-access/group.data.service.js
```diff
@@ -2,8 +2,13 @@
 
 const Group = require('../../models/group')
 const groupDataService = {}
+const sqlService = require('./sql.service')
+const TYPES = require('tedious').TYPES
+const moment = require('moment')
+const R = require('ramda')
 
 /**
+ * @TODO: To be deleted when the data layer refactoring is done.
```
GuyHarwood: please add `@deprecated` with a message informing what the replacment sql function is called.

admin/services/data-access/group.data.service.js
```diff
@@ -20,6 +25,15 @@ groupDataService.getGroups = async function (query) {
     .exec()
 }
 
+/**
+ * Get groups filtered by query.
+ * @returns {Promise<*>}
+ */
+groupDataService.sqlGetGroups = async () => {
+  const sql = `SELECT g.* FROM ${sqlService.adminSchema}.[group] WHERE isDeleted=0 SORT BY name ASC`
```
GuyHarwood: you are aliasing the table as `g` but not creating the alias after the table name.

probably best to just remove the `g.` as you don't necessarily need it here

admin/services/group.service.js
```diff
@@ -36,14 +47,43 @@ groupService.getPupils = async function (schoolId, groupIdToExclude) {
   return filteredPupil
 }
 
+/**
+ * Get group by id.
+ * @param groupId
```
GuyHarwood: all these methods need to have the `@deprecated` attribute, as above

pris54:
NOTE: PR NOT READY FOR REVIEW YET (18 JAN 11:54)

admin/controllers/group.js
```diff
@@ -15,8 +15,10 @@ const groupPupilsPage = async (req, res, next) => {
   res.locals.pageTitle = 'Group pupils'
   let groups
 
+  console.log('-------USER', req.user)
```
GuyHarwood: use winston 😉

admin/services/data-access/group.data.service.js
```diff
@@ -125,9 +130,45 @@ groupDataService.sqlUpdateGroup = async (id, name) => {
...
+    }
+  ]
+
+  let sql = `SELECT p.[id], p.[foreName], p.[middleNames], p.[lastName], g.[group_id]
+    FROM [mtc_admin].[pupil] p 
+    LEFT JOIN [mtc_admin].[pupilGroup] g 
+      ON p.id = g.pupil_id 
+    WHERE p.school_id=@schoolId 
+    AND g.group_id=@groupId  
+    OR g.group_id IS NULL`
```
GuyHarwood: are you looking for _just_ the pupils in groups? or all pupils, and the group if they have one?

admin/services/group.service.js
```diff
@@ -11,7 +11,10 @@ const groupService = {}
  * @returns {Promise<Promise|*>}
  */
 groupService.getGroups = async function (query) {
-  return groupDataService.getGroups(query)
+  // @TODO: TO BE DELETED
```
GuyHarwood: please use `@deprecated`

admin/services/group.service.js
```diff
@@ -47,14 +50,22 @@ groupService.getPupils = async function (schoolId, groupIdToExclude) {
...
+
 /**
  * Get group by id.
  * @param groupId
  * @returns {Promise<*>}
  */
 groupService.getGroupById = async function (groupId) {
   if (!groupId) { return false }
-  return groupDataService.getGroup({'_id': groupId})
+  // @TODO: TO BE DELETED
```
GuyHarwood: use `@deprecated` tag.

also - create a new function with sql prefix rather than editing existing method

admin/services/data-access/group.data.service.js
pris54: Here I'm looking for all the pupils that are in the groupId passed as argument as well as the ones that are available to be in a group. This will produce a list of available pupils (we can't add a pupil to more than one group) as well as the ones already part of the current group (so that the checkbox is checked).
The code has changed a bit though, it's coming in the next push.

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +139,158 @@ groupDataService.update = async function (id, data) {
...
+            name: 'pupilId',
+            value: p,
+            type: TYPES.Int
+          }
+        ]
+
+        // @TODO: This can be improved to avoid the loop with a syntax as follows:
+        // INSERT INTO #SQLAuthority (ID, Value)
+        // VALUES (1, 'First'), (2, 'Second'), (3, 'Third');
+        sqlInsert = `INSERT INTO ${sqlService.adminSchema}.[pupilGroup] (group_id, pupil_id) VALUES (@groupId, @pupilId)`
```
jon-shipley: This loop needs to be refactored before merge please.  Not a TODO item.

It can be written in a single SQL statement fairly trivially:

`INSERT INTO ${sqlService.adminSchema}.[pupilGroup] 
(group_id, pupil_id) 
VALUES 
(42, 1),
(42, 2),
(42, 3),
...
`

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +139,158 @@ groupDataService.update = async function (id, data) {
...
+}
+
+/**
+ * Get pupils filtered by schoolId and/or groupId.
+ * and grouped by group_id.
+ * @param schoolId
+ * @param groupId
+ * @returns {Promise<*>}
+ */
+groupDataService.sqlGetPupils = async (schoolId, groupId) => {
```
jon-shipley: Please rename to `sqlFindPupils`

admin/services/group.service.js
```diff
@@ -1,49 +1,116 @@
...
-  return groupDataService.getGroup({'name': groupName})
+/**
+ * Create group.
+ * @param group
+ * @param groupPupils
+ * @returns {Promise<boolean>}
+ */
+groupService.create = async (group, groupPupils) => {
+  if (!group) { return false }
+  return new Promise(async (resolve, reject) => {
```
jon-shipley: Promise not needed in an async function.  Please remove and use throw to reject and return to resolve.

admin/spec/controllers/group.spec.js
```diff
@@ -14,7 +14,7 @@ const groupsMock = require('../mocks/groups')
 const groupDeletedMock = require('../mocks/group-deleted')
 const pupilsMock = require('../mocks/pupils-with-reason')
 
-describe('group.js controller', () => {
+xdescribe('group.js controller', () => {
```
jon-shipley: I don't see any reason why the updated tests should not be supplied with the code changes.

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +139,158 @@ groupDataService.update = async function (id, data) {
...
+
+/**
+ * Update pupils assigned to group.
+ * @param groupId
+ * @param pupilIds
+ * @returns {Promise<void>}
+ */
+groupDataService.sqlAssignPupilsToGroup = async (groupId, pupilIds) => {
+  // First, delete all pupils for the selected groupId
+  const sqlDelete = `DELETE FROM ${sqlService.adminSchema}.[pupilGroup]
```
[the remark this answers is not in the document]
GuyHarwood: Correct.  https://docs.microsoft.com/en-us/sql/t-sql/language-elements/begin-transaction-transact-sql

SQL Server now supports `TRY CATCH` so you can potentially come up with a nice code block that catches the error, rolls back the transaction, then throws an error to ensure tedious knows something went wrong

admin/services/data-access/group.data.service.js
GuyHarwood: @GeorgeChatzigiannis can you please apply this same concept to your loop block that contains the multiple insert statements

admin/data/sql/migrations/026.do.create-group.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE [mtc_admin].[group]
+(
+  [id] [int] IDENTITY(1,1) NOT NULL,
+  [name] [nvarchar](max) NOT NULL,
```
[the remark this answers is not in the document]
pris54: @jon-shipley thanks for the UNIQUE INDEX suggestion. There is a validation in place to avoid group name duplicates, however strictly we can have duplicates because we do a soft-delete, so the validation checks on active group names.
Yes, a group is always linked to a particular school.

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +25,76 @@ groupDataService.getGroups = async function (query) {
     .exec()
 }
 
+/**
+ * Get active groups (non-soft-deleted).
+ * @returns {Promise<*>}
+ */
+groupDataService.sqlGetGroups = async () => {
```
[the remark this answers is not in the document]
pris54: OK, I'll change it.

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +25,76 @@ groupDataService.getGroups = async function (query) {
...
 groupDataService.getGroup = async function (query) {
   return Group.findOne(query).lean().exec()
 }
 
+/**
+ * Get group by id.
+ * @param groupId
+ * @returns {Promise<void>}
+ */
+groupDataService.sqlGetGroup = async (groupId) => {
```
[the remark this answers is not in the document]
pris54: 👍

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +139,158 @@ groupDataService.update = async function (id, data) {
...
+      value: id,
+      type: TYPES.Int
+    },
+    {
+      name: 'name',
+      value: name,
+      type: TYPES.NVarChar
+    },
+    {
+      name: 'updatedAt',
```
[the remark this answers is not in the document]
pris54: OK

admin/services/data-access/group.data.service.js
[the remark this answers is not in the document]
pris54: Thanks, I'll look into this.

admin/controllers/group.js
```diff
@@ -110,18 +121,19 @@ const addGroup = async (req, res, next) => {
...
       validation: validationError.errors,
-      pupilsList: pupilsList
+      pupilsList,
+      selectedPupils
     })
   }
 
   try {
-    group = await groupDataService.create(group)
+    group = await groupService.create({'name': group.name}, group.pupils)
```
jon-shipley: Unless there are other parameters that can be passed in param 1 here I would just supply name as param 1 as a string.

admin/controllers/group.js
```diff
@@ -110,18 +121,19 @@ const addGroup = async (req, res, next) => {
...
   } catch (error) {
     return next(error)
   }
 
   req.flash('info', 'New group created')
-  req.flash('groupId', group._id)
+  req.flash('groupId', group.id)
```
jon-shipley: This looks like an unconventional use of req.flash... what are you trying to achieve here?

admin/data/sql/migrations/026.do.create-group.sql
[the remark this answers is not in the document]
jon-shipley: Ok, thanks @pris54

admin/services/data-access/group.data.service.js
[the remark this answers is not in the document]
jon-shipley: Looks like the rename hasn't happened yet

admin/services/data-access/group.data.service.js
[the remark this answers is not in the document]
jon-shipley: Looks like the rename hasn't happened yet.  Actually this is finding one by id, so you can just call `sqlService.findOneById(tableName, id)`

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +25,76 @@ groupDataService.getGroups = async function (query) {
...
+
+  const params = [
+    {
+      name: 'groupName',
+      value: groupName,
+      type: TYPES.NVarChar
+    }
+  ]
+
+  const result = await sqlService.query(sql, params)
```
jon-shipley: As it is possible to have multiple groups with the same name in the DB, this could return more than one result.  So - are you sure a db-constraint is not required?

admin/data/sql/migrations/026.do.create-group.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE [mtc_admin].[group]
```
jon-shipley: I strongly suggest we do NOT use reserved words for our table (and column) names.  Please rename, e.g. `groupName`

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +139,158 @@ groupDataService.update = async function (id, data) {
...
+    },
+    {
+      name: 'updatedAt',
+      value: moment.utc(),
+      type: TYPES.DateTimeOffset
+    }
+  ]
+  return sqlService.modify(
+    `UPDATE ${sqlService.adminSchema}.[group] 
+    SET name=@name, updatedAt=@updatedAt
```
jon-shipley: updatedAt should be removed - the app should not be setting it

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +139,158 @@ groupDataService.update = async function (id, data) {
...
+ */
+groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
+  const params = [
+    {
+      name: 'groupId',
+      value: groupId,
+      type: TYPES.Int
+    }
+  ]
+  await sqlService.modify(`DELETE ${sqlService.adminSchema}.[pupilGroup] WHERE group_id=@groupId`, params)
```
jon-shipley: The jsdoc says this is a soft-delete - but the code is doing a hard delete.  Please update one or the other.

admin/services/group.service.js
```diff
@@ -1,49 +1,116 @@
...
+  return groupDataService.sqlGetGroups()
 }
 
+/**
+ * Get pupils filtered by schoolId and groupId.
+ * @param schoolId
+ * @param groupIdToExclude
+ * @returns {Promise<*>}
+ */
 groupService.getPupils = async function (schoolId, groupIdToExclude) {
```
jon-shipley: Why do want to exclude a single group?  Why not make it an array of groups that could be excluded?  Could you supply some more context please?

admin/services/group.service.js
```diff
@@ -1,49 +1,116 @@
...
+
+/**
+ * Update group (group and pupils assigned to groups).
+ * @param id
+ * @param group
+ * @returns {Promise<boolean>}
+ */
+groupService.updateGroup = async (id, group) => {
+  if (!id || !group || !group.name) { return false }
+  return new Promise(async (resolve, reject) => {
```
jon-shipley: You don't need to use a Promise here as the function is async.

admin/services/group.service.js
GuyHarwood: From what i have learnt today, this is to get the pupils that are not in any group so they can be added to one.  However there is too much going on in one function.  needs better name and separation

admin/controllers/group.js
pris54: It's used to know which table element needs to be highlighted without having to pass a parameter in the URL.

admin/controllers/group.js
```diff
@@ -110,18 +120,19 @@ const addGroup = async (req, res, next) => {
...
   try {
-    group = await groupDataService.create(group)
+    groupId = await groupService.create(group.name, group.pupils, req.user.schoolId)
   } catch (error) {
     return next(error)
   }
 
   req.flash('info', 'New group created')
-  req.flash('groupId', group._id)
+  req.flash('groupId', groupId)
```
jon-shipley: I still don't understand this.  `req.flash()` is designed for sending messages.

admin/data/sql/migrations/037.do.create-group.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE [mtc_admin].[group]
```
jon-shipley: `group` is a reserved word and should be avoided.

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +24,99 @@ groupDataService.getGroups = async function (query) {
...
+ * @returns {Promise<*>}
+ */
+groupDataService.sqlFindGroups = async (schoolId) => {
+  const sql = `
+  SELECT g.id, g.name, COUNT(pg.pupil_id) as pupilCount 
+  FROM ${sqlService.adminSchema}.[group] g
+  LEFT OUTER JOIN ${sqlService.adminSchema}.pupilGroup pg 
+  ON g.id = pg.group_id
+  WHERE g.isDeleted=0
+  AND g.school_id=@schoolId
```
jon-shipley: school_id doesn't seem to appear in the diff for the CREATE TABLE [group].

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +24,99 @@ groupDataService.getGroups = async function (query) {
...
 groupDataService.getGroup = async function (query) {
   return Group.findOne(query).lean().exec()
 }
 
+/**
+ * Get group by id.
+ * @param groupId
+ * @returns {Promise<void>}
+ */
+groupDataService.sqlFindGroup = async (groupId, schoolId) => {
```
jon-shipley: Suggest renaming this to something like: `sqlFindOneById` as per our convention

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +24,99 @@ groupDataService.getGroups = async function (query) {
...
+  const result = await sqlService.query(sql, params)
+  return R.head(result)
+}
+
+/**
+ * Get group by name.
+ * @param groupName
+ * @returns {Promise<void>}
+ */
+groupDataService.sqlFindGroupByName = async (groupName, schoolId) => {
```
jon-shipley: Suggest renaming this to `sqlFindOneByName` as per our usual naming conventions.

school_id doesn't seem to appear in the diff for the CREATE TABLE [group].  If it was in there this query would be safer if there was an index on (school_id, name)

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +161,199 @@ groupDataService.update = async function (id, data) {
...
+
+    -- Use RAISERROR inside the CATCH block to return
+    -- error information about the original error that
+    -- caused execution to jump to the CATCH block.
+    RAISERROR (@ErrorMessage, -- Message text.
+               @ErrorSeverity, -- Severity.
+               @ErrorState -- State.
+               );
+END CATCH
+  `
```
jon-shipley: 👍 much better with a transaction

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +161,199 @@ groupDataService.update = async function (id, data) {
...
+  console.log(params)
```
jon-shipley: debug? remove?

admin/services/data-access/group.data.service.js
```diff
@@ -65,8 +161,199 @@ groupDataService.update = async function (id, data) {
...
+groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
+  const params = [
+    {
+      name: 'groupId',
+      value: groupId,
+      type: TYPES.Int
+    }
+  ]
+  const sql = `
+  BEGIN TRY
```
jon-shipley: If this transaction block is duplicated (as it seems to be) why not extract it into a wrapper and put it in `sqlService`?

admin/services/group.service.js
```diff
@@ -1,49 +1,79 @@
...
+ * Update group (group and pupils assigned to groups).
+ * @param id
+ * @param group
+ * @returns {Promise<boolean>}
+ */
+groupService.update = async (id, group, schoolId) => {
+  if (!id || !group || !group.name || !schoolId) {
+    throw new Error('id, group.name and schoolId are required')
+  }
+  return new Promise(async (resolve, reject) => {
```
jon-shipley: I don't think this Promise is required as this is already an async function

admin/services/data-access/group.data.service.js
GuyHarwood: we cant have an index on school_id & name due to incompatible data types :/

GuyHarwood: regarding index -this will be added now we have reduced col length on name

admin/services/data-access/group.data.service.js
GuyHarwood: will add

admin/services/data-access/group.data.service.js
GuyHarwood: new script as discussed

admin/services/data-access/group.data.service.js
GuyHarwood: done

admin/services/data-access/group.data.service.js
GuyHarwood: done

admin/data/sql/migrations/038.do.create-pupilGroup.sql
```diff
@@ -0,0 +1,22 @@
...
+  ) WITH (
+    PAD_INDEX = OFF,
+    STATISTICS_NORECOMPUTE = OFF,
+    IGNORE_DUP_KEY = OFF,
+    ALLOW_ROW_LOCKS = ON,
+    ALLOW_PAGE_LOCKS = ON
+  )
+)
+
+--`GRANT DELETE ON [mtc_admin].[checkFormWindow] TO ${config.Sql.Application.Username}`
```
jon-shipley: Why is this commented out?  Can it be removed?

admin/services/data-access/group.data.service.js
```diff
@@ -20,17 +24,99 @@ groupDataService.getGroups = async function (query) {
...
 groupDataService.getGroup = async function (query) {
   return Group.findOne(query).lean().exec()
 }
 
+/**
+ * Get group by id.
+ * @param groupId
+ * @returns {Promise<void>}
+ */
+groupDataService.sqlFindGroupById = async (groupId, schoolId) => {
```
jon-shipley: I would prefer it to be name `sqlFindOneById`  - the one indicates it returns an object and not an array.  The `Group` bit is redundant as this is the GroupDataService.

admin/services/data-access/sql.service.js
```diff
@@ -414,4 +414,38 @@ sqlService.buildParameterList = (ary, type) => {
   }
   return {params, paramIdentifiers}
 }
+
+sqlService.modifyWithTransaction = async (sqlStatements, params) => {
```
jon-shipley: Looks good - many thanks for creating this 👍

admin/data/sql/migrations/038.do.create-pupilGroup.sql
GuyHarwood: thats odd, maybe this came in from main DLR branch refresh. will take a look

GuyHarwood:
@jon-shipley comment was deleted, should be showing up in diff

PR #395, opened 2018-01-18
Bug/18679 promise resolution issues

GuyHarwood:
Various bug fixes for tests and components that were silently failing during test runs.

There are inconsistencies in tests and promise handling that we need to address as tech debt after the feb trial.

### DIFF
--- a/admin/spec/mocks/check-windows.js
+++ b/admin/spec/mocks/check-windows.js
@@ -3,27 +3,27 @@ const moment = require('moment')
 module.exports = [
   {
     id: 1,
-    checkEndDate: moment('2018-01-20 00:00:00').toDate(),
-    checkStartDate: moment('2018-01-10 00:00:00').toDate(),
-    adminStartDate: moment('2017-10-18 00:00:00').toDate(),
+    checkEndDate: moment('2018-01-20 00:00:00'),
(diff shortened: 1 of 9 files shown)


admin/spec/mocks/check-windows.js
```diff
@@ -3,27 +3,27 @@ const moment = require('moment')
...
     name: 'Test window 3',
     formCount: 2,
     isDeleted: false
   },
   {
     id: 2,
-    checkEndDate: moment('2017-11-28 00:00:00').toDate(),
-    checkStartDate: moment('2017-10-25 00:00:00').toDate(),
-    adminStartDate: moment('2017-10-19 00:00:00').toDate(),
+    checkEndDate: moment('2017-11-28 00:00:00'),
```
GeorgeChatzigiannis: Shouldn't this be moment('2017-11-28 00:00:00').utc()
Otherwise it would the utc flag on the moment object is false

GuyHarwood: it doesnt matter in the context of the tests.
tbh - hard coding dates in tests is an utterly awful thing to do.
that test in particular could be better.
i have created a bug card to pick this up later today

PR #402, opened 2018-01-22
Feature/18298 pupil data service refactor part 3

jon-shipley:
Refactor tests for pinService.getPupilsWithActivePins()


admin/services/data-access/pupil.data.service.js
```diff
@@ -313,6 +313,7 @@ pupilDataService.sqlFindPupilsWithActivePins = async (dfeNumber) => {
   AND s.dfeNumber = @dfeNumber
   AND p.pinExpiresAt IS NOT NULL
   AND p.pinExpiresAt > GETUTCDATE()
+  ORDER BY p.lastName ASC, p.foreName ASC
```
GeorgeChatzigiannis: what if both lastname and forename are the same?
Shouldn't we additionally include p.middleNames(if exist) in that case?

jon-shipley: I think sorting by 2 fields is enough, at least for now.

PR #411, opened 2018-01-23
Fixes for count methods to correctly verify pin validation

GeorgeChatzigiannis:


admin/services/data-access/pupil-restart.data.service.js
```diff
@@ -109,7 +110,8 @@ pupilRestartDataService.sqlFindLatestRestart = async function (pupilId) {
     }
   ]
   const result = await sqlService.query(sql, params)
-  return R.head(result)
+  const count = R.head(result)
```
jon-shipley: This is a restart isn't it?

GeorgeChatzigiannis: its a count of restarts for a pupil_id

GeorgeChatzigiannis: it is yes, fixed

PR #413, opened 2018-01-23
Bug/18524 settings not stored as decimal

GuyHarwood:
allowed question time limit and loading time limit to have point values


activemq:
@GuyHarwood @jon-shipley - I have just tested this and it is still failing. 
      expected: "5.56"
      got: "5"

GuyHarwood:
@activemq did you run the migration script update? this should all decimals

PR #417, opened 2018-01-24
Feature/18016 update reasons

pris54:
Update two reasons for pupils not to take the check.


admin/data/sql/migrations/046.do.update-reasons.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE [mtc_admin].[attendanceCode] SET reason = 'Just arrived with EAL' where reason = 'Just arrived';
```
GeorgeChatzigiannis: this needs to be renamed as number 50

pris54: @GeorgeChatzigiannis 46 was available and I added this entry myself =) Somebody overwrote my entry. I'll update the number

PR #425, opened 2018-01-24
pupil status fixes

GeorgeChatzigiannis:


admin/services/data-access/check.data.service.js
```diff
@@ -236,7 +236,27 @@ checkDataService.sqlCreate = async function (check) {
 }
 
 /**
- * Find the latest check for pupil
+ * Find the latest check for pupil that is flagged as not started
+ * @param pupilId - the pupil taking the check
+ * @return {Promise.<void>} - lean Check objects
+ */
+checkDataService.sqlFindLastCheckByPupilId = async function (pupilId) {
+  let sql = `SELECT TOP 1 * FROM ${sqlService.adminSchema}.[check] WHERE pupil_id = @pupilId
```
jon-shipley: sql could be const?

admin/services/data-access/check.data.service.js
```diff
@@ -236,7 +236,27 @@ checkDataService.sqlCreate = async function (check) {
...
+    AND startedAt IS NULL ORDER BY createdAt DESC`
```
jon-shipley: Ordering by `createdAt` will give the order that the pupil logged in.  This isn't particularly clear from the SQL though. Perhaps we need to store that too?  This may be fine for what you want.

PR #431, opened 2018-01-25
Feature/18869 groups for pupil not taking check

pris54:
Integration of groups in feature 'reasons for pupils to not take the check'.
PENDING: Failing unit test.


admin/controllers/school.js
```diff
@@ -363,7 +371,8 @@ const getSelectPupilNotTakingCheck = async (req, res, next) => {
     pupilsList: pupils,
     htmlSortDirection,
     arrowSortDirection,
-    highlight: []
+    highlight: [],
```
GeorgeChatzigiannis: If you are always passing down an empty array for the highlight, shouldn't you instead fallback in the corresponding ejs like this `highlight || []`

pris54: @GeorgeChatzigiannis I didn't look into that code, all I did was adding a comma! Let's please move any refactoring / improvements to a separate PR.

PR #435, opened 2018-01-26
Feature/18915 sql service dynamic parameter generation support decimal precision

jon-shipley:


admin/services/data-access/sql.service.js
```diff
@@ -240,11 +248,20 @@ sqlService.modify = (sql, params) => {
...
-        if (param.precision) {
-          options.scale = param.scale
-          options.precision = param.precision
+        if (R.pathEq(['type', 'name'], 'Decimal', param) ||
+          R.pathEq(['type', 'name'], 'Numeric', param)) {
+          options.precision = param.precision || 28
+          options.scale = param.scale || 5
         }
-        request.addParameter(param.name, param.type, param.value, options)
+        const opts = param.options ? param.options : options
```
GeorgeChatzigiannis: can be simplified as `const opts = param.options || options`

PR #436, opened 2018-01-26
feature/18868 - Groups for pupil register

pris54:
Populate groups column in pupil's list.

### DIFF
--- a/admin/assets/stylesheets/application.scss
+++ b/admin/assets/stylesheets/application.scss
@@ -416,6 +416,8 @@ table.spacious {
   }
 
   th, td {
+    display: table-cell;
+
     &:first-child {
       white-space: nowrap;
     }
--- a/admin/controllers/school.js
+++ b/admin/controllers/school.js
@@ -44,12 +44,12 @@ const getPupils = async (req, res, next) => {
   const order = JSON.parse(sortOrder)
   res.locals.sortOrder = typeof order === 'boolean' ? !order : true
   res.locals.sortClass = order === false ? 'sort up' : 'sort'
+
   let pupilsFormatted
   let groupsIndex = []
 
   try {
-    const groups = await groupService.getGroups(req.user.schoolId)
-    groups.map((obj) => { groupsIndex[obj.id] = obj.name })
+    groupsIndex = await groupService.getGroupsAsArray(req.user.schoolId)
   } catch (error) {
     next(error)
   }
--- a/admin/controllers/school.js
+++ b/admin/controllers/school.js
@@ -431,7 +431,7 @@ const removePupilNotTakingCheck = async (req, res, next) => {
 
     // Update the flash message for the user on the next screen
     const pupil = await pupilDataService.sqlFindOneBySlugAndSchool(pupilSlug, req.user.School)
-    req.flash('info', `Reason removed for pupil ${pupil.lastName}, ${pupil.foreName}`)
+    req.flash('info', `Reason removed for ${pupil.lastName}, ${pupil.foreName}`)
 
     // Send the information required for highlighting
     const highlight = JSON.stringify(pupilSlug)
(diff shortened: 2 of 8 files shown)


admin/assets/stylesheets/application.scss
```diff
@@ -416,6 +416,8 @@ table.spacious {
   }
 
   th, td {
+    display: table-cell;
```
pris54: this line fixes the spacing and aligning problems that appear when a table cell is highlighted (green left border)

admin/services/data-access/pupil.data.service.js
```diff
@@ -171,10 +171,10 @@ pupilDataService.sqlFindPupilsByDfeNumber = async function (dfeNumber, sortDirec
       sortBy = 'lastName'
   }
   const sql = `
-      SELECT 
-        p.*    
+      SELECT p.*, g.group_id 
       FROM ${sqlService.adminSchema}.${table} p 
       INNER JOIN school s ON s.id = p.school_id
+      LEFT JOIN ${sqlService.adminSchema}.[pupilGroup] g ON p.id = g.pupil_id
```
GeorgeChatzigiannis: just one question here: How is the query behaving if you don't have an assigned group?

GuyHarwood: thats what the LEFT JOIN is for

admin/controllers/school.js
```diff
@@ -45,14 +46,22 @@ const getPupils = async (req, res, next) => {
   res.locals.sortOrder = typeof order === 'boolean' ? !order : true
   res.locals.sortClass = order === false ? 'sort up' : 'sort'
   let pupilsFormatted
+  let groupsIndex = []
+
+  try {
+    const groups = await groupService.getGroups(req.user.schoolId)
```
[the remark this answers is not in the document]
pris54: @GuyHarwood we agreed it returns the id as the dfeNumbe was already available.

admin/services/data-access/pupil.data.service.js
pris54: @GeorgeChatzigiannis the left join ensures we still get a successful result, if there are no pupils assigned to groups you'll get a group_id column with null values.

GeorgeChatzigiannis: @pris54 ok just wanted to confirm if it was tested and working as intended basically

admin/controllers/school.js
```diff
@@ -45,14 +45,22 @@ const getPupils = async (req, res, next) => {
...
+    groups.map((obj) => { groupsIndex[obj.id] = obj.name })
```
jon-shipley: I doubt this line should be in the controller.  The service should do all the work required for the controller.  The service call can make a generic service call then and marshal the result.

pris54: @jon-shipley getGroups is used in other places, it is only in this case that the mapping is needed. If you wish I can create a new method in the service just to call getGroups and do the mapping? Please confirm.

jon-shipley: Yes - that sounds like the best approach. 👍

pris54: @jon-shipley Done.

admin/controllers/school.js
```diff
@@ -44,15 +44,23 @@ const getPupils = async (req, res, next) => {
   const order = JSON.parse(sortOrder)
   res.locals.sortOrder = typeof order === 'boolean' ? !order : true
   res.locals.sortClass = order === false ? 'sort up' : 'sort'
+
   let pupilsFormatted
+  let groupsIndex = []
+
+  try {
```
GeorgeChatzigiannis: wouldn't it make more sense to place the single line this await call(below) within the try catch block below?
that would also reduce the code a bit.

PR #437, opened 2018-01-29
File upload fixes

jon-shipley:
# File upload fixes

Fix server side by adding mime type `octet-stream` for Chrome

Add `encoding="multipart/form-data" to fix IE Edge

Update client-side duplicate submission handling to fix IE8,9,10,11


admin/views/test-developer/upload-new-form.ejs
```diff
@@ -67,13 +67,13 @@
 
         <h2 class="heading-medium">Upload file(s)</h2>
 
-        <form action="/test-developer/upload-new-form" method="post" enctype="multipart/form-data" id="upload-form">
+        <form action="/test-developer/upload-new-form" method="post" enctype="multipart/form-data" encoding="multipart/form-data" id="upload-form">
```
GeorgeChatzigiannis: @jon-shipley should we include the encoding attribute then in the rest of the areas where we upload forms?

jon-shipley: @GeorgeChatzigiannis - It's a specific hack for IE Edge.  It's not standard HTML.  It may help if the form doesn't upload in IE Edge.

admin/assets/javascripts/custom-file-upload.js
```diff
@@ -25,8 +26,12 @@ $(function () {
     })
 
     $formElement.submit(function () {
+      if (isSubmitted) {
+        return false
+      }
       $formElement.val('Sending...')
-      $(this).attr('disabled', 'disabled')
```
pris54: @jon-shipley wondering why this was removed?

jon-shipley: Hi @pris54 

Good question :)

This line was identified as causing IE8-11 to fail to upload the csv file.  It would supply a multi-part boundary but no actual content.  This was checked in detail  (I installed a proxy). My hypothesis is that on a fast machine IE would start the upload, the control would be disabled, and IE would *notice* and fail to send the file.

I assumed that this line was in there to prevent double-clicking submitting the form twice, and came up with an alternative using a boolean flag.

pris54: @jon-shipley I knew you would have a good reason to remove it =) And yes, it's sole purpose was to disable the submit button after the first submission.

PR #443, opened 2018-01-30
updated admin tests

activemq:


admin/test/rerun.txt
```diff
@@ -0,0 +1 @@
+features/generate_pupil_pins.feature:96
```
jon-shipley: Don't think this line is meant to be checked in?

PR #445, opened 2018-01-30
Prepare the check when generating pins, rather than on pupil login

jon-shipley:
Move the time we create the check from pupil login to pin-generation.


admin/services/check-form.service.js
```diff
@@ -48,10 +48,9 @@ const checkFormService = {
 
   /**
    * Extract the questions from the check-form, and add an `order` property.
-   * @param checkForm
+   * @param questions
    */
-  prepareQuestionData: function (checkForm) {
-    const questions = JSON.parse(checkForm.formData)
```
GuyHarwood: hard to read the flow in diff view, but i take it you are parsing the raw JSON string contained in `formData` elsewhere?

admin/spec/controllers/pupil-pin.spec.js
```diff
@@ -5,15 +5,16 @@ const httpMocks = require('node-mocks-http')
...
 require('sinon-mongoose')
 
+const checkStartService = require('../../services/check-start.service')
 const config = require('../../config')
+const dateService = require('../../services/date.service')
+const pinGenerationService = require('../../services/pin-generation.service')
+const pinService = require('../../services/pin.service')
+const pupilDataService = require('../../services/data-access/pupil.data.service')
+const qrService = require('../../services/qr.service')
 const School = require('../../models/school')
```
GuyHarwood: how come we still reference this model and need sinon-mongoose?

admin/spec/controllers/questions.spec.js
```diff
@@ -128,7 +132,7 @@ describe('Questions controller', () => {
         await controller.getQuestions(req, res)
       } catch (error) {
         console.error(error)
-        expect('questions controller not to').toBe('error')
+        fail('not expected to throw')
```
GuyHarwood: 👌

admin/spec/service/check-form.service.spec.js
```diff
@@ -40,26 +39,25 @@ describe('check-form.service', () => {
 
   describe('#allocateCheckForm - Happy path', () => {
     beforeEach(() => {
-      service = setupService(function () { return Promise.resolve([checkFormMock]) })
+      service = require('../../services/check-form.service')
```
GuyHarwood: much nicer

admin/spec/service/check-form.service.spec.js
```diff
@@ -40,26 +39,25 @@ describe('check-form.service', () => {
...
       }
+      done()
     })
 
     describe('#prepareQuestionData()', () => {
       it('should prepare the question data', async (done) => {
         try {
           const checkForm = await service.allocateCheckForm()
-          const questions = service.prepareQuestionData(checkForm)
+          const questions = service.prepareQuestionData(JSON.parse(checkForm.formData))
```
GuyHarwood: dont forget that the db contains raw stringified JSON

admin/services/check-form.service.js
jon-shipley: Yes, it's parsed in `checkStartService.pupilLogin` before it is sent back to the controller

admin/spec/controllers/pupil-pin.spec.js
jon-shipley: Good spot.  Looks like a mistake.  I've removed the mongoose model now.

admin/services/check-start.service.js
```diff
@@ -1,46 +1,89 @@
...
+    throw new Error('CheckForm not found: ' + check.checkForm_id)
+  }
+  const checkForm = R.head(res)
+  const checkData = {
+    id: check.id,
+    pupilLoginDate: moment.utc()
+  }
+
+  await checkDataService.sqlUpdate(checkData)
+  const questions = JSON.parse(checkForm.formData)
```
GeorgeChatzigiannis: what happens if formData is undefined?
that will throw an error on parsing.
The error will be prevented if you do this instead:
`const questions =  checkForm.formData && JSON.parse(checkForm.formData)`

GuyHarwood: checkForm.formData is non nullable

GeorgeChatzigiannis: fair enough

admin/services/check-start.service.js
```diff
@@ -1,46 +1,100 @@
...
+  if (!dfeNumber) {
+    throw new Error('dfeNumber is required')
+  }
+
+  // Validate the incoming pupil list to ensure that the pupils are real ids
+  // and that they belong to the user's school
+  const pupils = await pupilDataService.sqlFindByIdAndDfeNumber(pupilIds, dfeNumber)
+  const providedSet = new Set(pupilIds)
+  const databaseSet = new Set(pupils.map(p => p.id.toString())) // the incoming form params are stringified numbers
+  const difference = new Set([...providedSet].filter(x => !databaseSet.has(x)))
```
GeorgeChatzigiannis: What's the case of this going wrong?
Only case I could think of is if someone modifies the pupilIds that are being passed from UI.

GeorgeChatzigiannis: If that's the case then this should be reusable in other areas where pupil ids are being submitted.

admin/spec/service/check-complete.service.spec.js
```diff
@@ -55,8 +55,8 @@ describe('check-complete.service', () => {
     xit('adds a timestamp', async (done) => {
       // TODO: Relocate this test once timestamp recording occurs in application service tier
       await service.completeCheck(completedCheck)
-      const args = spy.calls.mostRecent().args[0]
-      expect(args.hasOwnProperty('receivedByServerAt')).toBe(true)
+      // const args = spy.calls.mostRecent().args[0]
```
GeorgeChatzigiannis: is that test needed?
its already excluded and most of it commented out.

jon-shipley: It was commented out as the linter didn't like `spy` which is not defined.

admin/services/check-start.service.js
jon-shipley: Yes, it could be generic.  But do we need it yet?

admin/test/features/step_definitions/generate_pupil_pins_steps.rb
```diff
@@ -195,6 +195,13 @@
   expect(generate_pupil_pins_page.find_pupil_row(@pupil_name).pin.text).to eql pupil_pin.to_s
 end
 
+Then(/^check form should be assigned to the pupil$/) do
+  pupil_upn = @stored_pupil_details['upn'].to_s
+  pupil_id = SqlDbHelper.pupil_details(pupil_upn)['id']
+  check_entry = SqlDbHelper.check_details(pupil_id)
+  expect(check_entry['checkForm_id'].nil?).to be_falsey, "Check Form is not assigned to the Pupil when pin is generated"
+end
```
activemq: @sinhaamit1980 To make this test potentially even better, is it worth maybe going to the 'checkForm' table with the check_entry['checkForm_id'] and the check assigned to the pupil is not deleted?

admin/test/features/generate_pupil_pins.feature
```diff
@@ -171,3 +171,7 @@ Feature: Generate Pupil PINs
     Given I have generated pins for all pupils in a group
     When a pupil becomes available for pin generation again
     Then I should be able to filter by groups on the generate pins page
+
```
activemq: @sinhaamit1980 maybe add a test for when there are no check forms? Just to ensure nothing breaks?

admin/services/check-start.service.js
GeorgeChatzigiannis: Unless you want to refactor this to a separate method later I believe yes.
Sounds like it should be extracted as a separate service method under pupil service?

jon-shipley: It's not pupil specific - it needs:--

param1: array of stringified numbers which is the incoming param list
param2: array of objects with Ids from the database to check against

Therefore this could be more generic: e.g. `setValidationService` perhaps? I'm open to better suggestions.

We also have to decide if want to accept the stringified versions, or wether the duty is now on the caller to convert to numbers, as the database is storing Ints for the `ids`?

GeorgeChatzigiannis: Yea it does sounds like a better approach.
I am guessing the caller would convert to numbers.

admin/services/check-form.service.js
GuyHarwood: 👍

PR #453, opened 2018-02-01
Feature/18866 groups for restarts

pris54:
Group filtering for pupil restarts feature.
It includes refactoring to make the group filter more generic.
Last commit seems unrelated, it includes linting fixes.


admin/services/group.service.js
```diff
@@ -86,4 +86,19 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
...
+/**
+ * Find groups that have pupils that can get PINs assigned.
+ * @param schoolId
+ * @param pupils
+ * @returns {Promise<*>}
+ */
+groupService.filterGroupsByPupil = async (schoolId, pupils) => {
+  let pupilIds = ''
+  if (pupils.length > 0) {
+    pupils.map(p => { if (p.group_id) pupilIds += "'" + p.group_id + "', " })
```
GeorgeChatzigiannis: i am abit confused with this if block.
What exactly is it supposed to be doing?

pris54: If you follow-up the code you'll see that it concatenates values to be used in an IN statement in the data access service.

jon-shipley: It needs to use parameters on the insert if it's user-supplied data.  Please confirm?  You may find `sqlService.buildParameterList` list useful.

GuyHarwood: @pris54 this work should be done in the data service.  Pass an array to the data service and build a set of parameters.  see other data services for reference.  passing a string to be used in a query is dangerous

GuyHarwood: @pris54 see `checkWindowDataService.sqlFindCheckWindowsAssignedToForms` for a reference on how to do it correctly.  the `groupDataService.sqlFindGroupsByIds` method should never have made it to master.  we obviously didnt catch this in a previous review

pris54: @GuyHarwood Changes done.

admin/services/group.service.js
```diff
@@ -93,12 +93,11 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
  * @returns {Promise<*>}
  */
 groupService.filterGroupsByPupil = async (schoolId, pupils) => {
-  let pupilIds = ''
-  if (pupils.length > 0) {
-    pupils.map(p => { if (p.group_id) pupilIds += "'" + p.group_id + "', " })
-    pupilIds = pupilIds.length > 0 ? pupilIds.substr(0, pupilIds.length - 2) : ''
+  if (pupils.length < 1) {
+    return false
```
GuyHarwood: is returning `false` significant in any way?  if not, just return, or just dont bother calling this method in the first place if there are no pupils to process...

admin/services/group.service.js
```diff
@@ -93,12 +93,11 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
...
   }
-  return groupDataService.sqlFindGroupsByIds(schoolId, pupilIds)
+  console.log('------ PUPILS', pupils)
```
GuyHarwood: please use winston or remove this

admin/services/group.service.js
```diff
@@ -86,4 +86,18 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
...
+ * Find groups that have pupils that can get PINs assigned.
+ * @param schoolId
+ * @param pupils
+ * @returns {Promise<*>}
+ */
+groupService.filterGroupsByPupil = async (schoolId, pupils) => {
+  if (pupils.length < 1) {
+    return false
+  }
+  console.log('------ PUPILS', pupils)
```
jon-shipley: debug? Please remove.

admin/services/data-access/group.data.service.js
```diff
@@ -309,11 +309,29 @@ groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
...
+        whereClause += ','
+      }
+      params.push({
+        name: `p${index}`,
+        value: ids[index].group_id,
+        type: TYPES.Int
+      })
+    }
+  }
+  whereClause = whereClause + ')'
```
jon-shipley: There is a where clause helper for this: sqlService.buildParameterList.  No need to maintain code in more than one place.

admin/services/data-access/group.data.service.js
```diff
@@ -309,11 +309,29 @@ groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
...
+  sql = sql + whereClause
```
jon-shipley: This isn't good practise.  Better to create a new string than modify an existing one.  Most other data services avoid this for the most part.

admin/services/group.service.js
```diff
@@ -86,4 +86,18 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
   return newGroup.insertId
 }
 
+/**
+ * Find groups that have pupils that can get PINs assigned.
+ * @param schoolId
+ * @param pupils
+ * @returns {Promise<*>}
+ */
+groupService.filterGroupsByPupil = async (schoolId, pupils) => {
```
jon-shipley: Should this be called `findGroupsByPupil`?  It doesn't seem to be filtering at all.  And is it taking the whole of a pupil object as input? Is the whole object required or would an array of pupilIds work?

admin/services/pin-generation.service.js
```diff
@@ -63,12 +63,10 @@ pinGenerationService.getPupils = async (dfeNumber, sortField, sortDirection) =>
...
 pinGenerationService.filterGroups = async (schoolId, pupils) => {
-  let pupilIds = ''
-  if (pupils.length > 0) {
-    pupils.map(p => { if (p.group_id) pupilIds += "'" + p.group_id + "', " })
-    pupilIds = pupilIds.length > 0 ? pupilIds.substr(0, pupilIds.length - 2) : ''
+  if (pupils.length < 1) {
+    return false
   }
-  return groupDataService.sqlFindGroupsByIds(schoolId, pupilIds)
+  return groupDataService.sqlFindGroupsByIds(schoolId, pupils)
```
jon-shipley: It looked better to me as pupilId.   Is the whole pupil object required?  Why?

admin/services/data-access/group.data.service.js
```diff
@@ -309,11 +309,29 @@ groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
  */
 groupDataService.sqlFindGroupsByIds = async (schoolId, groupIds) => {
   if (!schoolId || !groupIds || groupIds.length < 3) return false
-  const sql = `SELECT id, name 
+
+  let sql = `SELECT id, name 
     FROM ${sqlService.adminSchema}.[group] 
-    WHERE school_id=${schoolId} 
-    AND id IN (${groupIds})`
+    WHERE school_id=${schoolId}`
```
jon-shipley: This MUST be parameterised.

admin/services/data-access/group.data.service.js
pris54: @jon-shipley @GuyHarwood asked me to use 'sqlFindCheckWindowsAssignedToForms' as a reference to write this method, and so I did.

admin/services/data-access/group.data.service.js
pris54: @jon-shipley @GuyHarwood asked me to use 'sqlFindCheckWindowsAssignedToForms' as a reference to write this method, and so I did.

admin/services/data-access/group.data.service.js
pris54: Done.

admin/services/group.service.js
pris54: Method renamed. 
'sqlFindGroupsByIds' is doing the clean-up needed, I would rather avoid going through this object twice.

admin/services/group.service.js
pris54: Removed.

admin/services/pin-generation.service.js
pris54: Renamed. 
Please see previous explanation regarding the argument passed.

admin/services/data-access/group.data.service.js
GuyHarwood: ...but if Jon has indicated that there is a reusable helper method that respects DRY, maybe take a look at it instead of creating yet more duplicate code???

admin/services/data-access/group.data.service.js
```diff
@@ -304,16 +304,39 @@ groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
...
+  if (!schoolId || !pupilIds || pupilIds.length < 3) return false
+
+  let sql = `SELECT id, name 
     FROM ${sqlService.adminSchema}.[group] 
-    WHERE school_id=${schoolId} 
-    AND id IN (${groupIds})`
-  const params = []
+    WHERE school_id=@schoolId`
+
+  let whereClause = ' AND id IN ('
```
jon-shipley: Please replace this where clause build out with a call to `sqlService.buildParameterList` which will reduce the amount of code in this function.

admin/spec/service/group.service.spec.js
```diff
@@ -57,6 +70,18 @@ describe('group.service', () => {
       expect(pupils).toEqual(pupilsMock)
       done()
     })
+
+    it('should return an error if schoolId is missing', async (done) => {
+      const schoolId = null
+      const groupIdToExclude = 1
+      try {
+        await groupService.getPupils(schoolId, groupIdToExclude)
+        done()
```
jon-shipley: `done()` is common to both code paths and could be moved after the try/catch block.  You also need a `fail()` check to ensure that the test fails if an error is not thrown.

Also, you don't strictly need the `done` call as you are not using callbacks, `await` makes the code execute in sequence.

admin/services/data-access/group.data.service.js
pris54: @GuyHarwood I've used it already, it's sqlService.buildParameterList

GuyHarwood: yes, this comment is 3 days old.  see above where it says show/hide outdated.

PR #454, opened 2018-02-01
feature/19056 Group index fix

pris54:


pris54:
@GuyHarwood I'm deleting this unique index, it creates a bug, please read the bug ticket and how to replicate. There is logic in place (that has been tested before and after the SQL migration) to prevent duplicates.

activemq:
Works as expected

PR #455, opened 2018-02-01
Prevent pupil from logging in when check start date is in the future

GeorgeChatzigiannis:


admin/services/check-start.service.js
```diff
@@ -21,6 +21,10 @@ const checkStartService = {
     const checkForm = await checkFormService.allocateCheckForm()
     const checkWindow = await checkWindowDataService.sqlFindOneCurrent()
 
+    if (checkWindow && moment(checkWindow.checkStartDate).isAfter(moment.utc())) {
```
jon-shipley: This will return a 500 server error.  Isn't logging in too early a client error and should result in a 4xx code?  Or perhaps server temporarily unavailable `503`?

We also need to check that the check-window has not ended. this is also an error, but a more permanent one.

I'm thinking that this check needs it's own service and service call.

admin/controllers/questions.js
```diff
@@ -23,6 +24,12 @@ const getQuestions = async (req, res) => {
   } catch (error) {
     return apiResponse.unauthorised(res)
   }
+
+  try {
+    await checkWindowService.isLogInAllowed()
```
jon-shipley: `Login` is usually one word

admin/controllers/questions.js
```diff
@@ -23,6 +24,12 @@ const getQuestions = async (req, res) => {
...
+  } catch (error) {
+    return apiResponse.sendJson(res, 'Forbidden', 403)
```
jon-shipley: 👌

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,31 @@ describe('check-window.service', () => {
...
+        try {
+          await service.isLogInAllowed()
+        } catch (error) {
+          fail('not expected to throw')
+        }
+        done()
+      })
+      it('should disallow pupil to login if outside of range', async (done) => {
+        spyOn(checkWindowDataService, 'sqlFindOneCurrent').and.returnValue({
+          checkStartDate: moment.utc().add(1, 'days'),
```
jon-shipley: I think there needs to be two tests here - 1 for the check start being out of range and 1 for the check end?

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,31 @@ describe('check-window.service', () => {
...
+          checkEndDate: moment.utc().subtract(1, 'days')
+        })
+        try {
+          await service.isLogInAllowed()
```
jon-shipley: You need to `fail()` here if it doesn't throw

admin/controllers/questions.js
```diff
@@ -31,6 +32,12 @@ const getQuestions = async (req, res) => {
     await pupilLogonEventService.storeLogonEvent(null, schoolPin, pupilPin, false, 401, 'Unauthorised')
     return apiResponse.unauthorised(res)
   }
+
+  try {
+    await checkWindowService.isLoginAllowed()
```
GuyHarwood: this will prevent anyone from using the system unless there is an active and current check window.

Can you think of any situation where this might be a problem?

GeorgeChatzigiannis: @GuyHarwood based on our discussion now separate errors are being thrown based on the scenario

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,45 @@ describe('check-window.service', () => {
...
+      it('should disallow pupil to login if checkEndDate is in the past', async (done) => {
+        spyOn(checkWindowDataService, 'sqlFindOneCurrent').and.returnValue({
+          checkStartDate: moment.utc().subtract(3, 'days'),
+          checkEndDate: moment.utc().subtract(2, 'days')
+        })
+        try {
+          await service.isLoginAllowed()
+          fail('not expected to throw')
+        } catch (error) {
+          expect(error.message).toBe('Pupil not allowed to log in')
```
GuyHarwood: i think we have misunderstood the discussion here - the point of this extra work was to indicate why they cannot login.

so instead of just `Pupil not allowed to log in` can we indicate....
- `There is no active check window`
- `The active check window has expired`
- `The active check window has not started`
These actual messges should be checked with the BAs.  if they dont want them shown to the pupil, at least store them via winston so when debugging you have a clear indication of why they cannot login

GeorgeChatzigiannis: @GuyHarwood the code you were commenting on is outdated. If you check the latest commit I made yesterday there are tests to indicate `The active check window has expired` and `The active check window has not started`.
The only part that is missing is detection of no active check window which I will cover now.

GuyHarwood: strange, i clicked on 'view changes'.

thanks for the heads up, looks much better 👍

GeorgeChatzigiannis: Yea conversation tab can be misleading sometimes in terms of latest code state.
Btw updated

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,55 @@ describe('check-window.service', () => {
...
+          checkEndDate: moment.utc().add(1, 'days')
+        })
+        try {
+          await service.isLoginAllowed()
+        } catch (error) {
+          fail('not expected to throw')
+        }
+        done()
+      })
+      it('should disallow pupil to login if there is now active window', async (done) => {
```
jon-shipley: spelling? s/now/no/ ?

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,55 @@ describe('check-window.service', () => {
...
+        spyOn(checkWindowDataService, 'sqlFindOneCurrent').and.returnValue(Promise.resolve({}))
+        try {
+          await service.isLoginAllowed()
+          fail('not expected to throw')
+        } catch (error) {
+          expect(error.message).toBe('There is no active check window')
+        }
+        done()
```
jon-shipley: You know, as we are not using callbacks, there is no need for `done` to be here.  `await` removed the need for it.

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,55 @@ describe('check-window.service', () => {
...
+      })
+      it('should disallow pupil to login if checkStartDate is in the future', async (done) => {
```
jon-shipley: It might be clearer to say something like 'it should prevent pupils logging in  if checkStartDate is in the future'  or you could go with what you have with small change '...disallow pupils logging in...'

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,55 @@ describe('check-window.service', () => {
...
+        spyOn(checkWindowDataService, 'sqlFindOneCurrent').and.returnValue(Promise.resolve({
+          checkStartDate: moment.utc().add(1, 'days'),
+          checkEndDate: moment.utc().add(2, 'days')
+        }))
+        try {
+          await service.isLoginAllowed()
+          fail('not expected to throw')
+        } catch (error) {
+          expect(error.message).toBe('The active check window has not started')
```
jon-shipley: I find it odd to refer to the check window being "active" when it has not started.  Perhaps just "There is no open check window" would be clearer.   

Even more odd, is that `checkWindowDataService.sqlFindOneCurrent` would return a checkWindow that is not open, as I would expect it to return null| undefined if there wasn't a current active check Window.

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,55 @@ describe('check-window.service', () => {
...
+        }
+        done()
+      })
+      it('should disallow pupil to login if checkEndDate is in the past', async (done) => {
+        spyOn(checkWindowDataService, 'sqlFindOneCurrent').and.returnValue(Promise.resolve({
```
jon-shipley: Are you sure that `sqlFindOneCurrent` will even return a checkWindow from the past?  It wouldn't be current if it did...

admin/spec/service/check-window.service.spec.js
GeorgeChatzigiannis: @jon-shipley done

admin/spec/service/check-window.service.spec.js
GeorgeChatzigiannis: @jon-shipley i thought I had them removed. Guess I didn't. Updated

admin/spec/service/check-window.service.spec.js
GeorgeChatzigiannis: @jon-shipley now->now fixed

admin/services/data-access/check-window.data.service.js
```diff
@@ -230,6 +230,23 @@ const checkWindowDataService = {
       ))
     }
     return Promise.all(inserts)
+  },
+  /**
+   * Find active check windows
+   * @return {Object}
+   */
+  sqlFindActiveCheckWindows: async () => {
+    const sql = `SELECT * FROM ${sqlService.adminSchema}.${table}
```
jon-shipley: Please format SQL so it is readable.

admin/spec/controllers/questions.spec.js
```diff
@@ -274,5 +283,21 @@ describe('Questions controller', () => {
...
     })
+    it('returns error if the checkWindow service throws', async (done) => {
+      const req = goodReq
+      const controller = setupController({
+        'check-window.service.hasActiveCheckWindow': function () { return Promise.reject(new Error('a mock')) }
+      })
+      try {
+        await controller.getQuestions(req, res)
+      } catch (error) {
+        expect('questions controller not to').toBe('error')
```
jon-shipley: This should be a fail() call please

admin/spec/controllers/questions.spec.js
```diff
@@ -274,5 +283,21 @@ describe('Questions controller', () => {
...
+        done()
```
jon-shipley: Do you need this `done()`.  Won't it make the callback but carry on execution anyway?

admin/spec/service/check-window.service.spec.js
```diff
@@ -189,5 +189,25 @@ describe('check-window.service', () => {
...
+        } catch (error) {
+          fail('not expected to throw')
+        }
+        done()
+      })
+      it('it should disallow pupil to login if there are no active windows', async () => {
+        spyOn(checkWindowDataService, 'sqlFindActiveCheckWindows').and.returnValue([])
+        try {
+          await service.hasActiveCheckWindow()
+          fail('not expected to throw')
```
GuyHarwood: i think this should be `expected to throw` ?

admin/services/check-window.service.js
```diff
@@ -100,6 +101,19 @@ const checkWindowService = {
...
+  /**
+   * Check if pupil is allowed to log in
+   * @param pupilId
+   */
+
+  hasActiveCheckWindow: async (pupilId) => {
+    const currentCheck = await checkDataService.sqlFindOneForPupilLogin(pupilId)
+    const hasCheck = currentCheck && Object.keys(currentCheck).length > 0
+    if (!hasCheck) throw new Error(`There is no check record for pupil id ${pupilId}`)
+    const activeCheckWindow = await checkWindowDataService.sqlFindActiveCheckWindows(currentCheck.checkWindow_id)
```
jon-shipley: According to our preferred naming scheme as this function does not have a `one` in it I suspect it returns an array rather than an object.

admin/services/data-access/check-window.data.service.js
```diff
@@ -230,6 +230,34 @@ const checkWindowDataService = {
       ))
     }
     return Promise.all(inserts)
+  },
+  /**
+   * Find active check windows
+   * @param checkWindowId
+   * @return {Object}
+   */
+  sqlFindActiveCheckWindows: async (checkWindowId) => {
```
jon-shipley: Name is plural suggesting an array but return val is a single object.  Suggest calling it `sqlFindOneActiveCheckWindow` instead, or going for plural and returning the array.

GeorgeChatzigiannis: @jon-shipley updated

PR #458, opened 2018-02-02
Bug/19041 nca tools authentication

GuyHarwood:
Refactoring work to store pem certs in azure storage, which preserves cert in original format.

This PR also includes small changes to the logDNA configuration and setup, which is now live.


GuyHarwood:
For some reason these changes cause [over 50 tests to fail](https://circleci.com/gh/DFEAGILEDEVOPS/MTC/1451)

I have created a new branch from master, and applied each change one by one, and also tried removing the logdna/azure changes.  Doesn't make a difference.

WIll come back to this on Monday with a fresh pair of eyes

GuyHarwood:
Failing tests to fix

admin/authentication/nca-tools-authentication-strategy.js
```diff
@@ -24,9 +25,16 @@ module.exports = async function (req, done) {
     loginMethod: 'nca-tools'
   }
 
+  let ncaPublicKey, mtcPrivateKey
+  try {
+    ncaPublicKey = await certificateService.getNcaPublicKey()
+    mtcPrivateKey = await certificateService.getMtcPrivateKey()
+  } catch (error) {
+    winston.error('unable to retrieve certificates:', error)
+    return done(null, false)
```
jon-shipley: This is probably a slightly wrong return.  `done(null, false)` means authentication failed.  Returning done(error, null) means that authentication had an error, which is the case here.

Other lines are also throwing an error, and if that works fine that's great too, but I would need to double-check that.

admin/services/data-access/azure-blob.data.service.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const service = {
+  getBlobText: async (containerName, blobName) => {
+    return new Promise((resolve, reject) => {
+      if (!blobService) {
+        reject(new Error('Azure Storage Connection String required'))
+      }
+      blobService.getBlobToText(containerName, blobName, (error, text) => {
+        if (error) reject(error)
```
jon-shipley: Should be `return reject` or it will drop through into the `return resolve` line immediately after.

PR #459, opened 2018-02-04
Bug/19060 remember group filtering

pris54:
Depends on PR #453 - (now merged)
Remember group filters and output when sorting lists.


admin/assets/javascripts/group-filters.js
```diff
@@ -3,36 +3,124 @@
...
+        // groupFilters.setGroupIds('add', $(this).val())
+        groupFilters.updateSortingLink('add', $(this).val())
       } else {
+        $(this).attr('data-checked', false)
         groupIds.splice($.inArray($(this).val(), groupIds), 1)
+        // groupFilters.setGroupIds('remove', $(this).val())
+        groupFilters.updateSortingLink('remove', $(this).val())
       }
-      tableRowVisibility(groupIds)
+      console.log('===== GROUP IDS', groupIds)
```
jon-shipley: This debug could be removed I suspect

admin/gulpfile.js
```diff
@@ -13,8 +13,9 @@ const jsBundleFiles = [
   './assets/javascripts/details.polyfill.js',
   './assets/javascripts/global-scripts.js',
   './assets/javascripts/jquery-modal.js',
-  './assets/javascripts/custom-file-upload.js',
-  './assets/javascripts/group-filters.js'
+  './assets/javascripts/custom-file-upload.js'
+  // ,
+  // './assets/javascripts/group-filters.js'
```
jon-shipley: Is this needed or not?

admin/services/data-access/group.data.service.js
```diff
@@ -304,16 +304,39 @@ groupDataService.sqlMarkGroupAsDeleted = async (groupId) => {
...
+      }
+      params.push({
+        name: `p${index}`,
+        value: ids[index].group_id,
+        type: TYPES.Int
+      })
+    }
+  }
+  whereClause += ')'
+  sql += whereClause
```
jon-shipley: Please use `sqlService.buildParameterList` if possible.

admin/spec/service/group.service.spec.js
```diff
@@ -71,6 +96,18 @@ describe('group.service', () => {
       expect(group).toEqual(groupMock)
       done()
     })
+
+    it('should return an error if schoolId or groupId are missing', async (done) => {
+      const groupId = '123456abcde'
+      const schoolId = null
+      try {
+        await groupService.getGroupById(groupId, schoolId)
```
jon-shipley: You need `fails` in *all* tests that are expecting a throw condition, to safe-guard against the error not being thrown.

E.g.
```
try {
   await service.doSomethingThatWeExpectToThrow(badArg)
   fail('expected to throw') // <------ this line
} catch (error) {
   expect(...)
}
```

admin/spec/service/group.service.spec.js
```diff
@@ -71,6 +96,18 @@ describe('group.service', () => {
...
+        done()
```
jon-shipley: Calling done from both code paths is unusual.  Either call it once at the end of the script, or don't have it in if it's not needed (as here)

admin/services/group.service.js
```diff
@@ -86,4 +86,17 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
...
 }
 
+/**
+ * Find groups that have pupils that can get PINs assigned.
+ * @param schoolId
+ * @param pupils
+ * @returns {Promise<*>}
+ */
+groupService.findGroupsByPupil = async (schoolId, pupils) => {
+  if (pupils.length < 1) {
```
GuyHarwood: what if schoolId is undefined?

admin/services/pin-generation.service.js
```diff
@@ -62,11 +62,9 @@ pinGenerationService.getPupils = async (dfeNumber, sortField, sortDirection) =>
  * @param pupils
  * @returns {Promise<*>}
  */
-pinGenerationService.filterGroups = async (schoolId, pupils) => {
-  let pupilIds = ''
-  if (pupils.length > 0) {
-    pupils.map(p => { if (p.group_id) pupilIds += "'" + p.group_id + "', " })
-    pupilIds = pupilIds.length > 0 ? pupilIds.substr(0, pupilIds.length - 2) : ''
+pinGenerationService.filterGroups = async (schoolId, pupilIds) => {
+  if (pupilIds.length < 1) {
```
GuyHarwood: what does return false do in this context?

admin/spec/service/group.service.spec.js
```diff
@@ -57,6 +70,18 @@ describe('group.service', () => {
       expect(pupils).toEqual(pupilsMock)
       done()
     })
+
+    it('should return an error if schoolId is missing', async (done) => {
+      const schoolId = null
+      const groupIdToExclude = 1
+      try {
+        await groupService.getPupils(schoolId, groupIdToExclude)
+        done()
```
GuyHarwood: this will silently fail if an error is not thrown.  you should do something like this `fail('error not thrown')`

admin/spec/service/group.service.spec.js
```diff
@@ -137,6 +193,28 @@ describe('group.service', () => {
       })
     })
 
+    describe('unhappy path', () => {
+      beforeEach(() => {
+        service = proxyquire('../../services/group.service', {
```
GuyHarwood: you shouldnt need proxyquire, it was only needed for mongoose.

admin/assets/javascripts/global-scripts.js
```diff
@@ -377,4 +444,42 @@ $(function () {
...
+  /**
+   * Filtering pupils by group.
+   */
+  if ($('#filterByGroup').length > 0) {
+    var groupIds = []
+    var activeGroupsIds = groupFilters.findActiveGroups()
+
+    if (activeGroupsIds) {
+      $('#filter-content').removeClass('hidden')
+      var groupIdsArr = decodeURIComponent(activeGroupsIds).split(',')
```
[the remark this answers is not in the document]
pris54: @jon-shipley Initially this filter was meant to be a hide/show solution, however we lose the filter selections when the lists are re-sorted (a GET request), the only way to pass data is the URL. This can be manipulated as any of the other URL parameters passed.

admin/assets/javascripts/group-filters.js
```diff
@@ -3,36 +3,124 @@
...
+      }
+    }
+  }
+
   /**
    * Filtering pupils by group.
    */
   if ($('#filterByGroup').length > 0) {
     var groupIds = []
+    var activeGroupsIds = groupFilterss.findActiveGroups()
```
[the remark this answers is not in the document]
pris54: @jon-shipley it seems you're reviewing old code.

admin/assets/javascripts/group-filters.js
pris54: @jon-shipley it seems you're reviewing old code.

admin/gulpfile.js
pris54: Those lines have been removed.

admin/spec/service/group.service.spec.js
pris54: Unit tests have been updated in the parent PR.

admin/spec/service/group.service.spec.js
pris54: Unit tests have been updated in the parent PR. The current coverage is 97+%

admin/spec/service/group.service.spec.js
pris54: Removed in parent PR

admin/services/pin-generation.service.js
pris54: Replaced with empty array

admin/assets/javascripts/group-filters.js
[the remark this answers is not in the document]
GuyHarwood: @pris54 Does that mean there is a newer edit in another branch? or you have updated that line of code in this branch?

admin/spec/service/group.service.spec.js
GuyHarwood: @pris54 i'm pretty confused by this old code/parent PR thing.  what's going on here?  are you managing 2 pieces of work? if so, how come we are adding this in one branch and modifying in another?

pris54: @GuyHarwood This PR depends on PR 453 which I was expecting it would be merged long ago, but it hasn't hence I'm finding myself having to maintain this dependency.

PR #461, opened 2018-02-05
base64 images for pending submission screen

GeorgeChatzigiannis:
This is to tackle the case where the app switches completely off from the network and image are required to be loaded.


pupil-spa/src/app/submission-pending/submission-pending.component.scss
```diff
@@ -1,16 +1,18 @@
-.spinner {
+img.spinner {
```
GuyHarwood: how is the spinner being loaded?

GeorgeChatzigiannis: this didn't work with css so I had to add it in html instead as base64.

GeorgeChatzigiannis: It did work in disconnected mode as well.

PR #462, opened 2018-02-05
Feature/18239 school data import

GuyHarwood:
Add 60 pupils to each of the 5 schools in UAT only


admin/data/sql/migrations/062.do.school-dfeNumber-non-unique.sql
```diff
@@ -0,0 +1,3 @@
+DROP INDEX mtc_admin.school.school_dfeNumber_uindex
+ALTER TABLE mtc_admin.school ALTER COLUMN dfeNumber NVARCHAR(20) NOT NULL
+CREATE UNIQUE INDEX school_dfeNumber_uindex ON mtc_admin.school (dfeNumber)
```
jon-shipley: Why would dfeNumber be an varchar rather than an int?

GuyHarwood: because the data dept now prefix academy trusts with `MAT`

jon-shipley: That's odd. So they are encoding two pieces of data in a single column?  I hope the remaining dfeNumber is unique.  If so, we could have new column called `isMAT` and parse it at load time.  I would prefer that query to find MATs than a prefix search on an nvarchar.

GuyHarwood: they also duplicate dfeNumbers while a school transitions to an academy.  This is a whole topic we should discuss privately.  Given the upcoming trial only contains a subset of schools, i think we can potentially sidestep most of the pain for now.  But it does need actioning during March-June

GuyHarwood: @jon-shipley ive removed this migration for now, as the school data will be done separately

admin/data/sql/not-used-reference/062.do.school-dfeNumber-non-unique.sql
```diff
@@ -0,0 +1,3 @@
+DROP INDEX mtc_admin.school.school_dfeNumber_uindex
+ALTER TABLE mtc_admin.school ALTER COLUMN dfeNumber NVARCHAR(20) NOT NULL
+CREATE UNIQUE INDEX school_dfeNumber_uindex ON mtc_admin.school (dfeNumber)
```
GeorgeChatzigiannis: new line needed

admin/data/sql/not-used-reference/062.undo.school-dfeNumber-non-unique.sql
```diff
@@ -0,0 +1,3 @@
+DROP INDEX mtc_admin.school.school_dfeNumber_uindex
+ALTER TABLE mtc_admin.school ALTER COLUMN dfeNumber INT NOT NULL
+CREATE UNIQUE INDEX school_dfeNumber_uindex ON mtc_admin.school (dfeNumber)
```
GeorgeChatzigiannis: new line needed

admin/data/sql/not-used-reference/062.do.school-dfeNumber-non-unique.sql
GuyHarwood: done

admin/data/sql/not-used-reference/062.undo.school-dfeNumber-non-unique.sql
GuyHarwood: done

GuyHarwood:
@GeorgeChatzigiannis Migration 62 is the latest.  Those files are in a sub directory for reference.
I will remove the migration number to avoid confusion.

PR #463, opened 2018-02-06
Bug/travis not working

jon-shipley:


pupil-spa/test/features/step_definitions/pin_expiry_steps.rb
```diff
@@ -16,6 +16,7 @@
   current_time = Time.now + 86400
   new_time = Time.new(current_time.year, current_time.mon, current_time.day, 22, 00, 00, "+02:00").strftime("%Y-%m-%d %H:%M:%S.%LZ")
   SqlDbHelper.set_pupil_pin_expiry(@pupil['foreName'], @pupil['lastName'], @pupil['school_id'], new_time)
+  SqlDbHelper.create_check(new_time, new_time, @pupil['id'])
```
jon-shipley: @sinhaamit1980 You don't need create check to login with the test accounts - the check will be set-up automatically from the server side.

sinhaamit1980: This is for real user. There are some tests which is trying to pull pupil from database and setting pin. For these tests we will have to add insert statement for the check table

PR #464, opened 2018-02-06
allow test accounts to login

GuyHarwood:
patch to allow test accounts to login.

This will be extracted to a service with configuration that denies test accounts unless explicitly configured as such.  Tech Debt card created for that work

### DIFF
--- a/admin/controllers/questions.js
+++ b/admin/controllers/questions.js
@@ -7,6 +7,9 @@ const configService = require('../services/config.service')
 const jwtService = require('../services/jwt.service')
 const pupilAuthenticationService = require('../services/pupil-authentication.service')
 const pupilLogonEventService = require('../services/pupil-logon-event.service')
+const checkWindowService = require('../services/check-window.service')
+const R = require('ramda')
+const winston = require('winston')
 
 /**
  * If the Pupil authenticates: returns the set of questions, pupil details and school details in json format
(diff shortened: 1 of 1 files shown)


admin/controllers/questions.js
```diff
@@ -51,6 +54,18 @@ const getQuestions = async (req, res) => {
...
+      // prepare a check instance on the fly for test accounts
+      const currentWindows = await checkWindowService.getCurrentCheckWindowsAndCountForms()
+      const firstWindow = R.head(currentWindows)
+      if (!firstWindow) {
+        const errorMessage = 'test account unable to login as no current check window available'
+        winston.warn(errorMessage)
+        await pupilLogonEventService.storeLogonEvent(data.pupil.id, schoolPin, pupilPin, false, 500, errorMessage)
+        return apiResponse.serverError(res)
+      }
+      await checkStartService.prepareCheck([data.pupil.id], data.school.dfeNumber)
```
jon-shipley: 💯

admin/controllers/questions.js
```diff
@@ -51,6 +54,18 @@ const getQuestions = async (req, res) => {
 
   // start the check
   try {
+    if (data.pupil.isTestAccount) {
```
jon-shipley: Do we need to add further checks? E.g. `NODE_ENV !== 'production'`?

GuyHarwood: I've added this to the tech debt backlog and stipulated explicit configuration to enable test accounts.

PR #471, opened 2018-02-07
18942 Handle PIN collision

GeorgeChatzigiannis:
Detect pin collision and perform pin resubmission for the set of pupils that have received already existing pins


admin/config.js
```diff
@@ -23,7 +23,8 @@ module.exports = {
   TIME_BETWEEN_QUESTIONS: 2,
   LINES_PER_CHECK_FORM: 25,
   Data: {
-    allowedWords: process.env.ALLOWED_WORDS || 'aaa,bcd,dcd,tfg,bxx'
+    allowedWords: process.env.ALLOWED_WORDS || 'aaa,bcd,dcd,tfg,bxx',
+    pinSubmissionMaxAttempts: process.env.PIN_SUBMISSION_MAX_ATTEMPTS || 15
```
GuyHarwood: I am struggling with the concept of max tries given that there are 10K possible pin combinations and fewer pupils than that in each school

admin/services/pin-generation.service.js
```diff
@@ -88,19 +92,37 @@ pinGenerationService.isValid = async (p) => {
 /**
  * Generate pupils pins
  * @param pupilsList
+ * @param dfeNumber
  * @returns {Array}
  */
-pinGenerationService.updatePupilPins = async (pupilsList) => {
-  const ids = Object.values(pupilsList || null)
+pinGenerationService.updatePupilPins = async (pupilsList, dfeNumber) => {
+  let ids = Object.values(pupilsList || null)
```
GuyHarwood: is pupilsList supposed to be an array? if so, state that in the jsDoc comments above and check for it with `Array.isArray`

admin/services/pin-generation.service.js
```diff
@@ -88,19 +92,37 @@ pinGenerationService.isValid = async (p) => {
...
     }
   })
   const data = pupils.map(p => ({ id: p.id, pin: p.pin, pinExpiresAt: p.pinExpiresAt }))
-  return pupilDataService.sqlUpdatePinsBatch(data)
+  try {
+    await pupilDataService.sqlUpdatePinsBatch(data)
+  } catch (error) {
+    // Handle duplicate pins
+    if (error.number === 2601 &&
+      pinGenerationService.pinSubmissionAttempts < pinGenerationService.pinSubmissionMaxAttempts) {
```
GuyHarwood: let's say the first attempt fails.  How many more attempts will this logic make?

admin/spec/service/pin-generation.service.spec.js
```diff
@@ -160,26 +160,100 @@ describe('pin-generation.service', () => {
...
+        const unsavedPupilsMock = pupilArray.filter(p => p.id > 3 && p.id <= 6)
+        const unsavedPupilIdsMock = unsavedPupilsMock.map(p => p.id)
+        spyOn(pupilDataService, 'sqlFindPupilsWithActivePins').and.returnValue(storedPupilsWithPins)
+        const submittedIds = pupilArray.map(p => p.id)
+        try {
+          await pinGenerationService.updatePupilPins(submittedIds, 9991999)
+        } catch (error) {
+          fail('not expected to throw')
+        }
+        expect(pupilDataService.sqlUpdatePinsBatch).toHaveBeenCalledTimes(2)
```
GuyHarwood: why are we just testing for 2 attempts?

PR #472, opened 2018-02-07
Override pin expiry on travis

GeorgeChatzigiannis:
WIP


.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
 global:
   - STD_LOG_FILE="./build.log"
   - CXX=g++-4.8
+  - OVERRIDE_PIN_EXPIRY="true"
```
GuyHarwood: see travis docs, should be in env section...
https://docs.travis-ci.com/user/environment-variables/

GeorgeChatzigiannis: I am changing this now.
@GuyHarwood Shouldn't STD_LOG_FILE also be placed in env section then?
It's expecting a value from `process.env.STD_LOG_FILE` `in config.js`

GuyHarwood: IIRC that was only used when azure linux was in preview as a workaround.

you can potentially remove it completely (and the config property) if its not used anywhere

PR #476, opened 2018-02-08
allow test user login

GuyHarwood:
move test account check to before check window check

### DIFF
--- a/admin/controllers/questions.js
+++ b/admin/controllers/questions.js
@@ -36,6 +36,23 @@ const getQuestions = async (req, res) => {
   }
 
   try {
+    if (data.pupil.isTestAccount) {
+      // prepare a check instance on the fly for test accounts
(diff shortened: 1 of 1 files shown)


admin/controllers/questions.js
```diff
@@ -35,6 +35,23 @@ const getQuestions = async (req, res) => {
     return apiResponse.unauthorised(res)
   }
 
+  try {
```
GeorgeChatzigiannis: are we not adding any test for this case?

GuyHarwood: In short - No.  Had a discussion about this earlier in the week and created a tech debt card for it.  The entire controller tests need refactoring, and there will be further work on the controller itself to rehouse these key parts of logic into one service.

its become very fragmented, with various service calls tripping over one another.   Its evolved into that over time, and just needs a good design and refactor.

GeorgeChatzigiannis: Alright

PR #482, opened 2018-02-09
Display date of birth field when required on print pins page

GeorgeChatzigiannis:


activemq:
Looks good to me 👍 
screen shot 2018-02-09 at 15 11 39

PR #483, opened 2018-02-12
Bug/19269 check window query

GuyHarwood:
Remove all dependencies on `checkWindowDataService.sqlFind` and implement as separate methods.

Fixes the bug that was causing pupil checks to be allocated to future windows


admin/services/data-access/check-window.data.service.js
```diff
@@ -155,7 +114,17 @@ const checkWindowDataService = {
    * @returns {Promise.<*|Promise.<void>>}
    */
   sqlFindCurrent: async (sortBy, sortDirection) => {
-    return checkWindowDataService.sqlFind(sortBy, sortDirection, false, true)
+    const sql = `SELECT [id], [name], adminStartDate, checkStartDate, checkEndDate, isDeleted
+                  FROM ${sqlService.adminSchema}.[checkWindow] WHERE isDeleted=0 AND
+                  (adminStartDate <=@currentTimestamp AND checkEndDate >=@currentTimestamp)`
```
GeorgeChatzigiannis: so checkStartDate is not required?

GuyHarwood: that's a good point.  There are two perspectives of 'current'.  One is that it is open for administration, the other is that it is active for pupils.

I think we need to be explicit about this, so i will go through all call sites and see what they are expecting.  ultimately it can be broken into two methods - those open for admin, and those that are open for checks to start.

GuyHarwood:
@jon-shipley good catch, re-added!

PR #486, opened 2018-02-12
bug/19059 - Bug fix.

pris54:
Re-validate sticky banner when group filters are applied to pupil's list.

### DIFF
--- a/admin/assets/javascripts/global-scripts.js
+++ b/admin/assets/javascripts/global-scripts.js
@@ -475,6 +475,11 @@ $(function () {
       groupFilters.tableRowVisibility(groupIds)
       stickyBanner.outputCheckedCheckboxes(inputStatus.countCheckedCheckboxes())
       stickyBanner.stickyBannerPositioning()
+      if (pupilsNotTakingCheck.isCheckboxChecked()) {
+        stickyBanner.toggle(true)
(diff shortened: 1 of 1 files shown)


admin/assets/javascripts/global-scripts.js
```diff
@@ -475,6 +475,11 @@ $(function () {
       groupFilters.tableRowVisibility(groupIds)
       stickyBanner.outputCheckedCheckboxes(inputStatus.countCheckedCheckboxes())
       stickyBanner.stickyBannerPositioning()
+      if (pupilsNotTakingCheck.isCheckboxChecked()) {
+        stickyBanner.toggle(true)
+      } else {
+        stickyBanner.toggle(false)
+      }
```
GeorgeChatzigiannis: Just a minor improvement that can be achieved to reduce the code:
`const isCheckboxChecked = !!pupilsNotTakingCheck.isCheckboxChecked() //handles null or undefined as well`
`stickyBanner.toggle(isCheckboxChecked)`

GuyHarwood: that 'improvement' is very subjective

GuyHarwood: i prefer the original code

PR #489, opened 2018-02-12
update copy

activemq:


admin/views/restart/restart-overview.ejs
```diff
@@ -18,8 +18,8 @@
         </header>
         <br />
         <span class="font-medium">
-            If a pupil was significantly interrupted while taking the check,
-            use this page to notify the STA that they will restart the check.
+            If a pupils progress was significantly interrupted while taking the check,
```
pris54: Wondering if this is the correct English grammar? Should it not be "If a pupil's progress"

GuyHarwood: i would double check with the business to be sure, but i think `pupil's` is short for `pupil is`

GuyHarwood: https://en.wiktionary.org/wiki/pupil

GuyHarwood: the plural of pupil is pupils

activemq: Understood, copy was passed down by BA's, double checking with them now. @GuyHarwood @pris54

activemq: @GuyHarwood @pris54 BA's agree, changing to "pupil's"

PR #490, opened 2018-02-12
19083 Google Analytics on Pupil SPA

GeorgeChatzigiannis:


pupil-spa/src/environments/environment.ts
```diff
@@ -19,5 +19,6 @@ export const environment = {
   // Minimum display time for submission pending view
   submissionPendingViewMinDisplay: 2000,
   // Contact number in case check submission fails
-  supportNumber: '0345 278 8080'
+  supportNumber: '0345 278 8080',
+  googleAnalyticsTrackingCode: 'UA-92976848-11'
```
jon-shipley: Is this a non-prod env?

GeorgeChatzigiannis: is non-prod and needs to be changes
@GuyHarwood any thoughts on the dev value of googleAnalyticsTrackingCode?

GuyHarwood: As per our discussion- it needs to be disabled by default, and only active when a tracking code is present in the environment file. When deployed it will be built in production mode, so needs to be configured as such

pupil-spa/src/app/check/check.component.ts
```diff
@@ -203,6 +227,10 @@ export class CheckComponent implements OnInit {
         this.auditService.addEntry(new CheckSubmissionPending());
         this.isWarmUp = false;
         this.viewState = 'submission-pending';
+        this.window.ga('send', {
+          hitType: 'pageview',
+          page: '/submission-pending'
```
jon-shipley: Possibly these URLs should all be under the '/check/' prefix.  Would that make more sense to GA?

pupil-spa/src/environments/environment.ts
jon-shipley: Empty string works for me.

GeorgeChatzigiannis: @GuyHarwood done

pupil-spa/src/app/check/check.component.ts
GeorgeChatzigiannis: @jon-shipley If possible I'd like to avoid the check prefix and use something else.
Regexps don't work as I was expecting on excludedRoutes and since I have excluded the keyword 'check' that won't work well.

PR #495, opened 2018-02-13
Feature/18295 ps report data service refactor to sql server

jon-shipley:


admin/app.js
```diff
@@ -29,9 +29,8 @@ const azure = require('./azure')
  * use LogDNA transport for winston if configuration setting available
  */
 const winston = require('winston')
-let logdna
 if (config.Logging.LogDna.key) {
-  logdna = require('logdna')
+  require('logdna')
```
GuyHarwood: 👍

admin/services/certificate-store.service.js
```diff
@@ -2,24 +2,18 @@ const azureBlobDataService = require('./data-access/azure-blob.data.service')
 const config = require('../config')
 
 let ncaPublicKey, mtcPrivateKey
-let callCount = 0
```
GuyHarwood: thanks for tidying this up 👍

admin/services/date.service.js
```diff
@@ -35,6 +36,7 @@ const dateService = {
 
   checkAndFormat: function (date, format) {
     if (!(date instanceof Date || date instanceof moment)) {
+      winston.warn(`Date parameter is not a Date or Moment object: ${date}`)
       return ''
```
GuyHarwood: i wonder if we should be throwing an error here?  or at least returning false 🤔

jon-shipley: Yes, I've considered that, but if we did then we wouldn't be able to directly use the output of dateService in the UI, we would have to catch errors and provide defaults.

PR #496, opened 2018-02-13
Feature/19284 copy changes

pris54:
Copy changes for the feb trial.


.travis.yml
```diff
@@ -19,6 +19,7 @@ env:
   - TEST_TYPE=admin-e2e-3
   - TEST_TYPE=pupil-e2e-1
   - TEST_TYPE=pupil-e2e-2
+  - OVERRIDE_PIN_EXPIRY="true"
```
GuyHarwood: this is not in the correct place.  it will cause a separate build with the other 5 types.  it needs to be in the global section to apply to the all parallel builds

.travis.yml
```diff
@@ -19,6 +19,7 @@ env:
...
 global:
   - STD_LOG_FILE="./build.log"
   - CXX=g++-4.8
```
GuyHarwood: this file is clearly out of date because these have been removed.  the feb trial is already up to date with master so are you working from a stale codebase?

pris54: @GuyHarwood this is a branch of the feb-trial branch which I always update before creating a new branch from it, not sure how this got in here.

GuyHarwood: ive just checked the feb trial travis file and its up to date.

i would do a pull on the feb trial branch and merge it back into your PR branch, and also double check that you definitely created the PR branch from the feb trial branch

GuyHarwood: if that all goes to plan you shouldnt need to modify the travis file at all.  give me a shout if you need another pair of eyes on it

pris54:
@GuyHarwood The issue was that .travis.yml had conflicts. These have now been resolved.

GuyHarwood:
@pris54 that's a bit odd, because you shouldnt have merge conflicts on travis.yml.  your PR branch is sourced from `support/feb-trial` and therefore the only changes should be anything new you add as part of this work.  One of the service calls looked familiar from another PR, so i suspect it was more likely to be an error in branching

PR #497, opened 2018-02-13
bug/19116 - Bug fix. Wrong path to NCA url.

pris54:


admin/controllers/authentication.js
```diff
@@ -25,8 +25,8 @@ const getSignIn = (req, res) => {
   if (req.isAuthenticated()) {
     res.redirect('/school/school-home')
   } else {
-    if (rolesConfig.NCA_TOOLS_AUTH_URL) {
-      res.redirect(rolesConfig.NCA_TOOLS_AUTH_URL)
+    if (config.NCA_TOOLS_AUTH_URL) {
```
GuyHarwood: does this work now?  you can test it on your local machine by adding any URL to the env var

pris54: @GuyHarwood - I tested it locally and it works. It worked while working on it on the first PR, then I broke it myself (argh..!) while doing some 'grepping' looking for wrong references to roles in the config file (see comments in PBI 18514) - It's been fixed and retested.

PR #498, opened 2018-02-14
bug/19297 - Bug fix

pris54:
Trim spaces in group name field for better validation.


admin/assets/javascripts/global-scripts.js
```diff
@@ -382,7 +382,7 @@ $(function () {
      * @returns {boolean}
      */
     isGroupNameComplete: function () {
-      var elName = $('input#name').val()
+      var elName = $('input#name').val().trim()
```
jon-shipley: I suspect that this won't work in IE8.

pris54: 🙄  I'll test it on IE8. However this is an edge case and it's not a critical fix because we still get a server side error on the screen.

pris54: @jon-shipley - Tests passed and @sinhaamit1980 is happy with this PR. Can we please get this approved?

GuyHarwood: Are we even supporting IE8. I can never remember 🤔

jon-shipley: @GuyHarwood - functional support for IE8-10

jon-shipley: @pris54 

Why is this a client-side fix and not server side?
We need to support IE functionally and calling methods that may not be implemented doesn't work for me.   You could use Jquery.trim() though?

GuyHarwood: In that case safest, and likely easiest would be as Jon says, server side

pris54:
@GuyHarwood @jon-shipley I'm not sure you understand what this bug is about... I can demo it to you if it helps.

admin/assets/javascripts/global-scripts.js
jon-shipley: @pris54 

Fix it?

screen shot 2018-02-14 at 14 00 50

pris54:
@GuyHarwood can I please get this PR approved? I've done the changes requested.

PR #499, opened 2018-02-14
Marking service : store the marks applied for each answer

jon-shipley:
Add two new fields to the psychometric report:

* `QnSco` - the marks applied to the answer
* `QnTimeoutSco` - time out with correct answer (1= correct, 0= incorrect)

To do this we are storing all marked answers in the `answer` table which is produced from the marking service.


admin/bin/psychometricians-report-v2.js
```diff
@@ -38,7 +40,13 @@ async function main () {
...
-  process.exit(0)
 }
 
 main()
+  .then(() => {
+    poolService.drain()
+  })
+  .catch(e => {
+    console.warn(e)
+    poolService.drain()
```
GuyHarwood: 💯

admin/services/data-access/sql.pool.service.js
```diff
@@ -29,10 +29,10 @@ let pool = null
 
 const sqlPoolService = {}
 
-  /**
-   * Initialise the connection pool.  Called once per application instance
-   * @return {Promise}
-   */
+/**
+ * Initialise the connection pool.  Called once per application instance
+ * @return {Promise}
```
GeorgeChatzigiannis: Is the return needed here?
this method doesn't seem like it's returning anything.

admin/services/data-access/sql.pool.service.js
```diff
@@ -41,10 +41,10 @@ sqlPoolService.init = () => {
   })
 }
 
-  /**
-   * Get a connection from the pool.
-   * @return {Promise}
-   */
+/**
+ * Get a connection from the pool.
+ * @return {Promise}
```
GeorgeChatzigiannis: Same here

admin/services/data-access/sql.pool.service.js
```diff
@@ -60,4 +60,11 @@ sqlPoolService.getConnection = () => {
   })
 }
 
+/**
+ * Disconnect all pool connections
```
GeorgeChatzigiannis: JSDOc return is missing here.

admin/services/data-access/sql.pool.service.js
jon-shipley: removed

admin/services/data-access/sql.pool.service.js
GuyHarwood: getConnection returns a promise, see the full code

jon-shipley: it certainly does resolve `resolve(connection)`

GeorgeChatzigiannis: ok github diff didn't help me on seeing this clearly.

PR #502, opened 2018-02-15
Bug/19404-extract-migrations-to-build-task

GuyHarwood:


deploy/sql-migrations/migrate-db.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
jon-shipley: Why do we need the same file twice?

GuyHarwood: ah yes, forgot to remove the one in migrations dir!

GuyHarwood: removed

PR #504, opened 2018-02-15
Bugfix: pupil login

jon-shipley:


admin/services/data-access/sql.service.js
```diff
@@ -178,6 +178,8 @@ sqlService.adminSchema = '[mtc_admin]'
  * @return {Promise<*>}
  */
 sqlService.query = (sql, params = []) => {
+  winston.debug(`sql.service.query(): ${sql}`)
+  winston.debug('sql.service.query(): Params ', R.compose(R.map(R.pick(['name', 'value'])))(params))
```
GuyHarwood: 🥇

GeorgeChatzigiannis: Are you 100% sure the compose is needed?

jon-shipley: @GeorgeChatzigiannis - I'm sure someone can find a way to remove `Compose`.

GeorgeChatzigiannis: Well the output seems to be the same with or without it for me.

PR #518, opened 2018-02-18
SPA: Capture device properties

jon-shipley:
Add a service to capture device properties that may affect the quality of service

screen shot 2018-02-20 at 15 38 55


pupil-spa/src/app/services/device/device.service.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  /**
+   * Warning: limited support (Chrome, FF)
+   * @return {{hardwareConcurrancy: any}}
+   */
+  getCpuInformation() {
+    return {
+      hardwareConcurrency: this.window.navigator.hardwareConcurrency
```
GeorgeChatzigiannis: I am not sure if navigator works well with ie9.

jon-shipley:
@GuyHarwood This is, so far, only the `deviceService` to capture device info.  This doesn't have a reporting component yet - first we need to capture details we anticipate may help us.

I think a good place to put this would in on successful login, so we capture details at the start of the check (e.g. we might capture some battery info) as there is a chance that if it is useful we could suggest not starting the check if, for example, the device had low battery level.  I am open to other suggestions though.

This is a WIP.  I am mostly concerned with getting some logging in.  Reporting can be added in in a future PR.  The next step is to agree on *where* to add it in I think.

jon-shipley:
@GuyHarwood @GeorgeChatzigiannis - please review

admin/bin/inspect-spa-data.js
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env node
+'use strict'
+
+const util = require('util')
+
+const poolService = require('../services/data-access/sql.pool.service')
+const completedCheckDataService = require('../services/data-access/completed-check.data.service')
+
+let usage = function () {
```
GeorgeChatzigiannis: const?

jon-shipley: Hmmm

PR #520, opened 2018-02-20
Bug/19473 side menu links

pris54:
Fixed sidebar links for all three roles.
Updated links to 'pupil register'.
New solution to download Guidandance.pdf


admin/controllers/downloadFile.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+
+module.exports = async (req, res) => {
+  res.setHeader('Content-disposition', 'filename=multiple_pupils_errors.csv')
```
GuyHarwood: should this be here?

admin/routes/school.js
```diff
@@ -41,6 +42,7 @@ const school = (router) => {
   router.post('/group-pupils/add', isAuthenticated(rolesConfig.ROLE_TEACHER), (req, res, next) => group.addGroup(req, res, next))
   router.post('/group-pupils/edit', isAuthenticated(rolesConfig.ROLE_TEACHER), (req, res, next) => group.editGroup(req, res, next))
   router.get('/group-pupils/delete/:groupId', isAuthenticated(rolesConfig.ROLE_TEACHER), (req, res, next) => group.removeGroup(req, res, next))
+  router.get('/download-guidance', isAuthenticated(rolesConfig.ROLE_TEACHER), (req, res, next) => downloadFile(req, res, next))
```
GeorgeChatzigiannis: shouldn't that be `downloadGuidance`?
`downloadFile` sounds too generic to me

admin/routes/school.js
```diff
@@ -15,6 +15,7 @@ const { getHome,
   savePupilNotTakingCheck,
   removePupilNotTakingCheck,
   viewPupilsNotTakingTheCheck } = require('../controllers/school')
+const downloadFile = require('../controllers/downloadFile')
```
GeorgeChatzigiannis: shouldn't the controller name be download-file to follow the naming convention we are using for the rest of the controller names?
Also the controller's name is to generic to me. 
For example if you are building routes for the related panel why not call it related?

admin/controllers/downloadFile.js
jon-shipley: The filename should be lowercase with hyphens: download-file.js

pris54: My mistake. It's fixed.

admin/routes/school.js
pris54: @GeorgeChatzigiannis I'm expecting to extend this controller after the feb trial with some other downloads that need refactoring, that's why the generic name.

GeorgeChatzigiannis: @pris54 why not creating a controller called related (or something more distinctive if you wish) and include a method which is very specific?

admin/routes/school.js
pris54: @GeorgeChatzigiannis The routes are download-guidance for each of the three roles. The controller will grow soon (please see previous reply) to cover other downloads, hence the generic name.
I have modified the controller's file name to lower chase and hyphen as you suggested.

GuyHarwood: if it was generic you would need to be very careful about how you pass details to the controller about what to download.  It may just be safer, and easier to maintain if the controller just serves the one file.  the current name is fine until you introduce another.

pris54:
@GeorgeChatzigiannis - This controller is to download files. In this PR it only downloads the guidance pdf but there are other files that need refactoring (have a look at the tech debt spreadsheet), I'm aiming at using this same controller for all the downloads that need refactoring hence the name.

admin/controllers/download-file.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+module.exports = async (req, res) => {
+  const fileName = 'MTC_CAG_Feb_2018_Trial.pdf'
+  res.setHeader('Content-disposition', 'filename=' + fileName)
+  res.setHeader('Content-type', 'application/pdf')
+  res.download('public/PDFs/' + fileName)
```
jon-shipley: If the path `public/PDFs/` is actually public then the file can still be downloaded even if the user is not logged in, although guessing the filename is unlikely.

Is this something we want to prevent?

GuyHarwood: come to think of it - have we tried fixing this with something as simple as `target=_blank` to force it into a new window?

pris54: @jon-shipley yes, we need to protect 'public' folder too, and I'm also planning on moving those unique file names to either config or a specific file config file. This is tech debt that I'm hoping I'll be able to address after the feb trial.

jon-shipley: @pris54 - protecting the public folder sounds unusual - normally we just serve the file from outside the web root.

Is that a part of this PR?

pris54: @jon-shipley not part of this PR. I find unusual that we need to protect files that are effectively public as they are part of this repository =) I believe the logic is that they are public to certain roles. More clarification will be needed (soon) before addressing this particular tech debt (renaming the 'public' folder? removing it completely and server from 'assets'?)

pris54: @jon-shipley that may be also a good opportunity to look more deeply into role management.

GuyHarwood: sounds like our `public` folder is more of a `static` folder.
the only question i have is whether we *really* need to serve this through express.  
Its in the github repo, so is effectively public anyway.
Has anyone indicated it should be accessible to authorised users only?

pris54: @GuyHarwood those questions are pretty much the ones I've just posted in reply to Jon's comments. I think more clarification is needed after the feb trial, I'm making a note to discuss with Paul S. and BAs in the next planning meeting.

jon-shipley: Oh yes. Good point 🤦‍♂️ 

In that case I question why we even want node to serve the files at all. Better solutions exist.  I'm happy to approve this as-is, as there is no other architecture in place at this time.

pris54: @jon-shipley Agreed. Please see the bottom note on my previous reply to Guy.

PR #521, opened 2018-02-20
bug/19494 allow nca tools users without school assigned

GuyHarwood:
NCA Tools has been updated so that Service Managers and Test Developers can login to MTC without a school assigned.  This change ensures that MTC handles the missing data item appropriately.

### DIFF
--- a/admin/spec/service/nca-tools-user.service.spec.js
+++ b/admin/spec/service/nca-tools-user.service.spec.js
@@ -103,6 +103,17 @@ describe('nca-tools-user.service', () => {
       expect(user.mtcRole).toBe('TEACHER')
       done()
     })
+
+    it('does not look up school if not provided', async (done) => {
+      spyOn(schoolDataService, 'sqlFindOneByDfeNumber')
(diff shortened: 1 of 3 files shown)


admin/spec/service/nca-tools-user.service.spec.js
```diff
@@ -103,6 +103,17 @@ describe('nca-tools-user.service', () => {
...
+
+    it('does not look up school if not provided', async (done) => {
+      spyOn(schoolDataService, 'sqlFindOneByDfeNumber')
+      spyOn(userDataService, 'sqlFindOneByIdentifier').and.returnValue(Promise.resolve({ school_id: 999 }))
+      spyOn(userDataService, 'sqlUpdateSchool').and.returnValue(Promise.resolve())
+      spyOn(roleService, 'findByTitle').and.returnValue(Promise.resolve({ id: 1 }))
+      spyOn(userDataService, 'sqlCreate').and.returnValue(Promise.resolve())
+      await ncaToolsUserService.mapNcaUserToMtcUser({})
+      expect(schoolDataService.sqlFindOneByDfeNumber).not.toHaveBeenCalled()
+      done()
```
GeorgeChatzigiannis: done shouldn't be needed since there's an await in the async test

PR #522, opened 2018-02-20
Bug/19351 pupil register sorting 2

pris54:
Additional sorting by group. Unit tests.


admin/services/pupil.service.js
```diff
@@ -103,4 +103,28 @@ pupilService.sortByStatus = (pupilsList, sortDirection) => {
   })
 }
 
+/**
+ * Sort pupil's array by group.
+ * @param pupilsList
+ * @param sortDirection
+ * @returns {*}
+ */
+pupilService.sortByGroup = (pupilsList, sortDirection) => {
```
jon-shipley: Why doesn't this sort happen in the database with nulls on top?

pris54: @jon-shipley please have a look at how this array is been created, 'groups' and 'status' (outcome) are mapped after the initial query.

pris54: Nulls on top is not the desired behavior.

PR #523, opened 2018-02-20
Add config to allow the auto-marking to be disabled

jon-shipley:
Allow config to disable the automatic marking

Testing:

`AUTO_MARK=false yarn start` to override

### DIFF
--- a/admin/app.js
+++ b/admin/app.js
@@ -48,7 +48,7 @@ azure.startInsightsIfConfigured()
 
 const unsetVars = []
 Object.keys(config).map((key) => {
-  if (!config[key] && !devWhitelist.includes(key)) {
+  if (config[key] === undefined && !devWhitelist.includes(key)) {
     unsetVars.push(`${key}`)
   }
 })
--- a/admin/config.js
+++ b/admin/config.js
@@ -1,5 +1,7 @@
 'use strict'
 const os = require('os')
+const toBool = require('to-bool')
+
 const twoMinutesInMilliseconds = 120000
 const oneMinuteInMilliseconds = 60000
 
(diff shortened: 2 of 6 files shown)


jon-shipley:
Looking for 1 review

admin/config.js
```diff
@@ -1,5 +1,7 @@
 'use strict'
 const os = require('os')
+const toBool = require('to-bool')
```
GuyHarwood: 👌 there are quite a few others that need this

admin/config.js
```diff
@@ -21,6 +23,8 @@ module.exports = {
   SESSION_SECRET: process.env.NODE_ENV === 'production' ? process.env.SESSION_SECRET : 'anti tamper for dev',
   TIME_BETWEEN_QUESTIONS: 2,
   LINES_PER_CHECK_FORM: 25,
+  // AUTO_MARK: true | false - Automatically mark the check data when we receive it: boolean
+  AUTO_MARK: process.env.hasOwnProperty('AUTO_MARK') ? toBool(process.env.AUTO_MARK) : true,
```
GuyHarwood: i would _prefer_ to move away from the shouty `ENV_VAR` format and take the approach further down the file of the newer items.  `config.AutoMarkCompletedChecks` ?

admin/app.js
```diff
@@ -48,7 +48,7 @@ azure.startInsightsIfConfigured()
 
 const unsetVars = []
 Object.keys(config).map((key) => {
-  if (!config[key] && !devWhitelist.includes(key)) {
+  if (config[key] === undefined && !devWhitelist.includes(key)) {
```
GeorgeChatzigiannis: why detect specifically for undefined?
The way it was done before it would also catch null and empty strings as well

admin/config.js
jon-shipley: I am happy to change the config object property name, but less happy to make actual environment variables camelCase.  Unix has a widely held convention that exported environment names are upper case. 

Uppercase env vars are ugly; I thought it was a feature.

Would this work for you?

`autoMark: process.env.hasOwnProperty('AUTO_MARK') ? toBool(process.env.AUTO_MARK) : true,`

admin/app.js
jon-shipley: Consider:
```
AUTO_MARK=false yarn start
```
is a valid setting.  I am converting the string 'false' to the boolean value - `false` when the config is initialised - not all config values are strings (although all environment values are strings)

So we have to allow Falsey values in the config.

GeorgeChatzigiannis: I see

GeorgeChatzigiannis: So question is how do we detect null, empty strings then (if they do appear in the config)?

jon-shipley: If they have been defined by the user as that then it's accepted.

GuyHarwood:
@jon-shipley i have tested this, and it works as expected

admin/config.js
jon-shipley: Done - other bools converted

PR #529, opened 2018-02-22
Bug/markup cleanup + bug fix

pris54:
This PR has markup fixes and tweaks so that the code validates.
There is a bug fix re' a repeated title.
It includes linting fixes too.

### DIFF
--- a/admin/views/sign-in.ejs
+++ b/admin/views/sign-in.ejs
@@ -10,7 +10,8 @@
 
         <p class="lede" id="lead-paragraph">Sign-in to access the Check Development Service</p>
 
-        <form action="/sign-in" method="post" aria-label="Sign-in below" autocomplete="off">
+        <!-- Attributes 'autocapitalize' and 'autocorrect' may not validate but are needed for our current solution -->
+        <form action="/sign-in" method="post" aria-label="Sign-in below" autocomplete="off" autocapitalize="none">
             <div class="form-group">
                 <label for="username" class="form-label-bold"> Username </label>
                 <input class="form-control"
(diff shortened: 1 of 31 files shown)


admin/views/sign-in.ejs
```diff
@@ -10,7 +10,7 @@
 
         <p class="lede" id="lead-paragraph">Sign-in to access the Check Development Service</p>
 
-        <form action="/sign-in" method="post" aria-label="Sign-in below" autocomplete="off" autocapitalize="none">
+        <form action="/sign-in" method="post" aria-label="Sign-in below" autocomplete="off">
```
jon-shipley: Please either leave autocapitalise here or move it to the input fields.  It's needed on iOS tablets.

admin/views/sign-in.ejs
```diff
@@ -20,7 +20,7 @@
                        aria-required="true"
                        aria-label="Enter your user name."
                        spellcheck="false"
-                       autocorrect="off">
+                       autocomplete="off">
```
jon-shipley: `autocorrect` is, I think, needed for iOS.

admin/views/sign-in.ejs
pris54: @jon-shipley 'autocapitalize' comes up as invalid for element <form> - it throws a validation error (not warning). I found out that it's a "nonstandard attribute used by iOS Safari Mobile" - If you need it I'll put it back with a comment in both the markup and the README file.

admin/views/sign-in.ejs
pris54: @jon-shipley same as above.

PR #532, opened 2018-02-23
Fix CSP policy for Safari

jon-shipley:
* GA required connect-src for google-analytics
* unsafe-inline error fixed, but the underlying cause needs understanding


activemq:
screen shot 2018-02-23 at 13 08 29
screen shot 2018-02-23 at 13 09 02

PR #534, opened 2018-02-26
Bugfix - add support for 'Del' key for MS

jon-shipley:
Edge is reporting the 'Delete' key as 'Del'.   This bugfix adds in support for the non-standard 'Del' key.


pupil-spa/src/app/practice-question/practice-question.component.ts
```diff
@@ -229,6 +229,7 @@ export class PracticeQuestionComponent implements OnInit, AfterViewInit {
     switch (key) {
       case 'Backspace':
       case 'Delete':
+      case 'Del':
```
pris54: So it looks that it's not only Edge using 'Del':
[2] Internet Explorer 9 and Firefox 36 and earlier use "Del" instead of "Delete" for the Del key.
Source: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

jon-shipley: Thanks @pris54 👍

PR #535, opened 2018-02-27
Feature/18197 enhance date service to incorporate millisecond precision

jon-shipley:
Store moment objects with 3 milliseconds precision in the DB

screen shot 2018-02-27 at 09 34 27


jon-shipley:
@GuyHarwood - WIP for early review and comments

admin/services/date.service.js
```diff
@@ -34,6 +35,16 @@ const dateService = {
     return this.checkAndFormat(date, timeFormatWithSeconds)
   },
 
+  formatIso8601: function (momentDate) {
```
GeorgeChatzigiannis: Why explicitly only moment objects?
You can always convert a date to moment.

jon-shipley: Hi @GeorgeChatzigiannis 

Whilst I am certainly in favour of your general idea, in this case I deliberately rejected primitive JS Dates. 

Our app works with Moment objects, not Date objects. That said, the actual issue is that we need to be able to convert Moment objects to Date objects that include the timezone offset and milliseconds.  When moment converts to a timezone object it converts it to UTC, but we don't want to store UTC in the DB - we want to store the date with the TZ component.

The path I am suggesting here is we go:

`moment -> string -> Tedious -> date -> DB`

as I *assume* that Tedious is converting the ISO date to a native Date. This is clearly not ideal, but then the JS Date library is also far from ideal.

To answer your question if we wish to store a Date Object in the DB then we should pass it through as is.  We don't need to call this function to convert it to a string as it is already a Date.

There is additional work required here, to implement this conversion in our Tedious adapter, either implicitly allowing Tedious to change the datatype, or explicitly where we convert it in our code.

admin/tests-integration/check-form.service.spec.js
```diff
@@ -18,7 +18,7 @@ describe('check-form.service', () => {
   const availableForms = []
   const seenForms = []
 
-  describe('in-memory tests of check-form.service.allocateCheckForm', () => {
+  xdescribe('in-memory tests of check-form.service.allocateCheckForm', () => {
```
GuyHarwood: how come this is disabled?

jon-shipley: 1. - it only generates a csv - it does do the analysis in the test
2. - it takes about 2 minutes to run as it runs 600k simulations

PR #538, opened 2018-02-27
feature/19053 - Display groups in pin cards.

pris54:
Display group name in pin print cards.

### DIFF
--- a/admin/controllers/pupil-pin.js
+++ b/admin/controllers/pupil-pin.js
@@ -122,6 +122,13 @@ const getGeneratedPinsList = async (req, res, next) => {
   })
 }
 
+/**
+ * Get Print PINs.
+ * @param req
+ * @param res
+ * @param next
+ * @returns {Promise<*>}
+ */
 const getPrintPins = async (req, res, next) => {
   res.locals.pageTitle = 'Print pupils'
   let pupils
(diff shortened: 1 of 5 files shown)


admin/controllers/pupil-pin.js
```diff
@@ -122,14 +122,21 @@ const getGeneratedPinsList = async (req, res, next) => {
...
+ */
 const getPrintPins = async (req, res, next) => {
   res.locals.pageTitle = 'Print pupils'
   let pupils
   let school
   let qrDataURL
   const date = dateService.formatDayAndDate(new Date())
   try {
-    pupils = await pinService.getPupilsWithActivePins(req.user.School)
+    pupils = await pinService.getPupilsWithActivePins(req.user.School, req.user.schoolId)
```
GuyHarwood: why do you need 2 identifiers for the school?

pris54: @GuyHarwood because ```getPupilsWithActivePins``` will now also get groups using ```getGroupsAsArray``` and this method needs the schoolId, not the dfeNumber.

admin/services/pin.service.js
```diff
@@ -12,12 +13,18 @@ const pinService = {}
 /**
  * Get pupils with active pins
  * @param dfeNumber
- * @returns {Array}
+ * @param schoolId (optional, defaults to null)
+ * @returns {Promise<*>}
  */
-pinService.getPupilsWithActivePins = async (dfeNumber) => {
+pinService.getPupilsWithActivePins = async (dfeNumber, schoolId = null) => {
```
pris54: second argument added as optional with a null as default

GeorgeChatzigiannis: Do you have tests to support these changes in this method?

jon-shipley: This just looks wrong.  Can we not choose one or the other please?

PR #541, opened 2018-02-28
bug/19323 - Cancel button fix [master]

pris54:
Cancel link fixed. Tiny fix.
NOTE: Awaiting BAs to confirm if this fix is needed for the feb-trial branch too.


GuyHarwood:
@activemq do we not have any test automation around this journey?

activemq:
@GuyHarwood Unfortunately no. Dont think they were added when the feature was originally added, will add them as part of this PR.

PR #542, opened 2018-02-28
create docker image from admin public assets

GuyHarwood:
builds an image from the public assets in the admin application so they can be served independently of the admin express application.


GuyHarwood:
@activemq @sinhaamit1980  this PR doesnt change *any* code within either applications.  Do we have flaky tests here?

jon-shipley:
Looks good so far.   Will it become part of the local `docker-compose` setup?

GuyHarwood:
@jon-shipley it _could_ do.  However, I have designed it in such a way that it only requests assets from a different domain when configured.  So this doesn't affect local development unless you explicitly set the `ASSET_PATH` env var.  As the admin app isn't a part of the docker-compose setup, i suggest we leave it out.

PR #546, opened 2018-03-02
disable insights error handling and logging

GuyHarwood:
this should stop app insights package from trying to process errors.
also stops it capturing logging from winston.

### DIFF
--- a/admin-assets/nginx/default.conf
+++ b/admin-assets/nginx/default.conf
@@ -5,6 +5,8 @@ server {
    # don't send the nginx version number in error pages and Server header
   server_tokens off;
 
+  error_log /dev/stdout debug;
+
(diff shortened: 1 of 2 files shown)


admin-assets/nginx/default.conf
```diff
@@ -5,6 +5,8 @@ server {
    # don't send the nginx version number in error pages and Server header
   server_tokens off;
 
+  error_log /dev/stdout debug;
```
jon-shipley: Just to confirm - is this wanted / required for ops or is this debug?

GuyHarwood: this isnt part of this PR!  this has managed to get in when traversing branches last week

PR #547, opened 2018-03-02
Update copy for the March trials (and March training)

jon-shipley:
Copy changes requested for the March trials

screen shot 2018-03-02 at 17 20 47

screen shot 2018-03-02 at 17 20 53


activemq:
Ruby changes lgtm

PR #549, opened 2018-03-05
19508 Display pins overview page errors and prevent pins generation

GeorgeChatzigiannis:
screen shot 2018-03-06 at 12 26 37 pm


admin/lib/errors/pins-overview.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+
+module.exports = {
+  'noCurrentCheckWindow': 'SM01. Please contact helpdesk',
```
jon-shipley: Possibly it might be easier if all the error codes where in a single file - we will need to provide documentation on every single one of them.

In fact - perhaps the documentation should be a part of the PR that introduces each message?

What does `SM` stand for here?  Is there a scheme?

admin/services/pin.service.js
```diff
@@ -78,4 +80,14 @@ pinService.expireMultiplePins = async (pupilIds) => {
   return pupilDataService.sqlUpdatePinsBatch(data)
 }
 
+/**
+ * Detects possible check form and check window error that should prevent pin generation
+ * @returns {String}
+ */
+pinService.getPinsOverviewError = async () => {
```
jon-shipley: Isn't this really a wrapper for the validator?   It seems a bit anaemic... perhaps it is not needed?

admin/services/pin.service.js
```diff
@@ -78,4 +80,14 @@ pinService.expireMultiplePins = async (pupilIds) => {
...
+  const checkWindow = await checkWindowDataService.sqlFindOneCurrent()
+  const allForms = await checkFormService.getAllFormsForCheckWindow(checkWindow && checkWindow.id)
```
jon-shipley: Would you really want to call this without an argument?  It calls the data-layer directly - it does not throw.

admin/services/pin.service.js
```diff
@@ -78,4 +80,14 @@ pinService.expireMultiplePins = async (pupilIds) => {
...
+  return pinsOverviewValidator.validate(checkWindow, allForms)
```
jon-shipley: 👍

admin/views/pupil-pin/generate-pins-overview.ejs
```diff
@@ -8,6 +8,16 @@
 </div>
 <div class="grid-row">
     <div class="column-two-thirds">
+        <% if (error) { %>
+        <div class="error-summary" role="group" aria-labelledby="error-summary-heading-1">
+            <h2 class="heading-medium error-summary-heading" id="error-summary-heading-1">
+                Unable to proceed. Error: <%= error %>
```
jon-shipley: Hard-coding this `Unable to proceed` part of the message seems brittle.  We should have consistent error handling in the front-end.

admin/views/pupil-pin/generate-pins-overview.ejs
```diff
@@ -43,7 +53,11 @@
         <form action="/pupil-pin/generate-pins-list/lastName/asc" name="generate-pins-list">
             <div class="form-group">
                 <div class="form-buttons">
+                    <% if (error) { %>
+                    <input type="submit" class="button" disabled value="Generate PINs"/>
```
jon-shipley: Can you confirm that `disabled` works in IE8?  Or does it need `disabled="disabled"`?

admin/services/pin.service.js
GeorgeChatzigiannis: @jon-shipley could just call a validator from the controller and add everything in there indeed.
I just thought of 1) always call a validator from a service 2)keep the validator intact in terms of data service methods and just pass down the required data to-be-validated as we do with the rest of them. If we prefer a direct call to validator that incorporates data service method calls I'll change it.

admin/services/pin.service.js
GeorgeChatzigiannis: @jon-shipley I just noticed the implementation of it.
Literally only thing it does is call to `sqlFetchSortedActiveFormsByName` data method which seems redundant as a normal service method.
In addition the data layer method returns all the forms that are not deleted if checkWindow id is not provided
 Isn't that wrong?

admin/views/pupil-pin/generate-pins-overview.ejs
GeorgeChatzigiannis: @jon-shipley I'll amend to include the common part in the errors

admin/lib/errors/pins-overview.js
GeorgeChatzigiannis: I believe it stands for Service Manager.
I'll refactor this to include the errors within a single error file as more will appear soon

admin/views/pupil-pin/generate-pins-overview.ejs
GeorgeChatzigiannis: https://www.w3schools.com/tags/att_input_disabled.asp
IE>6.0 based on w3 schools.

admin/lib/errors/pins-overview.js
GeorgeChatzigiannis: updated

admin/views/pupil-pin/generate-pins-overview.ejs
jon-shipley: 👍

admin/services/pin.service.js
jon-shipley: @GeorgeChatzigiannis If you don't have a checkWindow to put in the `getAllFormsForCheckWindow` call we need to throw an error.

admin/views/pupil-pin/generated-pins-list.ejs
```diff
@@ -9,6 +9,16 @@
 
 <div class="grid-row">
     <div class="column-two-thirds">
+        <% if (error) { %>
+        <div class="error-summary" role="group" aria-labelledby="error-summary-heading-1">
+            <h2 class="heading-medium error-summary-heading" id="error-summary-heading-1">
+                <%= error %>
+            </h2>
+            <p>Multiplication tables check helpline<br>
+                0345 278 8080
```
jon-shipley: I suspect this helpline number is used in multiple places - it should be in config though.
@GuyHarwood, @pris54  - thoughts?

admin/lib/validator/pins-overview-validator.js
```diff
@@ -0,0 +1,22 @@
+const serviceManagerErrorMessages = require('../errors/service-manager')
+const testDeveloperErrorMessages = require('../errors/test-developer')
+
+/**
+ * Validate check window and related check forms
+ * @param {Object} checkWindow
+ * @param {Array} allForms
+ * @returns {String}
+ */
+module.exports.validate = (checkWindow, allForms) => {
```
jon-shipley: This doesn't follow the front end validator pattern we have.  Validators should return a ValidatorError object that is either empty or contains errors.

Also - this appears to be named incorrectly - it is validating that there is a checkWindow (but not that it is in date) and the check Forms.  So we might call it something like CheckWindowAndCheckFormValidator.   But if we do that, we see that it might actually be better to create 2 validators - one for checkWindow and the other for CheckForms.

However, the current implementation of ValidationError is more geared towards showing errors from user-input in forms, which is what the other validators do, whereas what's needed here is to validate other entities in the DB - a set of predicates.

Is this even a validator then? It might be better to move this all to the service and call the service something like `checkWindowSanityCheckService.check()` that returns the error string.  @GuyHarwood - thoughts?    If we need to validate check windows here it seems likely we need to sanity check them elsewhere too.

admin/views/pupil-pin/generated-pins-list.ejs
GeorgeChatzigiannis: @jon-shipley added in config

PR #551, opened 2018-03-06
Bug/19723 revert sass root variable change

GuyHarwood:
Turns out we don't need the variable that allows us to prefix the root of linked assets.

any referenced assets are relative to the root of the CSS file anyway.


admin-assets/nginx/default.conf
```diff
@@ -75,4 +77,4 @@ server {
 
   # prevents file downloads from automatically opening in older versions of IE https://helmetjs.github.io/docs/ienoopen/
   add_header X-Download-Options noopen;
-}
+}
```
GeorgeChatzigiannis: EOF

admin/gulpfile.js
```diff
@@ -72,9 +72,9 @@ gulp.task('build', ['sass', 'bundleJs', 'copyImages', 'copyPDFs', 'copyCsvFiles'
 
 gulp.task('sass', function () {
   return gulp.src('./assets/**/*.scss')
-    .pipe(sassVariables({
+/*     .pipe(sassVariables({
```
GeorgeChatzigiannis: why not remove it completely for now? can always revert it through git

jon-shipley: yes - please remove if it's not in use.

GuyHarwood: @GeorgeChatzigiannis @jon-shipley yep, it should be gone.  commented temporarily to see if all worked ok without it.

PR #553, opened 2018-03-06
19752 Check window end date value correction

GeorgeChatzigiannis:
Set check window end date to the last minute of the particular day


jon-shipley:
@GuyHarwood - I thought the same - but this controller does not yet have a dedicated service, so the controller is bloated.   I would suggest it's a tech debt item.

GuyHarwood:
@jon-shipley good call, will add tech debt card

PR #556, opened 2018-03-07
Feature/18246 ps report to mark and process again

jon-shipley:
Psychometric report update

* adds pupil forename and surname into report
* Allows the script report to run the marking using a new flag '-m'


admin/bin/psychometricians-report-v2.js
```diff
@@ -17,15 +19,23 @@ async function main () {
...
-    await psychometricianReportService.batchProduceCacheData(checks.map(c => c.id))
+    // Make sure all completed checks are marked and ps-report data cached
+
+    if (requiresMarking) {
+      await completedCheckProcessingService.process()
+    } else {
+      // Just process everything in one batch.
+      // TODO: batchify it in a service
+      const checks = await psychometricianReportCacheDataService.sqlFindUnprocessedChecks()
+      if (!checks || !checks.length) return
```
GeorgeChatzigiannis: @jon-shipley could replace this with Array.isArray?

jon-shipley: isArray() would return true for an empty array though

PR #558, opened 2018-03-07
feature/19771 - Date format on PIN print card 

pris54:
BAs requested a change in the date format.


activemq:
screen shot 2018-03-07 at 15 40 02

PR #560, opened 2018-03-07
Feature/19773 change the guidance document

jon-shipley:
Update the administration guidance document with an updated PDF

* use direct links to the file to allow it to be served from a separate domain
* removes unecessary RBAC control
* remove old documents


activemq:
👍 tested all service manager, test dev and teacher pages, opens in a new tab instead of a download. Is this how we want it? On the feb trial branch it was a download

GuyHarwood:
@activemq which browsers did you try?  i'm curious to see how it behaves in IE, specifically

jon-shipley:
@GuyHarwood - everything is already in lower-case as far as I know.  Did I miss something?

In addition to the lower-case work you carried out this PR also renames `PDFs`  to `pdfs`

admin/helpers/index.js
```diff
@@ -49,4 +49,5 @@ module.exports = function (app) {
   app.locals.deployVersion = appVersion
   app.locals.formatGdsDate = formatGdsDate
   app.locals.formatFullGdsDate = formatFullGdsDate
+  app.locals.guidancePdf = `${config.AssetPath}pdfs/mtc-administration-guidance-2018-03.pdf`
```
GeorgeChatzigiannis: Is this going to be a standard approach for all the resources such as the guidance pdf?

GuyHarwood: @GeorgeChatzigiannis probably, yes.  it means we can hook in the asset domain easily at startup, when its used

GuyHarwood:
@jon-shipley think ive misread the commit, which mentions the upper case PDF file, but probably because its been removed.

admin/helpers/index.js
jon-shipley: @GeorgeChatzigiannis - this solution improves on the feb-trial download solution as it means we can offload the download of this large file to a better server: e.g. nginx or a CDN rather than express.

As we are likely to change the guidance frequently I chose to resolve it via a variable, so that in the future it only needs changing in one place.

We can't just copy a new file over the old file, as this would prevent us from using far-future cache headers.

pris54:
@jon-shipley Glad to see that we are back to the idea that a straight forward link to the public folder is the best solution.

GuyHarwood:
@activemq @sinhaamit1980 Travis is failing this on SPA tests

activemq:
@GuyHarwood its failing on the state tests which @sinhaamit1980 is working on #557

PR #565, opened 2018-03-09
Add a new message if all assigned forms have been assigned to a checkwindow

jon-shipley:


pris54:
@jon-shipley Just to note, I can't think of a way to reach this message unless the last form is assigned while someone else is *at the same time* opening the screen to assign forms. When all forms are assign the link to assign form becomes a text informing the user that there are no more forms to assign.

PR #569, opened 2018-03-12
17002 Detect previously unfinished check on login page

GeorgeChatzigiannis:
- Detect unfinished check upon user manual redirection to the login page
- Redirect the user to the check submission process and prevent potential login
- Display appropriate title when a pending submission is detected


pupil-spa/src/app/login/login.component.ts
```diff
@@ -22,10 +23,15 @@ export class LoginComponent implements OnInit {
...
     private warmupQuestionService: WarmupQuestionService,
-    private registerInputService: RegisterInputService
+    private registerInputService: RegisterInputService,
+    private checkRecoveryService: CheckRecoveryService
   ) { }
 
   ngOnInit() {
+    const hasUnfinishedCheck = this.checkRecoveryService.hasUnfinishedCheck();
+    if (hasUnfinishedCheck) {
+      this.router.navigate(['check']);
```
GuyHarwood: is this navigating back to the recovery page?

pupil-spa/src/app/services/check-recovery/check-recovery.service.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+        providers: [
+          StorageService,
+          CheckRecoveryService
+        ]
+      }
+    );
+    storageService = inject.get(StorageService);
+    checkRecoveryService = inject.get(CheckRecoveryService);
+  });
+  it('hasUnfinishedCheck should return true if there is no CheckComplete entry in the audit', () => {
```
GuyHarwood: would it not be cleaner to determine if the check has been submitted simply by finding an entry in localStorage, or do you feel this would be more deterministic?

pupil-spa/src/app/login/login.component.ts
GeorgeChatzigiannis: @GuyHarwood yes, it will navigate back to the submission pending screen and perform the end of check process.

pupil-spa/src/app/services/check-recovery/check-recovery.service.spec.ts
GeorgeChatzigiannis: @GuyHarwood I was discussing this idea of having an extra (boolean) property in the local storage to verify whether a check is submitted or not earlier today with @jon-shipley
I have the impression that this property will still rely on the audit log eventually so possibly wouldn't provide some added value apart from being a shortcut.

GuyHarwood: sounds like a duplicate way of determining state, which just makes the system less robust.

if the check has been submitted, surely the local storage payload will be gone?

GeorgeChatzigiannis: Currently the local storage is being wiped out upon successful new pupil login.
Meaning that the redirect will only happen if the audit array (which still is present at this stage) is populated and the type 'CheckComplete' is missing.
Aside from relying in the audit section of the local storage would there be a better approach that comes in mind?

GuyHarwood: i would have thought that this made the most sense.....
### Happy Path
- User completes check
- Check is submitted successfully
- Check data is wiped from local storage

### Unhappy Path
- User completes check
- Check is not submitted successfully 
- Check remains in memory
- Next user attempts to logon
- Previous check is discovered
- Previous check is uploaded

GeorgeChatzigiannis: @GuyHarwood updated as per our discussion

design/schemas/database/completedchecks.json
```diff
@@ -39,10 +39,6 @@
             {
                 "clientTimestamp" : "2017-09-14T23:52:30.684Z",
                 "type" : "PauseRendered"
-            },
```
GuyHarwood: 👍

PR #571, opened 2018-03-13
Introduce loading time limit and question time limit within groups

GeorgeChatzigiannis:
- Question and loading Timings will now have the following override sequence:
1. group timings
2. setting timings
3. default (config) timings


admin/data/sql/migrations/069.do.add-config-to-groups.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE mtc_admin.[group] ADD loadingTimeLimit tinyint NULL
```
pris54: @GeorgeChatzigiannis both fields need to be decimal(5,2)

admin/services/config.service.js
```diff
@@ -19,8 +20,12 @@ const configService = {
     let loadingTime = TIME_BETWEEN_QUESTIONS
 
     const timeSettings = await settingDataService.sqlFindOne()
-
-    if (timeSettings) {
+    const group = await groupDataService.sqlFindOneBySchoolId(pupil.school_id)
+    const hasGroupTimeLimits = group && group.loadingTimeLimit && group.questionTimeLimit
```
GuyHarwood: you cant look up the pupil group like this.  There can be many groups in a school, and the pupil may or may not be in one of them.  You should be looking up the group and its settings with a JOIN on the pupilGroup table

admin/data/sql/migrations/069.do.add-config-to-groups.sql
```diff
@@ -0,0 +1,2 @@
...
+ALTER TABLE mtc_admin.[group] ADD questionTimeLimit tinyint NULL
```
pris54: @GeorgeChatzigiannis same as above

admin/services/data-access/group.data.service.js
```diff
@@ -57,6 +57,28 @@ groupDataService.sqlFindOneById = async (groupId, schoolId) => {
   return R.head(result)
 }
 
+/**
+ * Get group by school id.
+ * @param schoolId
+ * @returns {Promise<void>}
+ */
+groupDataService.sqlFindOneBySchoolId = async (schoolId) => {
```
GuyHarwood: as above, this is not the right query

admin/data/sql/migrations/069.do.add-time-limits-to-group.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE mtc_admin.[group] ADD loadingTimeLimit tinyint NULL
```
pris54: @GeorgeChatzigiannis both fields need to be decimal(5,2)

admin/data/sql/migrations/069.do.add-time-limits-to-group.sql
```diff
@@ -0,0 +1,2 @@
...
+ALTER TABLE mtc_admin.[group] ADD questionTimeLimit tinyint NULL
```
pris54: @GeorgeChatzigiannis same as above.

admin/data/sql/migrations/069.do.add-time-limits-to-group.sql
GuyHarwood: good catch!

admin/services/data-access/group.data.service.js
pris54: @GeorgeChatzigiannis you may find more than one group with this query, so maybe `sqlFindOne...` is misleading

admin/services/data-access/group.data.service.js
```diff
@@ -57,6 +57,28 @@ groupDataService.sqlFindOneById = async (groupId, schoolId) => {
...
+  const sql = `SELECT * 
+    FROM ${sqlService.adminSchema}.[group]
+    WHERE school_id=@schoolId`
```
pris54: @GeorgeChatzigiannis you may want to add isDeleted=0 to make sure you're not fetching deleted groups.

PR #573, opened 2018-03-14
Extend access token expiration based on check window end date time

GeorgeChatzigiannis:
Provide epoch time representation of check window's end date time as jwt expiration.

### DIFF
--- a/admin/services/jwt.service.js
+++ b/admin/services/jwt.service.js
@@ -12,8 +12,8 @@ const pupilDataService = require('./data-access/pupil.data.service')
 const jwtService = {
   /**
    *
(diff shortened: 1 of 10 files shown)


admin/services/jwt.service.js
```diff
@@ -12,21 +13,24 @@ const jwtService = {
   /**
    *
    * @param pupil
+   * @param checkWindowEndDate
```
jon-shipley: What type of object is this?   Please document.

GeorgeChatzigiannis: @jon-shipley updated

PR #580, opened 2018-03-16
Update to Angular 5

GeorgeChatzigiannis:
### Angular 5 Update

- Requires update to global (local) angular/cli
- This update tackles the automatic karma rerun failure which occurred on the latest version of angular 4.


pris54:
@GeorgeChatzigiannis Are we 100% sure this update won't break anything else?

GeorgeChatzigiannis:
@pris54 it's under review but so far it doesn't seem broken to me.
This is not an urgent task as it stands so it can wait

PR #581, opened 2018-03-16
19874 - backfill check.startedAt date if its missing

pris54:


admin/spec/service/check-complete.service.spec.js
```diff
@@ -38,6 +40,7 @@ describe('check-complete.service', () => {
       spyOn(markingService, 'mark').and.returnValue(Promise.resolve())
       spyOn(jwtService, 'decode').and.returnValue({ sub: 1 })
       spyOn(completedCheckDataService, 'sqlFindOneByCheckCode')
+      spyOn(checkDataService, 'sqlFindOneByCheckCode').and.returnValue(Promise.resolve(checkMock))
```
GeorgeChatzigiannis: @jon-shipley what if the StartedAt date is not valid? don't we need a test for that as well

admin/services/check-complete.service.js
```diff
@@ -25,6 +28,18 @@ checkCompleteService.completeCheck = async function (completedCheck) {
   // TODO: This timestamp should be recorded in the application service tier instead
   completedCheck.receivedByServerAt = moment.utc()
 
+  const existingCheck = await checkDataService.sqlFindOneByCheckCode(completedCheck.data.pupil.checkCode)
+  if (!existingCheck.startedAt) {
+    winston.debug('Check submission for a check that does not have a startedAt date')
+    // determine the check started time from the audit log - CAUTION this is client data
+    const startedAt = moment(psUtilService.getClientTimestampFromAuditEvent('CheckStarted', completedCheck))
```
GeorgeChatzigiannis: very nice 👍

admin/spec/service/check-complete.service.spec.js
jon-shipley: Yes.  I didn't bother as the pupil status re-work may render this irrelevant

GeorgeChatzigiannis: Well as long as it is in future debt it's fine.

admin/services/check-complete.service.js
```diff
@@ -25,6 +28,18 @@ checkCompleteService.completeCheck = async function (completedCheck) {
   // TODO: This timestamp should be recorded in the application service tier instead
   completedCheck.receivedByServerAt = moment.utc()
 
+  const existingCheck = await checkDataService.sqlFindOneByCheckCode(completedCheck.data.pupil.checkCode)
+  if (!existingCheck.startedAt) {
```
GeorgeChatzigiannis: what if the existing check is null or undefined?
that will throw an exception

jon-shipley: It will throw `TypeError: Cannot read property 'startedAt'` - the controller will trap this and return a server error.

admin/services/check-complete.service.js
GuyHarwood: 👌

PR #585, opened 2018-03-20
Class disruption mandatory textarea on restart select screen

GeorgeChatzigiannis:


admin/assets/stylesheets/elements/_restarts.scss
```diff
@@ -59,11 +59,14 @@ input[type=radio]:checked ~ .hide {
   }
 }
 
-#didNotCompleteInfo {
-  padding-bottom: pem(5px);
+.restartReasonInfo {
```
pris54: @GeorgeChatzigiannis Can we make it into kebab case (lower case and hyphen separated)? That's how we are naming new css classes unless they are ids. So it would be `restart-reason-info`

admin/lib/validator/restart-validator.js
```diff
@@ -11,7 +11,10 @@ const restartValidator = {}
 restartValidator.validateReason = (reason, specifyReason) => {
   const validationError = new ValidationError()
   if (reason === 'DNC' && specifyReason.length === 0) {
-    validationError.addError('didNotCompleteInfo', restartErrorMessages.didNotCompleteInfo)
+    validationError.addError('didNotCompleteInfo', restartErrorMessages.errorMessage)
+  }
+  if (reason === 'CLD' && specifyReason.length === 0) {
```
pris54: @GeorgeChatzigiannis Would there be value in moving these arbitrary codes (CLD, DNC, etc) to a config file or somewhere more centralised?

admin/assets/stylesheets/elements/_restarts.scss
GeorgeChatzigiannis: @pris54 done

admin/lib/validator/restart-validator.js
jon-shipley: Removing magic strings is always a good idea!

GeorgeChatzigiannis: @pris54 @jon-shipley updated. please review once again

admin/config.js
```diff
@@ -27,7 +27,17 @@ module.exports = {
   Data: {
     allowedWords: process.env.ALLOWED_WORDS || 'aaa,bcd,dcd,tfg,bxx',
     pinSubmissionMaxAttempts: process.env.PIN_SUBMISSION_MAX_ATTEMPTS || 100,
-    helplineNumber: process.env.HELPLINE_NUMBER || '0345 278 8080'
+    helplineNumber: process.env.HELPLINE_NUMBER || '0345 278 8080',
+    validationRestartCodes: [
```
jon-shipley: I am not sure that this is config.  What is it?

GuyHarwood: yeah, seems like an *odd* place to have this.  why not in the DB?

GeorgeChatzigiannis:
@GuyHarwood @jon-shipley @pris54 the pupil restart codes are in the db and a const was generated to make this part more readable. Please review again.

admin/config.js
GeorgeChatzigiannis: @GuyHarwood  @jon-shipley it's updated

admin/lib/consts/pupil-restart-reasons.js
```diff
@@ -0,0 +1,4 @@
+module.exports = Object.freeze({
+  classroomDisruption: 'CLD',
+  didNotComplete: 'DNC'
+})
```
jon-shipley: 👌

PR #588, opened 2018-03-21
17599 - Pupil name and dob

pris54:
Rules to display DOB and middle names on all pupil lists.


admin/services/pupils-not-taking-check.service.js
```diff
@@ -1,30 +1,45 @@
...
+    } else if (sortDirection === 'asc') {
+      return a.reason < b.reason ? -1 : 1
+    } else {
+      return a.reason < b.reason ? 1 : -1
+    }
+  })
+  return sortedPupilsList
+}
+
+pupilsNotTakingCheckService.pupils = async (schoolId, sortField, sortDirection) => {
```
GuyHarwood: this method name give me no idea what it does

admin/services/pupils-not-taking-check.service.js
```diff
@@ -1,30 +1,45 @@
...
+  const pupils = await pupilDataService.sqlFindSortedPupilsWithAttendanceReasons(schoolId, sortField, sortDirection)
+  pupilIdentificationFlag.addIdentificationFlags(pupils)
+  return pupils
+}
+
+pupilsNotTakingCheckService.pupilsWithReasons = async (schoolId) => {
```
GuyHarwood: this method name give me no idea what it does

admin/services/pupils-not-taking-check.service.js
pris54: @GuyHarwood I struggled with naming these two methods (see your other comment), I'm relying on the service name to know what it is about. Would `getPupils` be good enough?

admin/services/pupils-not-taking-check.service.js
pris54: @GuyHarwood please see my previous reply. In this case would `getPupilsWithReasons` be enough?

admin/services/pupils-not-taking-check.service.js
GuyHarwood: build the name through its purpose, and prefix it with get.

what is it doing exactly, sorting?

admin/services/pupils-not-taking-check.service.js
GuyHarwood: it needs to be prefixed with get, if thats what its doing?

pris54:
@GuyHarwood Please let me know if the new method names make more sense.

PR #593, opened 2018-03-27
Feature/20144 Refactor add and edit check window functionality

GeorgeChatzigiannis:


admin/services/check-window.service.js
```diff
@@ -114,6 +93,88 @@ const checkWindowService = {
...
+    // fetch past current and future check windows
+    const checkWindows = await checkWindowDataService.sqlFindAllCheckWindows(sortField, sortDirection)
+    const pastCheckWindows = []
+    const newCheckWindows = []
+    checkWindows.map(cw => {
+      const adminStartDate = moment(cw.adminStartDate)
+      const checkStartDate = moment(cw.checkStartDate)
+      const checkEndDate = moment(cw.checkEndDate)
+
+      if (moment(checkStartDate).isAfter(moment(checkEndDate))) {
```
jon-shipley: Aren't `checkStartDate` and `checkEndDate` already moment objects?  Why clone them?

admin/services/check-window.service.js
```diff
@@ -114,6 +93,88 @@ const checkWindowService = {
...
+        id: cw.id,
+        name: cw.name,
+        adminStartDate: adminStartDate.format('D MMM YYYY'),
+        checkDates: dateService.formatCheckPeriod(checkStartDate, checkEndDate),
+        isPast
+      }
+
+      if (isPast) {
+        checkWindow.canRemove = false
+        pastCheckWindows.push(checkWindow)
```
jon-shipley: Adding `checkWindow` to different arrays and then concatenting the arrays doesn't actually seem to be required.  The code would be very slightly simpler without that step.  Is there a reason for it?

admin/services/check-window.service.js
```diff
@@ -114,6 +93,88 @@ const checkWindowService = {
...
+      adminStartYear: moment(checkWindow.adminStartDate).format('YYYY'),
+      checkStartDay: moment(checkWindow.checkStartDate).format('D'),
+      checkStartMonth: moment(checkWindow.checkStartDate).format('MM'),
+      checkStartYear: moment(checkWindow.checkStartDate).format('YYYY'),
+      checkEndDay: moment(checkWindow.checkEndDate).format('D'),
+      checkEndMonth: moment(checkWindow.checkEndDate).format('MM'),
+      checkEndYear: moment(checkWindow.checkEndDate).format('YYYY'),
+      existingAdminStartDate: adminStartDate,
+      existingCheckStartDate: checkStartDate,
+      adminIsDisabled: moment(currentDate).isAfter(adminStartDate) ? 1 : 0,
```
jon-shipley: Why not use `true` or `false` and convert them in the UI?

admin/services/data-access/check-window.data.service.js
```diff
@@ -237,6 +237,32 @@ const checkWindowDataService = {
     ]
     const result = await sqlService.query(sql, params)
     return R.head(result)
+  },
+
+  /**
+   * Find all check windows
+   * @param sortBy
+   * @param sortDirection
+   * @return {Object}
```
jon-shipley: This function returns an array of object?  So it should be `@returns [{Object}] - array of checkWindows`

admin/views/service-manager/check-windows-form.ejs
```diff
@@ -33,11 +33,11 @@
                     <span class="error-message"><%= error.checkWindowName %></span>
                     <% } %>
                 </label>
-                <input class="form-control <%= error.checkWindowName ? 'form-control-error' : '' %>"
+                <input class="form-control <%= typeof error !== 'undefined' && error.checkWindowName ? 'form-control-error' : '' %>"
```
jon-shipley: With all these checks can't the error object be initialised so it isn't `undefined`?  Would simplify imo.

admin/services/check-window.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/services/check-window.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/services/data-access/check-window.data.service.js
GeorgeChatzigiannis: @jon-shipley correct, it's updated

admin/views/service-manager/check-windows-form.ejs
GeorgeChatzigiannis: @jon-shipley updated

admin/services/check-window.service.js
```diff
@@ -142,6 +142,39 @@ const checkWindowService = {
...
+    }
+    const checkWindow = await checkWindowDataService.sqlFindOneById(id)
+    const currentDate = moment().format('YYYY-MM-D')
+    const adminStartDate = moment(checkWindow.adminStartDate, 'D MM YYYY').format('YYYY-MM-D')
+    const checkStartDate = moment(checkWindow.checkStartDate, 'D MM YYYY').format('YYYY-MM-D')
+
+    return {
+      checkWindowId: id,
+      checkWindowName: checkWindow.name,
+      adminStartDay: moment(checkWindow.adminStartDate).format('D'),
```
jon-shipley: Why are the dates re-parsed every time?  Wouldn't it make sense here to parse them once with moment and _then_  extract the date-parts with format?

admin/views/service-manager/check-windows-form.ejs
```diff
@@ -48,59 +48,59 @@
...
                     </div>
                     <div class="form-group form-group-year">
                         <label class="form-label" for="adminStartYear">Year</label>
-                        <input class="form-control <%= error.adminStartYear ? 'form-control-error' : '' %>"
+                        <input class="form-control <%= error.isError('adminStartYear') ? 'form-control-error' : '' %>"
                                id="adminStartYear"
                                name="adminStartYear"
                                type="number"
-                               value="<%= checkWindowData.adminStartYear %>"
+                               value="<%= typeof checkWindowData !== 'undefined' && checkWindowData.adminStartYear %>"
```
jon-shipley: Why do we need all these checks ` typeof checkWindowData !== 'undefined'`?

admin/services/check-window.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/views/service-manager/check-windows-form.ejs
GeorgeChatzigiannis: @jon-shipley updated

PR #599, opened 2018-03-29
Feature/17255 school controller refactoring part 3

pris54:
Third and last part of the school controller refactoring.


admin/spec/service/pupil-add-service.spec.js
```diff
@@ -89,7 +104,7 @@ describe('pupil-add-service', () => {
       expect(saveArg['dob-month']).toBeUndefined()
       expect(saveArg['dob-year']).toBeUndefined()
     } catch (error) {
-      expect('this not to be thrown').toBe(error)
+      expect('Error: Invalid req.body and/or school id. Saving pupil failed.').toBe(error.toString())
```
jon-shipley: Do we really expect tests on both sides of the try/catch?

PR #602, opened 2018-04-05
Bug/tech docker timezone fix

danwilliams:
Sets the admin docker container to run in the Europe/London Timezone rather than UTC

### DIFF
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -1,5 +1,7 @@
 FROM node:8.4.0
 
+# Set the timezone to BST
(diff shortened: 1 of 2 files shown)


jon-shipley:
AFAICT the underlying reason to do this is that it 'fixes' a single unit test that runs successfully only under this setting.

@danwilliams  has pointed out that it is usual for servers to run under UTC.  We may want to discuss this.

GuyHarwood:
The fix is fine, but does raise the question - why does the test rely on GMT?

all our dates are stored UTC

which test is failing?  is it a ruby one or JS?

GuyHarwood:
im going to close this, as i believe the test should be refactored to accommodate UTC, rather than the host be configured to use BST

PR #604, opened 2018-04-05
Added receivedByServerAt field to application service tier

danwilliams:
* Added sql migration files for receivedByServerAt column and index

* Changed the code to update receivedByServerAt in the sql database
instead of adding it to the data field


admin/services/check-complete.service.js
```diff
@@ -25,8 +25,10 @@ checkCompleteService.completeCheck = async function (completedCheck) {
     await pupilDataService.sqlUpdate({ id: pupil.id, pinExpiresAt: moment.utc(), pin: null })
   }
   // Timestamp the request
-  // TODO: This timestamp should be recorded in the application service tier instead
-  completedCheck.receivedByServerAt = moment.utc()
+  const receivedByServerAt = moment.utc()
+
+  // update receivedByServerAt request timestamp
+  await checkDataService.sqlUpdateCheckReceivedByServerAt(completedCheck.data.pupil.checkCode, receivedByServerAt)
```
GeorgeChatzigiannis: @danwilliams did these changes have any impact on the tests?
I would suspect there are going to be async timeouts unless you spy on this check data service method.

danwilliams: Locally, there were no effects... however I do see some timeouts on Circle. I will look into it.

admin/data/sql/migrations/073.do.add-receivedbyserverat-to-check.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE mtc_admin.[check] ADD receivedByServerAt DATETIMEOFFSET NULL;
```
jon-shipley: 👍

admin/services/check-complete.service.js
jon-shipley: Good work, however, I think the intention here is to timestamp the reception of the data, so wouldn't it make more sense to alter `completedCheckDataService.sqlAddResult()` which is called from `checkCompleteService.completeCheck()`

danwilliams: Agreed after discussion - changes made 😃

admin/spec/service/check-complete.service.spec.js
```diff
@@ -56,6 +56,19 @@ describe('check-complete.service', () => {
       done()
     })
 
+    it('passes a moment to receivedByServerAt', async (done) => {
+      // override initial spyOn to callFake for parameter checking
+      completedCheckDataServiceSpy.and.callThrough()
```
jon-shipley: Is this line needed?  It seems to be overridden in the next line.

PR #607, opened 2018-04-06
Feature/20322 Service manager refactor

GeorgeChatzigiannis:
Refactor occurred in the following controller methods: 
- Checkwindow save
- Checkwindow mark delete
- Check get timings
- Check set timings
- test coverage


admin/controllers/service-manager.js
```diff
@@ -226,92 +207,39 @@ const controller = {
...
-    let adminIsDisabled = req.body.adminIsDisabled
-    let checkStartIsDisabled = req.body.checkStartIsDisabled
-
-    if (req.body.checkWindowId !== '') {
-      actionName = 'Edit'
-      urlActionName = 'edit'
-      flashMessage = 'Changes have been saved'
-    }
+  saveCheckWindow: async (req, res, next) => {
+    let requestData = req.body
```
jon-shipley: a Let... hmmm

admin/controllers/service-manager.js
```diff
@@ -226,92 +207,39 @@ const controller = {
...
-        req.body.adminStartYear = moment(req.body[ 'existingAdminStartDate' ]).format('YYYY')
-      }
-
-      if (!req.body[ 'checkStartDay' ] && !req.body[ 'checkStartMonth' ] && !req.body[ 'checkStartYear' ] && req.body[ 'existingCheckStartDate' ] && req.body[ 'checkStartIsDisabled' ] === '1') {
-        req.body.checkStartDay = moment(req.body[ 'existingCheckStartDate' ]).format('D')
-        req.body.checkStartMonth = moment(req.body[ 'existingCheckStartDate' ]).format('MM')
-        req.body.checkStartYear = moment(req.body[ 'existingCheckStartDate' ]).format('YYYY')
-      }
-
+      requestData = checkWindowService.formatUnsavedData(requestData)
```
jon-shipley: This seems odd. 

Either a) this modifies the data which is passed by reference (not great) or
b) it returns a new data structure that is formatted (much better)

If b) then shouldn't it be assigned to a new variable?

admin/controllers/service-manager.js
```diff
@@ -323,31 +251,18 @@ const controller = {
    * @returns {Promise.<*>}
    */
   removeCheckWindow: async (req, res, next) => {
-    let checkWindow
-
     if (!req.params.checkWindowId) {
       return res.redirect('/service-manager/check-windows')
     }
-
+    let result
```
jon-shipley: Can this not be const?

admin/services/check-window.service.js
```diff
@@ -142,6 +142,84 @@ const checkWindowService = {
...
+  },
+
+  /**
+   * Save check window
+   * @param {Object} requestData
+   */
+  save: async(requestData) => {
+    let checkWindow
+    if (requestData.checkWindowId) {
+      checkWindow = await checkWindowDataService.sqlFindOneById(requestData.checkWindowId)
```
jon-shipley: Should this await be wrapped in a try/catch in case it is rejected?

admin/controllers/service-manager.js
GeorgeChatzigiannis: @jon-shipley haha fixed

admin/controllers/service-manager.js
GeorgeChatzigiannis: @jon-shipley just pushed another commit for this one

admin/controllers/service-manager.js
GeorgeChatzigiannis: I don't see anything wrong with declaring a reassignable variable right before the try/catch block.
It could change to `const` but it's an approach we are using across most of the controllers.

admin/services/check-window.service.js
GeorgeChatzigiannis: there's a try/catch in the controller level.
It should capture the rejection in case it occurs.

jon-shipley:
@GeorgeChatzigiannis - please resolve the conflicts

admin/controllers/service-manager.js
jon-shipley: Sorry, it's fine, the diff wasn't making that clear

GeorgeChatzigiannis:
@jon-shipley resolved

admin/controllers/service-manager.js
```diff
@@ -8,10 +8,8 @@ const settingsValidator = require('../lib/validator/settings-validator')
 const checkWindowValidator = require('../lib/validator/check-window-validator')
 const checkWindowErrorMessages = require('../lib/errors/check-window')
 const checkWindowService = require('../services/check-window.service')
-const dateService = require('../services/date.service')
 const checkWindowDataService = require('../services/data-access/check-window.data.service')
```
GuyHarwood: my editor tells me this is no longer used.

GeorgeChatzigiannis: @GuyHarwood yea spotted that as well. I accidentally approved it when I was resolving conflicts, will be deleted on my next commit.

admin/spec/controllers/service-manager.js
```diff
@@ -0,0 +1,270 @@
...
+
+  function getReq (params) {
+    const req = httpMocks.createRequest(params)
+    req.breadcrumbs = jasmine.createSpy('breadcrumbs')
+    req.flash = jasmine.createSpy('flash')
+    return req
+  }
+
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create()
```
jon-shipley: I'm not sure why the sandbox is here?  I suspect it can be removed and a straight Jasmine implementation will work.

GuyHarwood: good point.  pure jasmine definitely preferred going forward

GeorgeChatzigiannis: @GuyHarwood @jon-shipley it is not used at all, it was leftover code.
Removed

PR #608, opened 2018-04-09
Tech chore: make the anomaly detection script more robust

jon-shipley:


admin/bin/detect-anomalies.js
```diff
@@ -146,8 +147,8 @@ function detectMissingAudits (check) {
 }
 
 function detectInsufficientVerticalHeight (check) {
-  const height = check.data.device.screen.innerHeight
-  const width = check.data.device.screen.innerWith || check.data.device.screen.innerWidth
+  const height = R.path(['data', 'device', 'screen', 'innerHeight'], check)
+  const width = R.path(['data', 'device', 'screen', 'innerWidth'], check) || R.path(['data', 'device', 'screen', 'innerWith'], check)
```
GeorgeChatzigiannis: In case innerWidth is not defined you default to `innerWith`
Is that supposed to be correct?

jon-shipley: Unfortunately, yes.

There was a typo in the SPA recording during the Feb and March trials.   It's now corrected, but I'm leaving it in the script until we no longer need to review the Feb and March trial data.

PR #615, opened 2018-04-17
Initial creation of pupil-auth, extract dependencies from admin-app

danwilliams:
Note, this is nowhere near ready for review yet. Just creating the PR early for any first-stage feedback.


GuyHarwood:
@danwilliams you need to add the appropriate label 'DNM- WIP'

danwilliams:
Note: I've left the shared services atm as I wasn't sure; it's more like a 'project preference' - from what I know most people have them 'shared' and here it is the same (I think we have 'shared' in pupil-spa too) - but if you like it the other way, I can change it, there shouldn't be any problems.

GeorgeChatzigiannis:
There's a docker-compose conflict that needs resolving.

pupil-auth/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:8.4.0
```
jon-shipley: IMO we should be pulling from node:8 - the latest is fine and will include security updates.

pupil-auth/app.js
```diff
@@ -0,0 +1,169 @@
...
+const objectSources = ["'self'"]
+
+if (config.AssetPath !== '/') {
+  // add CSP policy for assets domain
+  scriptSources.push(config.AssetPath)
+  styleSources.push(config.AssetPath)
+  imgSources.push(config.AssetPath)
+  objectSources.push(config.AssetPath)
+}
+app.use(helmet.contentSecurityPolicy({
```
jon-shipley: I suspect that this security policy can be reset to nothing.   It's just an API end-point isn't it?

pupil-auth/app.js
```diff
@@ -0,0 +1,169 @@
...
+// catch 404 and forward to error handler
+app.use(function (req, res, next) {
+  let err = new Error('Not Found')
+  err.status = 404
+  next(err)
+})
+
+// error handler
+app.use(function (err, req, res, next) {
+  let errorId = uuidV4()
```
jon-shipley: Could this be  `const`?

pupil-auth/config.js
```diff
@@ -0,0 +1,86 @@
...
+module.exports = {
+  AZURE_STORAGE_CONNECTION_STRING: process.env.AZURE_STORAGE_CONNECTION_STRING,
+  GOOGLE_TRACKING_ID: process.env.GOOGLE_TRACKING_ID,
+  MONGO_CONNECTION_STRING: process.env.MONGO_CONNECTION_STRING || 'mongodb://localhost/mtc',
+  NCA_TOOLS_AUTH_URL: process.env.NCA_TOOLS_AUTH_URL,
+  PORT: process.env.PORT || '3002',
+  PUPIL_APP_URL: process.env.PUPIL_APP_URL,
+  QUESTION_TIME_LIMIT: 5,
+  RESTART_MAX_ATTEMPTS: 2,
+  SESSION_SECRET: process.env.NODE_ENV === 'production' ? process.env.SESSION_SECRET : 'anti tamper for dev',
```
jon-shipley: Are sessions actually needed here?  I would have thought this would be stateless?

pupil-auth/config.js
```diff
@@ -0,0 +1,86 @@
...
+  TIME_BETWEEN_QUESTIONS: 2,
```
jon-shipley: It doesn't show questions does it?

pupil-auth/config.js
```diff
@@ -0,0 +1,86 @@
...
+  LINES_PER_CHECK_FORM: 25,
```
jon-shipley: I don't this this will be needed.

pupil-auth/config.js
```diff
@@ -0,0 +1,86 @@
...
+  // autoMark true | false - Automatically mark the check data when we receive it: boolean
+  autoMark: process.env.hasOwnProperty('AUTO_MARK') ? toBool(process.env.AUTO_MARK) : true,
```
jon-shipley: The auth-app won't be doing any marking - this can go.

pupil-auth/config.js
```diff
@@ -0,0 +1,86 @@
...
+  Data: {
+    allowedWords: process.env.ALLOWED_WORDS || 'aaa,bcd,dcd,tfg,bxx',
```
jon-shipley: The auth-app won't be generating any pins.  This can go.

pupil-auth/config.js
```diff
@@ -0,0 +1,86 @@
...
+    pinSubmissionMaxAttempts: process.env.PIN_SUBMISSION_MAX_ATTEMPTS || 100,
```
jon-shipley: The auth-app won't be generating any pins. This can go.

pupil-auth/controllers/auth.js
```diff
@@ -26,7 +26,7 @@ const storeLogonEvent = async (pupilId, schoolPin, pupilPin, isAuthenticated, ht
   }
 }
 
-const getQuestions = async (req, res) => {
+const getAuth = async (req, res) => {
```
jon-shipley: Finally!  ✅

danwilliams:
Okay, so the non-shared approach, then! 😄

admin/app.js
```diff
@@ -90,7 +90,6 @@ const index = require('./routes/index')
 const testDeveloper = require('./routes/test-developer')
```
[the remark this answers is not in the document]
danwilliams: Question for this, I think you looked at the wrong file? admin/app.js is the one that has all routes except auth - what I changed there was just to remove the old questions route (that is now extracted in pupil-auth) - and pupil-auth has, as you said, only one route handler.

Can you confirm?

pupil-auth/config.js
GuyHarwood: yes, stateless. please remove session setup and config

pupil-auth/config.js
GuyHarwood: there are quite a lot of env vars that can be removed.  Dan may need some guidance on this

admin/app.js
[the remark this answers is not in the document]
GuyHarwood: the only route and handler in this app should be one for the auth endpoint.  Everything else can go.

let me know if unclear and we can discuss

danwilliams:
This is now ready for review again 😄

admin/app.js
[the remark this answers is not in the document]
GuyHarwood: just realised - this is the admin app routes, not the auth API 😄

docker-compose.admin-test.yml
```diff
@@ -15,18 +15,35 @@ services:
     links:
     - sqldb
     - mongo
+  pupil-auth:
+    build: ./pupil-auth
+    container_name: mtc_pupil_auth
+    ports:
+    - "3003:3003"
```
GuyHarwood: 👍

pupil-auth/config.js
```diff
@@ -0,0 +1,57 @@
+'use strict'
+require('dotenv').config()
```
GuyHarwood: These can all go....
```
MONGO_CONNECTION_STRING
QUESTION_TIME_LIMIT
TIME_BETWEEN_QUESTIONS
Azure: {
+      Scale: process.env.SQL_AZURE_SCALE

OverridePinExpiry: process.env.hasOwnProperty('OVERRIDE_PIN_EXPIRY') ? toBool(process.env.OVERRIDE_PIN_EXPIRY) : false,
```

pupil-auth/services/pin-generation.service.js
```diff
@@ -0,0 +1,70 @@
+const moment = require('moment')
```
GuyHarwood: i look at this and see an emerging requirement for another standalone service - one that both the admin app and other APIs can call.  Seems like the only way to keep dependencies on key parts of logic from being duplicated.
@jon-shipley thoughts?

pupil-spa/src/environments/environment.ts
```diff
@@ -5,6 +5,7 @@
 
 export const environment = {
   apiURL: 'http://localhost:3001',
+  authURL: 'http://localhost:3002',
```
GuyHarwood: this is now 3003 i believe?

pupil-spa/Dockerfile
```diff
@@ -17,10 +17,13 @@ COPY . .
 RUN yarn run postinstall
 RUN yarn run sass
 ARG api_url=http://localhost:3001
+ARG auth_url=http://localhost:3002
```
GuyHarwood: pretty sure this is now 3003?

pupil-auth/services/pin-generation.service.js
jon-shipley: @GuyHarwood 

Yes, I have thought something similar.   It certainly doesn't fit well in the auth service.

GuyHarwood: thinking about the wider final design, the auth service should just do a lookup for the credentials, nothing more.  Meaning that the test user creds should always be there for pickup (if instance has test accounts enabled).  This is a separate piece of work

pupil-auth/config.js
danwilliams: OverridePinExpiry var is used in the pin-generation service for the expiry time, when creating a session for local users.

Questions:
1. QUESTION_TIME_LIMIT and TIME_BETWEEN_QUESTIONS are used in service/config.service.js as default values, but it looks like they get overwritten anyway in the config service, is that the intended way?
2. Regarding azure, do we want azure insights enabled for this microservice?

GuyHarwood: ok, in that case leave them in.  These will eventually be removed though when we have solved the decoupling issue

GuyHarwood: 2.  yes, app insights will be required for this

pupil-spa/src/app/services/user/user.service.ts
```diff
@@ -24,7 +24,7 @@ export class UserService {
       headers.append('Content-Type', 'application/json');
       const requestArgs = new RequestOptions({headers: headers});
 
-      await this.http.post(`${environment.apiURL}/api/questions`,
```
GuyHarwood: @danwilliams please make this configurable for now, until such time that we decide we are ready for the API endpoints to be used.   we can then switchover in our own time

GuyHarwood:
@activemq @sinhaamit1980 there will be no behavioural changes when this is deployed, as the API endpoint will just sit inactive until such time that we switch the SPA over.  We can integrate changes to automation tests as part of that future switchover

pupil-auth/routes/index.js
```diff
@@ -0,0 +1,70 @@
...
+    commitId = await getCommitId()
+  } catch (error) {
+
+  }
+
+  res.setHeader('Content-Type', 'application/json')
+  let obj = {
+    'Build': buildNumber,
+    'Commit': commitId,
+    'CurrentServerTime': Date.now()
```
jon-shipley: I would very much prefer: `moment().toISOString()`

pupil-auth/services/pupil-authentication.service.js
```diff
@@ -0,0 +1,36 @@
...
+const pupilAuthenticationService = {
+  /**
+   * Authenticate a pupil from the school pin and pupil pin
+   * @param pupilPin
+   * @param schoolPin
+   * @return {object}
+   */
+  authenticate: async (pupilPin, schoolPin) => {
+    // TODO: Consolidate the following data layer calls to a single one
+    const school = await schoolDataService.sqlFindOneBySchoolPin(schoolPin)
```
jon-shipley: We could make these two sql calls happen in parallel, to save some time.

pupil-spa/src/environments/environment.ts
```diff
@@ -5,6 +5,8 @@
 
 export const environment = {
   apiURL: 'http://localhost:3001',
+  authURL: 'http://localhost:3001/api/questions',
+  // authURL: 'http://localhost:3003/auth',
```
jon-shipley: Can this commented-out line be deleted?

GuyHarwood: its to make it obvious what to set it to when we enable the API endpoints

PR #617, opened 2018-04-17
SQL Server Auditing - 1st Pass

GuyHarwood:
Stored procedure that creates a trigger on each table (except the audit table) so that audit entries are created on each UPDATE, INSERT and DELETE.

The incoming row (or row being deleted) is serialised into JSON and stored in a column row, along with the table name and operation being performed.

Next steps...

1. create audit table definition
2. record date+time audit entry created
3. allow searches by table and datetime


jon-shipley:
It's very nice that you can write this stored procedure to apply the triggers 😎 and also I can see the benefit of writing to a single audit table - you no longer need to maintain the audit tables and the actual tables.

My concern is - will writing to the audit table every become a bottleneck - as there is only 1 table?  

Also, how will we retrieve information about it - I am guess that the JSON will be freeform and unindexed, so how do ask: who modified the x Row in the Y table on Z date?

GuyHarwood:
@jon-shipley when i commit the definition of the audit table, you will see the following columns...

- JSON data
- Table Name
- Date Created

with these columns you will be able to narrow down to specific table activity on a particular day and time

GuyHarwood:
@jon-shipley @danwilliams @GeorgeChatzigiannis final scripts and changes are in after local testing.  Please submit your reviews

admin/data/sql/migrations/074.do.create-auditlog-table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE mtc_admin.auditLog
+(
+	id int identity primary key,
```
jon-shipley: Any idea if an `int` will be large enough to store all audit's or should we move to `bigint` now?

admin/data/sql/migrations/074.do.create-auditlog-table.sql
```diff
@@ -0,0 +1,8 @@
...
+	rowData nvarchar(max) not null,
+	tableName nvarchar(255) not null,
+	operation nvarchar(20) not null,
+	createdAt datetimeoffset default getutcdate() not null
+)
```
jon-shipley: How do we get user Id into this?

admin/data/sql/migrations/074.do.create-auditlog-table.sql
GuyHarwood: good point - definitely should be `bigint`

admin/data/sql/migrations/074.do.create-auditlog-table.sql
GuyHarwood: that needs to be sorted before it can be classed as done.   

The most obvious approach is to ensure that all tables have a `recordedByUserId` field, and on every `INSERT` and `UPDATE` this is submitted as part of the SQL statement.  That way its captured by the trigger.  However, this doesn't provide us with a solution for `DELETE`.  Originally everything was going to be soft deletes and the mtc app user would not have any `DELETE` permissions.  This has proven to be awkward for the pupil groups and a couple of other scenarios.

It would be nice if we could somehow pass the current users id into the context of the call and capture it in the trigger.  That needs some investigation.  Any ideas welcome

jon-shipley: We need to pass the userId from the app to the DB, as only the app knows the user.   The app can then setup a session variable to store the user ID, which can be referenced from stored procedures.  We might need a fall back user, if one wasn't defined (like Guest) and one for the data migrations.  

It would then need to clean up the variable at the end.

GuyHarwood: at the SQL level, the migrations run under a high privilege account, and all app users run under the same restricted db user account.  We could modify this so that each new 'app' user has a corresponding sql db user account created, which would make it very easy to capture the current user via `CURRENT_USER()`

I will research both approaches and see which turns out to be most viable

jon-shipley: Yes, that's the other way to do it 👍.  We would need to create a new db user whenever we saw a new user from the external auth server.   There is probably some code we add to for that.

GuyHarwood:
@jon-shipley updates made.  user id functionality will be in a separate PR

PR #619, opened 2018-04-19
Feature/20617 pupilspa webspeech api fixes

danwilliams:


pupil-spa/src/app/services/speech/speech.service.ts
```diff
@@ -163,15 +171,71 @@ export class SpeechService implements OnDestroy {
...
+   */
+  waitForEndOfSpeech(): Promise<any> {
+    const _window = this.windowRefService.nativeWindow;
+    return new Promise(resolve => {
+      if (!this.isSpeaking()) {
+        // if there is nothing in the queue, resolve() immediately
+        resolve();
+      } else {
+        // wait for the last speechEnded event to resolve()
+        let subscription, timeout;
```
GeorgeChatzigiannis: I am guessing this can be a const since subscription is not being redeclared anywhere

pupil-spa/src/app/services/speech/speech.service.ts
```diff
@@ -163,15 +171,71 @@ export class SpeechService implements OnDestroy {
...
+   * it uses an emulated .speaking property because utterances
+   * can get randomly cancelled when ending at the same time
+   * with another utterance
+   */
+  waitForEndOfSpeech(): Promise<any> {
+    const _window = this.windowRefService.nativeWindow;
+    return new Promise(resolve => {
+      if (!this.isSpeaking()) {
+        // if there is nothing in the queue, resolve() immediately
+        resolve();
```
jon-shipley: It might be easier to understand if the `if (!this.isSpeaking())` is turned into a guard condition with a return in the body to exit early if the function doesn't need to do anything.   Then the entire `else` condition can be removed, and the content promoted to one level out - the main function body.

PR #621, opened 2018-04-19
Bug/20320 question response time is recorded for entries without numeric values

jon-shipley:
Bugfix - Psychometrician report

We capture all key inputs including all alphanumeric key presses.  However, not all keypresses materially affect the answer.  When reporting on First Key Press and Last Key Press, and the derived Total Response Time, we now only include keypresses that materially affect the answer.  Backspace is tricky, as it may or may not affect the answer depending on whether there is something to delete.


admin/services/psychometrician-util.service.js
```diff
@@ -154,65 +154,97 @@ psUtilService.getUserInput = function getUserInput (inputs) {
...
   if (!filtered.length) {
     return ''
   }
 
-  return R.pathOr('error', [0, 'clientInputDate'], filtered)
+  return R.pathOr('error', ['clientInputDate'], R.head(filtered))
+}
+
+psUtilService.filterInputsToAnswerKeys = function (inputs) {
+  const normalisedInputs = psUtilService.cleanUpInputEvents(inputs)
```
GeorgeChatzigiannis: what if this returns an empty array?

jon-shipley: It will return an empty array.

jon-shipley: Please explain?

PR #624, opened 2018-04-23
feature/20612 SQL Session store for Express

GuyHarwood:


admin/data/sql/migrations/077.do.sql-session-store.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE [mtc_admin].[sessions](
+  [sid] varchar(255) NOT NULL
```
jon-shipley: If this is 255 chars and un-indexed won't retrievals be very very slow?

GuyHarwood: its the primary key

GuyHarwood: its the primary key

admin/data/sql/migrations/077.do.sql-session-store.sql
```diff
@@ -0,0 +1,6 @@
...
+    CONSTRAINT [pk_sessions] PRIMARY KEY CLUSTERED ([sid] ASC),
+  [expires] datetimeoffset NOT NULL,
```
jon-shipley: Do we want to index the `expires` field - then we can run a query based on this col.

docker-compose.admin-test.yml
```diff
@@ -7,14 +7,11 @@ services:
     - "3001:3001"
     environment:
       SQL_SERVER: "sqldb"
-      MONGO_CONNECTION_STRING: "mongodb://mongo/mtc"
```
jon-shipley: 😀

admin/data/sql/migrations/077.do.sql-session-store.sql
GuyHarwood: given the overhead of indexing millisecond precision dates i would rather not.  The situations where we would need to query this are minimal, and would likely be done offline.

if it were a common lookup we could look at that.

PR #625, opened 2018-04-24
Feature/18249 ncatools identifier

danwilliams:


admin/spec/service/nca-tools-user.service.spec.js
```diff
@@ -115,51 +115,4 @@ describe('nca-tools-user.service', () => {
       expect(userDataService.sqlUpdateSchool).not.toHaveBeenCalled()
     })
   })
-
-  describe('recordLogonAttempt', () => {
```
GuyHarwood: just so i understand the bigger picture here - is this recorded elsewhere, in adminLogonEvent for example?

admin/services/nca-tools-user.service.js
```diff
@@ -49,12 +49,6 @@ const service = {
     userRecord.mtcRole = roleService.mapNcaRoleToMtcRole(ncaUser.UserType, school)
     return userRecord
   },
-  recordLogonAttempt: async (logonData) => {
```
GeorgeChatzigiannis: is this supposed to be removed?

admin/services/nca-tools-user.service.js
```diff
@@ -29,7 +28,8 @@ const service = {
       const mtcRoleName = roleService.mapNcaRoleToMtcRole(ncaUser.UserType, school)
       const role = await roleService.findByTitle(mtcRoleName)
       const user = {
-        identifier: ncaUser.EmailAddress,
+        identifier: ncaUser.externalAuthenticationId,
+        displayName: ncaUser.EmailAddress,
```
jon-shipley: This doesn't seem quite right.   Shouldn't the display name be a name, e.g. `ncaUser.UserName`?

GuyHarwood: We don't actually care what the email address is, and it sometimes isnt an actual email address! (go figure).

We do however store the contents of that field in the top right of the admin header, so we are accomodating this with the 'displayName' field.

GuyHarwood: btw - in the payload i think its actually called `userName`

admin/spec/service/nca-tools-user.service.spec.js
danwilliams: The recordLogonAttempt was the only piece of code using ncaToolsSession, and there is no code that calls it. I believe it is recorded in admin-logon-event (as in `authentication/local-strategy.js` @98-115 and `authentication/nca-tools-authentication-strategy.js` @52 and @60).

admin/services/nca-tools-user.service.js
danwilliams: Please see my response to Guy's question about `admin/spec/service/nca-tools-user.service.spec.js` above.

admin/spec/service/nca-tools-user.service.spec.js
GuyHarwood: this is what i assumed to be the case.

admin/services/nca-tools-user.service.js
danwilliams: So, just to be sure, is the displayName okay, or do we want the UserName instead? (The spec was saying "populate users.displayName field with email address").

GuyHarwood: IIRC email address is always populated, username is not.  leave as is for now

jon-shipley: What kind of auth service doesn't give us the real name?

PR #627, opened 2018-04-25
20098 Check Started Microservice

danwilliams:
Initial service creation, extracted dependencies from admin-app


danwilliams:
Please note, some outstanding questions from 17921 will have an effect on this branch.

check-started/routes/index.js
```diff
@@ -0,0 +1,70 @@
...
+    commitId = await getCommitId()
+  } catch (error) {
+
+  }
+
+  res.setHeader('Content-Type', 'application/json')
+  let obj = {
+    'Build': buildNumber,
+    'Commit': commitId,
+    'CurrentServerTime': Date.now()
```
jon-shipley: Also - here - I would very much prefer this to be `moment().toISOString()` as I can't work out seconds since the epoch in my head, so dealing in milliseconds is 1000 times harder than something I find impossible 😉

GuyHarwood: @jon-shipley good point! 😄

check-started/config.js
```diff
@@ -0,0 +1,38 @@
...
+      LoggingEnabled: process.env.hasOwnProperty('SQL_POOL_LOG_ENABLED') ? toBool(process.env.SQL_POOL_LOG_ENABLED) : true
+    }
+  },
+  Logging: {
+    Express: {
+      UseWinston: process.env.EXPRESS_LOGGING_WINSTON || false
+    }
+  },
+  Environment: getEnvironment(),
+  JwtSecret: process.env.JWT_SECRET || '0bd8fbdab903fb50455cb4cf7ba9fb5beed3f209687ed2644aa731ee8231e14c'
```
GuyHarwood: i am assuming this value is consistent across all apps that will share the token?
this could be set to something as simple as 'dev-token' for clarity

check-started/controllers/check-started.js
```diff
@@ -0,0 +1,33 @@
+const jwtService = require('../services/jwt.service')
```
GuyHarwood: this is how i eventually want the auth endpoint to look - clean and simple 👌

GuyHarwood:
@jon-shipley rather than have a 'whole data layer' package, it is better to take a vertical slice of exactly what this app needs.  for example - a check started service and dedicated data services that perform just the actions that this endpoint needs.

GeorgeChatzigiannis:
@danwilliams you got some conflicts in this PR

jon-shipley:
@danwilliams - please resolve conflicts.

Is this still WIP?

PR #630, opened 2018-04-26
20756 CSV links bug

GeorgeChatzigiannis:
- csv directory and files are lowercased
- gulp task is pointing to the appropriate directory


admin/.gitignore
```diff
@@ -46,8 +46,8 @@ public/javascripts/app.js
 public/pdfs
 public/images
 public/images/*
-public/csvs
-public/csvs/*
+public/csv
```
jon-shipley: better 😃

PR #635, opened 2018-04-29
20518 Mark answers and generate question data based on check form

GeorgeChatzigiannis:
- Produce question data from check form
- Marking is driven by check form


admin/services/marking.service.js
```diff
@@ -43,23 +44,35 @@ markingService.mark = async function (completedCheck) {
     processedAt: moment.utc()
   }
 
+  const checkForm = await checkFormDataService.sqlFindOneById(completedCheck.checkForm_id)
```
jon-shipley: The checkForm should not be retrieved from inside the loop - please fetch all checkForms for the batch in one query before the loop, and pass the relevant checkForm in.

admin/services/marking.service.js
```diff
@@ -43,23 +44,35 @@ markingService.mark = async function (completedCheck) {
...
 
-  for (let answer of completedCheck.data.answers) {
+  for (let question of formData) {
+    const currentIndex = questionNumber - 1
+    const answer = completedCheck.data.answers[currentIndex]
+    const currentAnswer = answer && answer.answer
     const data = R.clone(answer)
-    data.answer = R.slice(0, 60, answer.answer)
+    // Immediately break when answer is not found based on question index
+    if (!data) break
```
jon-shipley: Why do you want to use `break` here? 

We still need to mark the question.  If we can't find the question then they get 0 marks, which we should store.

admin/services/marking.service.js
```diff
@@ -43,23 +44,35 @@ markingService.mark = async function (completedCheck) {
...
+    data.answer = currentAnswer ? R.slice(0, 60, currentAnswer) : 'n/a'
```
jon-shipley: 'n/a' is not the answer the user submitted, so we have business storing it in the database.  We should only store facts.  I suggest we change this to an empty string.

admin/services/marking.service.js
GuyHarwood: @GeorgeChatzigiannis this has come up a few times.  You really need to think carefully about the performance implications

admin/services/marking.service.js
GeorgeChatzigiannis: @jon-shipley changed

admin/services/marking.service.js
GeorgeChatzigiannis: @jon-shipley @GuyHarwood that was a naive mistake from my end, it won't happen again.
This has been updated.

admin/services/marking.service.js
GeorgeChatzigiannis: @jon-shipley Valid point, it is updated now

admin/services/marking.service.js
```diff
@@ -19,10 +20,12 @@ markingService.batchMark = async function (batchIds) {
   }
 
   const completedChecks = await completedCheckDataService.sqlFindByIds(batchIds)
-
+  const checkFormIds = completedChecks.map(c => c.checkForm_id)
+  const checkForms = await checkFormDataService.sqlFindByIds(checkFormIds)
   for (let cc of completedChecks) {
     try {
-      await this.mark(cc)
+      const checkForm = checkForms.find(cf => cc.checkForm_id === cf.id)
```
jon-shipley: Do we really want the overhead of a `find` here for every iteration of the loop?  It would be more efficient to use a hash map.

admin/services/marking.service.js
```diff
@@ -43,23 +46,32 @@ markingService.mark = async function (completedCheck) {
     processedAt: moment.utc()
   }
 
+  const formData = JSON.parse(checkForm.formData)
```
jon-shipley: We should probably handle the `SyntaxError` that can be thrown from `JSON.parse`.  I would prefer it if the JSON was already parsed before it was passed to this `mark` function - this is usually done by using `checkForm.service` instead of `checkForm.data.service`

On line 44 of marking-service: `maxMarks` should be derived from the `checkForm` and not the supplied data.

admin/services/marking.service.js
```diff
@@ -43,23 +46,32 @@ markingService.mark = async function (completedCheck) {
...
 
-  for (let answer of completedCheck.data.answers) {
-    const data = R.clone(answer)
-    data.answer = R.slice(0, 60, answer.answer)
-    data.questionNumber = questionNumber
+  for (let question of formData) {
+    const currentIndex = questionNumber - 1
+    const answer = completedCheck.data.answers[currentIndex]
+    const currentAnswer = answer && answer.answer
+    let data = R.clone(answer)
```
jon-shipley: It doesn't seem particularly clear to me that `data` has properties `factor1`, `factor2`, `answer`, `questionNumber`.

admin/services/check-complete.service.js
```diff
@@ -45,7 +46,8 @@ checkCompleteService.completeCheck = async function (completedCheck) {
   if (config.autoMark) {
     // HACK temporary way to mark checks until we move to a dedicated scheduled process
     const check = await completedCheckDataService.sqlFindOneByCheckCode(completedCheck.data.pupil.checkCode)
-    await markingService.mark(check)
+    const checkForm = await checkFormDataService.sqlFindOneById(completedCheck.checkForm_id)
```
jon-shipley: Potentially this `checkForm` may not be found.  We wouldn't want to issue an error because of a marking problem, so perhaps it's best if wrap these two lines in a try/catch and swallow any errors.  We should only mark the check if there is a `checkForm`

admin/services/marking.service.js
jon-shipley: Please move this lookup to use an object/properties approach.

admin/services/marking.service.js
jon-shipley: Hasn't been addressed.

GeorgeChatzigiannis: Just saw these comments will address them now

admin/services/check-form.service.js
```diff
@@ -332,6 +332,23 @@ const checkFormService = {
 
   removeWindowAssignment: async (formId, windowId) => {
     return checkFormDataService.sqlRemoveWindowAssignment(formId, windowId)
+  },
+
+  getParsedCheckForms: async (batchIds) => {
```
jon-shipley: I suggest we just call this `getCheckFormsByIds` or `getCheckForms`.  At any rate, I don't see the need to mention parsing it in the name - the return value from a service should almost certainly be something usable from this tier.

jon-shipley: I suggest just calling the arg: `ids` - it doesn't know anything about the caller's context, and doesn't need to know.

admin/services/check-form.service.js
```diff
@@ -332,6 +332,23 @@ const checkFormService = {
...
+    }
+    let checkForms
+    try {
+      checkForms = await checkFormDataService.sqlFindByIds(batchIds)
+      checkForms = checkForms.map(cf => {
+        cf.formData = JSON.parse(cf.formData)
+        return cf
+      })
+    } catch (error) {
+      throw new Error(error)
```
jon-shipley: I don't understand this.  You catch an error only to re-throw it?  Why?

admin/services/marking.service.js
```diff
@@ -57,10 +58,16 @@ markingService.batchMark = async function (batchIds) {
   }
 
   const completedChecks = await completedCheckDataService.sqlFindByIds(batchIds)
-
+  const checkFormIds = completedChecks.map(c => c.checkForm_id)
+  const checkForms = await checkFormService.getParsedCheckForms(checkFormIds)
+  const checkFormsHashMap = checkForms.reduce((obj, item) => {
```
jon-shipley: good 👍

admin/services/check-form.service.js
GeorgeChatzigiannis: @jon-shipley changed to getCheckFormsByIds as it is more descriptive to the method.

GeorgeChatzigiannis: @jon-shipley done

admin/services/check-form.service.js
GeorgeChatzigiannis: @jon-shipley was redundant and now removed as it doesnt really serve any purpose.

PR #637, opened 2018-05-02
Feature/20533 record angular errors in localstorage during check

jon-shipley:
Log exceptions that rise up to the global error handler in `localStorage` in the `audit` log.   These are then sent back to the server on completing the check.

Additionally, all exceptions are swallowed, and the app carries on in an unknown state as best it can.

The anomaly detection script will now check for AppErrors in the `audit` log.


GuyHarwood:
@GeorgeChatzigiannis this is last-resort handling of app crashes.  I propose we put that on the backlog

pupil-spa/src/app/error-handler.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { TestBed } from '@angular/core/testing';
```
GeorgeChatzigiannis: Many thanks! 👍

pupil-spa/src/app/error-handler.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    errorHandler = injector.get(GlobalErrorHandler);
+  });
+
+  it('should be created', () => {
+    expect(errorHandler).toBeTruthy();
+  });
+
+  it ('handleError calls he audit service to persist the error', () => {
```
GeorgeChatzigiannis: typo he-> the

admin/bin/detect-anomalies.js
```diff
@@ -187,7 +188,9 @@ function detectChecksThatTookLongerThanTheTheoreticalMax (check) {
   const config = check.data.config
 
   // Calculate the max total time allowed for the check
-  const maxCheckSeconds = (numberOfQuestions * config.loadingTime) + (numberOfQuestions * config.questionTime)
+  const maxCheckSeconds = (numberOfQuestions * config.loadingTime) +
+    (numberOfQuestions * config.questionTime) +
+    (config.speechSynthesis ? numberOfQuestions * 2.5 : 0)
```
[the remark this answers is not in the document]
jon-shipley: @GeorgeChatzigiannis It takes about 2.5 seconds to read out a question.  Edge on Windows takes even longer.  As it says in the comments, there is another ticket to increase the accuracy of this.  At the moment we can't easily determine the timerStart event from the audit.

PR #638, opened 2018-05-03
adding device information tests

activemq:
This PR is to add coverage around the device information stored in Local storage and then persisted to the DB


test/pupil/features/step_definitions/device_steps.rb
```diff
@@ -0,0 +1,17 @@
+Then(/^I should see device information populated in local storage$/) do
+  device_info = JSON.parse(page.evaluate_script('window.localStorage.getItem("device");'))
+  expect(device_info['battery'].keys).to eql ["isCharging", "levelPercent", "chargingTime", "dischargingTime"]
```
jon-shipley: Battery stuff is Chrome-specific, fyi

test/pupil/features/step_definitions/device_steps.rb
```diff
@@ -0,0 +1,17 @@
...
+  expect(device_info['cpu'].keys).to eql ["hardwareConcurrency"]
+  expect(device_info['navigator'].keys).to eql ["userAgent", "platform", "language", "cookieEnabled", "doNotTrack"]
+  expect(device_info['networkConnection'].keys).to eql ["downlink", "effectiveType", "rtt"]
```
jon-shipley: Likely to be Chrome-specific

test/pupil/features/step_definitions/device_steps.rb
```diff
@@ -0,0 +1,17 @@
+Then(/^I should see device information populated in local storage$/) do
+  device_info = JSON.parse(page.evaluate_script('window.localStorage.getItem("device");'))
+  expect(device_info['battery'].keys).to eql ["isCharging", "levelPercent", "chargingTime", "dischargingTime"]
+  expect(device_info['cpu'].keys).to eql ["hardwareConcurrency"]
```
jon-shipley: Is not supported in every browser.

test/pupil/features/step_definitions/device_steps.rb
activemq: We are currently only running in headless chrome, so should be consistent, when we begin running on different browsers it may become a issue.

activemq: @GuyHarwood yep can do ill add it now

activemq:
@GuyHarwood @jon-shipley Now added conditional assertions dependant on the driver being used. If the driver does not record screen information for example, the test will skip that assertion and print out a message stating the current driver does not support screen information. Which makes for a more robust test.

PR #641, opened 2018-05-03
Feature/20878 pupil api

GuyHarwood:
Consolidation of all endpoint apps into one codebase.

Each route can be enabled via config.  This allows us to maintain just one codebase yet deploy in flexible configurations.

For example...

dev env - would have just 1 instance, serving all endpoints. easy to setup and manage.
live env - app instance per endpoint, to utilise maximum scale.


pupil-api/sql/createDatabase.js
```diff
@@ -0,0 +1,89 @@
...
+
+const createDatabase = async (connection) => {
+  try {
+    let azureOnlyScaleSetting = ''
+    if (config.Sql.Azure.Scale) {
+      azureOnlyScaleSetting = `(SERVICE_OBJECTIVE = '${config.Sql.Azure.Scale}')`
+    }
+    winston.info(`attempting to create database ${config.Sql.Database} ${azureOnlyScaleSetting} if it does not already exist...`)
+    const createDbSql = `IF NOT EXISTS(SELECT * FROM sys.databases WHERE name='${config.Sql.Database}') 
+    BEGIN CREATE DATABASE [${config.Sql.Database}] ${azureOnlyScaleSetting}; SELECT 'Database Created'; END ELSE SELECT 'Database Already Exists'`
```
jon-shipley: I wouldn't mind if this was readable, rather than all on one line.

pupil-api/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+COPY package.json /usr/src/app
+RUN yarn install --ignore-engines
+
+COPY . .
+RUN yarn build
+RUN yarn global add pm2
+
+EXPOSE 3005
```
danwilliams: Port 3005 is exposed, but in the docker-compose files it’s referred through port 3003?

pupil-api/src/lib/validator/pin-validator.js
```diff
@@ -0,0 +1,16 @@
...
+ * @param {string} pin
+ * @param {Moment} pinExpiresAt
+ * @returns {Boolean}
+ */
+pinValidator.isActivePin = (pin, pinExpiresAt) => {
+  if (!pinExpiresAt || !pin) return false
+  return moment(pinExpiresAt).isAfter(moment.utc())
+}
+
+module.exports = pinValidator
```
danwilliams: Here, and in the other places where `module.exports` is used, I believe "the typescript way" is to do an `export = whatever_was_exported_before` instead of the `module.exports =` ?

pupil-api/src/run-jasmine.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
```
danwilliams: This whole file (`pupil-api/src/run-jasmine.js`) is a duplicate of `pupil-api/run-jasmine.js`, which is unnecessary since the first one will get executed.

pupil-api/src/controllers/ping.ts
```diff
@@ -0,0 +1,63 @@
...
+import * as fs from 'fs'
+import * as moment from 'moment'
+
+const getPing: any = async (req: Request, res: Response) => {
+  // get build number from /build.txt
+  // get git commit from /commit.txt
+  let buildNumber: object | string = 'NOT FOUND'
+  let commitId: object | string = 'NOT FOUND'
+  try {
+    buildNumber = await this.getBuildNumber()
```
danwilliams: I think lines 14 & 20 should be without `this.` since they're declared as const now in the outer scope?

pupil-api/src/controllers/ping.ts
```diff
@@ -0,0 +1,63 @@
...
+  } catch (error) {
+    // error
+  }
+
+  try {
+    commitId = await this.getCommitId()
```
danwilliams: See line 14 above.

docker-compose.web.yml
```diff
@@ -0,0 +1,44 @@
+version: '3'
+services:
+  admin-app:
+    build: ./admin
+    container_name: mtc_admin
```
danwilliams: With `container_name`, the services in this compose file will conflict with others in `docker-compose.admin-test.yml` and `docker-compose.pupil-test.yml` (that’s why the ones in pupil-test have _2). Same on lines 16 and 31 also.

docker-compose.web.yml
```diff
@@ -0,0 +1,44 @@
...
+    - "3001:3001"
+    environment:
+      SQL_SERVER: "sqldb"
+      MIGRATE: 1
+    depends_on:
+    - sqldb
+    links:
+    - sqldb
+  pupil-api:
+    container_name: mtc_pupil_api
```
danwilliams: See line 5 above.

docker-compose.web.yml
```diff
@@ -0,0 +1,44 @@
...
+      API_AUTH: "true"
+      API_CHECK_START: "true"
+      API_CHECK_SUBMIT: "true"
+    depends_on:
+    - sqldb
+    - admin-app
+    links:
+    - sqldb   
+  pupil-app:
+    container_name: mtc_pupil
```
danwilliams: See line 5 above.

pupil-api/Dockerfile
GuyHarwood: good catch, ive switched to 3003 so we now have...
- admin:300
- pupil-spa:3002
- pupil-api:3003

pupil-api/src/lib/validator/pin-validator.js
GuyHarwood: yes, however this is still a js file.  i was planning to allow a mix and we convert them gradually.

pupil-api/src/run-jasmine.js
GuyHarwood: good catch, thanks :)

pupil-api/src/controllers/ping.ts
GuyHarwood: Yes.  i think the whole file should be refactored into a class tbh.  Will tackle that this morning, if i have time.

docker-compose.web.yml
GuyHarwood: after trying to work with these files i'm of the opinion that we need to work out a better way to manage the separate test runs.  Lots of duplication etc.  I'm open to ideas.

GuyHarwood:
@danwilliams @jon-shipley @GeorgeChatzigiannis this is now ready for final review.  The current state is as follows...

- Skeleton controllers that return `501: Not Implemented` until we implement new functionality against pupil-db
- superfluous services/files removed
- Everything 'typed', no JS
- SQL migrator converted to TS

.travis.yml
```diff
@@ -48,7 +48,7 @@ install:
   - docker info
   - cd ${TRAVIS_BUILD_DIR}/pupil-spa && travis_retry yarn install --frozen-lockfile
   - cd ${TRAVIS_BUILD_DIR}/admin && travis_retry yarn install --frozen-lockfile
-  - cd ${TRAVIS_BUILD_DIR}/pupil-auth && travis_retry yarn install --frozen-lockfile
+# - cd ${TRAVIS_BUILD_DIR}/pupil-api && travis_retry yarn install --frozen-lockfile
```
jon-shipley: Shouldn't this be enabled and tests run on Travis?

GuyHarwood: there are currently no tests due to the fact that there is no logic implemented yet

PR #642, opened 2018-05-03
20860 Upload multiple pupils bug

GeorgeChatzigiannis:
- Return whole object back to controller with error when occured


jon-shipley:
@GeorgeChatzigiannis - can you provide some more details on what the issue was and why the fix works?  This is still marked a `WIP` so I may be jumping the gun...

GeorgeChatzigiannis:
@jon-shipley pupil controller was expecting to evaluate `uploadResult.error` and the pupil-upload service was returning only the error instead of the whole upload result.
For that reason the exception was never caught/handled and the application was left hanging until it timed out.

activemq:
@GuyHarwood  @jon-shipley  @GeorgeChatzigiannis  @danwilliams Tested locally, but this worked locally before, real test is when it is merged to master and tested on the DEV environment.

PR #643, opened 2018-05-04
17624 Create container before uploading file in the blob storage

GeorgeChatzigiannis:
- Call createContainerIfNotExists blob service method prior to uploading the file to prevent errors if `csvupload` does not exist


admin/services/data-access/azure-file.data.service.js
```diff
@@ -2,9 +2,17 @@ const azure = require('azure-storage')
...
 const blobService = config.AZURE_STORAGE_CONNECTION_STRING ? azure.createBlobService() : {
   createBlockBlobFromText: () => { return { name: 'test_error.csv' } },
-  getBlobToText: () => 'text' }
+  getBlobToText: () => 'text',
+  createContainerIfNotExists: () => {}
+}
 
 const azureUploadFile = async (container, remoteFilename, text, streamLength) => {
+  await new Promise((resolve, reject) => {
+    blobService.createContainerIfNotExists(container, null, (error) => {
```
GeorgeChatzigiannis: @GuyHarwood second argument is optional for options:

`metadata`: The metadata key/value pairs.
`publicAccessLevel`: Specifies whether data in the container may be accessed publicly and the level of access.
`timeoutIntervalInMs`: The timeout interval, in milliseconds, to use for the request.

Doubt any of these are needed

PR #645, opened 2018-05-08
Feature/20788 pupil restart after attendance removed

danwilliams:
Changed the pin expiry for multiple pins to set the expiration timestamp to the current one (as in the singular `expirePupilPin` one), not sure that affects anything else.


admin/services/pin.service.js
```diff
@@ -62,9 +62,10 @@ pinService.expireMultiplePins = async (pupilIds) => {
     if (p.pin || p.pinExpiresAt) pupilData.push(p)
   })
   if (pupilData.length === 0) return
+  const currentTimeStamp = moment.utc()
   pupilData = pupilData.map(p => {
     p.pin = null
-    p.pinExpiresAt = null
+    p.pinExpiresAt = currentTimeStamp
```
jon-shipley: If there is no pin, why would the expiry be set to a particular date?  Is this set as a security precaution to prevent someone logging in with a null pin?

danwilliams: That was because of two reasons: that’s how it’s done in the `expirePupilPin` method too, and because the restart service checks that the pin has expired today (`admin/services/restart.service.js` @ 52), so if the timestamp is set to null (as it was before in the expireMultiplePins method), checking if it expired today will always return false, so the pupil won’t appear in the restart list.

jon-shipley: 👍

PR #646, opened 2018-05-08
Feature/20789 softdelete pupil reason check

danwilliams:
There was no `isDeleted` column in the database for soft-deleting, so I also created a migration id 81 that adds it.


admin/services/data-access/pupil.data.service.js
```diff
@@ -299,7 +299,7 @@ pupilDataService.sqlFindSortedPupilsWithAttendanceReasons = async (dfeNumber, so
   SELECT p.*, pg.group_id, ac.reason
   FROM ${sqlService.adminSchema}.${table} p 
     INNER JOIN ${sqlService.adminSchema}.[school] s ON p.school_id = s.id
-    LEFT OUTER JOIN ${sqlService.adminSchema}.[pupilAttendance] pa ON p.id = pa.pupil_id 
+    LEFT OUTER JOIN ${sqlService.adminSchema}.[pupilAttendance] pa ON p.id = pa.pupil_id AND (pa.isDeleted IS NULL OR pa.isDeleted = 0)
```
GeorgeChatzigiannis: If the column is not null shouldn’t it only include the ones equal to 0?

GuyHarwood:
@GeorgeChatzigiannis does this soft delete flag affect any of the pupil status calculations?

GeorgeChatzigiannis:
@GuyHarwood that's a very good point.  
@danwilliams The pupil status service will definitely require a change to accommodate the pupil attendance soft deletion since its looking at records only based on pupil id.

GuyHarwood:
It failed CI, so i hope thats because the automated tests around the raised concern were the ones that failed!

danwilliams:
@GeorgeChatzigiannis it doesn’t affect the pupil status because that service is using the methods in pupil-attendance.data.service to check for id, and I’ve changed them (`findByPupilIds` and `findOneByPupilId`) to check for isDeleted=0 too so that should be handled. Are you happy with this? Or have I missed anything? 😄

GeorgeChatzigiannis:
@danwilliams you are correct, pupil status service should work as intended with these changes.

What about the change I am proposing for pupil data service?

Under which case should we expect the value to be NULL?

danwilliams:
@GeorgeChatzigiannis I think it's supposed to accept nulls too, that’s why it’s a `LEFT OUTER JOIN`, because we need to get all users (that have no reasons, or that already had a reason assigned) in order to display them in the list where reasons can be assigned to users.

PR #647, opened 2018-05-08
Feature/20488 flatten the inputs register to a single array of objects with the question number

jon-shipley:
Flatten the input array storage in the SPA client.

Was - array of arrays
Now - single array with new input elements appended to the array


pupil-spa/src/app/services/register-input/registerInput.service.spec.ts
```diff
@@ -68,7 +68,7 @@ describe('RegisterInputService', () => {
       // registeredInputService line 55, where the `idx` is set to 100 if the
```
danwilliams: I believe these comments @ 67-70 are now not relevant anymore?

jon-shipley: Good spot 👍 
Now removed

PR #648, opened 2018-05-08
20861 Allow case insensitive gender validation for pupils

GeorgeChatzigiannis:
- This affects both single and multiple pupil validator as the validation process is shared


admin/spec/lib/validator/pupil-validator.spec.js
```diff
@@ -424,6 +424,13 @@ describe('pupil validator', function () {
         expect(validationError.get('gender')).toBe('Gender must be M or F')
         done()
       })
+      it('can be accepted in lowercase', async (done) => {
```
GuyHarwood: do we have a test for upperCase too?

GeorgeChatzigiannis: the existing successful scenario includes a gender in uppercase.(getBody method)

PR #651, opened 2018-05-09
Feature/20513 App usage counter

GeorgeChatzigiannis:
- Introducing app count service to monitor the amount of times the app has been used without a refresh or close tab/browsers while checks are performed
- Incremented upon successful login


pupil-spa/src/app/check/check.component.ts
```diff
@@ -45,7 +46,8 @@ export class CheckComponent implements OnInit {
               private warmupQuestionService: WarmupQuestionService,
               private auditService: AuditService,
               private storageService: StorageService,
-              protected windowRefService: WindowRefService) {
+              protected windowRefService: WindowRefService,
+              protected appCountService: AppCountService) {
```
danwilliams: I believe this can be private?

pupil-spa/src/app/check/check.component.ts
```diff
@@ -332,6 +334,7 @@ export class CheckComponent implements OnInit {
     const stateDesc = this.getStateDescription();
     console.log(`Refresh detected during state ${this.state} ${stateDesc}`);
     this.auditService.addEntry(new RefreshDetected());
+    this.appCountService.reset();
```
danwilliams: If a physical refresh of the page happens, because `appUsageCounter` is not stored in the localStorage or anywhere else, it will get reset to 0, so this might be unnecessary - but it’s good to double-check.

pupil-spa/src/app/app.module.ts
```diff
@@ -40,6 +40,7 @@ import { WarmupIntroComponent } from './warmup-intro/warmup-intro.component';
 import { WarmupLoadingComponent } from './warmup-loading/warmup-loading.component';
 import { WarmupQuestionService } from './services/question/warmup-question.service';
 import { WindowRefService } from './services/window-ref/window-ref.service';
+import { AppCountService } from './services/app-count/app-count.service';
```
jon-shipley: It's counting Checks, not Apps.   I would go with `CheckCountService` or `AppUsageService`.   `AppUsageService` works for me because we may extend this to store other wonderful variables like the time and date the app was initialised, the number of times the network dropped out, or whatever.

pupil-spa/src/app/check/check.component.ts
jon-shipley: As it is stored in memory it will be automatically re-initialised to 0, as @dan points out.  Are there any cases where this would actually do something?

pupil-spa/src/app/services/app-count/app-count.service.ts
```diff
@@ -0,0 +1,23 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class AppCountService {
```
jon-shipley: 👍

pupil-spa/src/app/check/check.component.ts
GeorgeChatzigiannis: @danwilliams it should have been yes but the service is not needed in the check component so it is now removed.

pupil-spa/src/app/app.module.ts
GeorgeChatzigiannis: @jon-shipley renamed

pupil-spa/src/app/services/app-count/app-count.service.ts
GeorgeChatzigiannis: @jon-shipley I thought it as well afterwards that reset was completely pointless.
Removed

PR #652, opened 2018-05-09
Add script to dump browser stats for all completed checks

jon-shipley:
Ad-hoc script from a spike to output stored browser data


admin/bin/dump-browser-stats.js
```diff
@@ -0,0 +1,99 @@
...
+
+  if (checkStart && checkStart !== 'error') {
+    const checkStartDate = moment(checkStart)
+    checkDate = checkStartDate.isValid() ? dateService.formatUKDate(checkStartDate) : ''
+  }
+  return checkDate
+}
+
+function getBrowserStats (check) {
+  const agent = useragent.lookup(R.path(['data', 'device', 'navigator', 'userAgent'], check))
```
GeorgeChatzigiannis: in case `userAgent` is not found within the path, is that working without issues?

jon-shipley: `useragent` has sensible defaults.

```
> node -i
> const userAgent = require('useragent')
undefined
> userAgent.lookup(undefined)
Agent { family: 'Other', major: '0', minor: '0', patch: '0', source: '' }
```

PR #656, opened 2018-05-10
21015 - Added functionality to hide question while it's read out

danwilliams:


jon-shipley:
Whilst the code is fine (apart from the missing tests that @GuyHarwood mentioned) I am not sure of the functionality:

* the screen re-flows the layout when the question is shown
* it is still possible to enter input before the question is read out / shown
* entering an answer seems to show the question before the question has finished reading

PR #659, opened 2018-05-11
Bug/20862 fix error message

jon-shipley:
Multiple pupil upload feature: bugfix: remove the UPN duplication message from the summary error screen

And also
* updated the error message when uploading a file containing more than 300 rows
* updated row counting and processing to ignore blank lines 
* updated quite a few lint errors (whitespace changes)
* updated README
* Safari now downloads the pupil error file, rather than showing it in the browser
* add an ArrayUtils module - every project has one!


admin/lib/array-utils.js
```diff
@@ -0,0 +1,28 @@
+const R = require('ramda')
+
+module.exports = {
```
GeorgeChatzigiannis: 👍

admin/lib/validator/csv-validator.js
```diff
@@ -13,15 +15,32 @@ module.exports.validate = (dataSet, header, element) => {
...
   return validationError
 }
+
+/**
+ * Count non-empty lines in the data
+ * NB Excel may add several empty lines to CSVs.  This count will ignore them.
+ * @param {Array} dataSet Array of Arrays
+ * @return {number}
+ */
+function countNonEmptyRows (dataSet) {
```
GeorgeChatzigiannis: perhaps that could be moved to utils as well?

GeorgeChatzigiannis: Also is this method tested?

PR #660, opened 2018-05-14
Added GA error event in the global error handler

danwilliams:


pupil-spa/src/app/error-handler.ts
```diff
@@ -39,5 +39,12 @@ export class GlobalErrorHandler implements ErrorHandler {
       hitType: 'pageview',
       page: '/error'
     });
+
+    window.ga('send', {
+      hitType: 'event',
+      eventCategory: 'error',
+      eventAction: url,
+      eventLabel: errorMessage
```
jon-shipley: I think we need to consider what might be in the content of the error message before sending it to third parties.  We would need to be sure we are not sending any personal details at all.  All in all - I'm not sure that is guaranteed at this time.  @GuyHarwood - thoughts?

GuyHarwood: @danwilliams Happy for this to go through, as any personally identifiable information should be wiped from memory before the practice questions commence (separate piece of work)

PR #662, opened 2018-05-15
Feature/20540 prevent duplicate answers

jon-shipley:
Prevent the possibility of accepting duplicate answers to the same question.  Instead, and new audit is recorded with the details of the duplicate.


pupil-spa/src/app/services/answer/answer.service.ts
```diff
@@ -15,8 +18,24 @@ export class AnswerService {
...
+      x => {
+        if (x.sequenceNumber === answer.sequenceNumber
+        && x.factor1 === answer.factor1
+        && x.factor2 === answer.factor2) {
+          return true;
+        }
+        return false;
+      });
+
+    if (isAlreadyAnswered.length) {
```
GuyHarwood: will this not always be initialised to zero or more?

jon-shipley: It's the result of a `filter` so will be an array.  If it finds any there will an element in the array.  Perhaps the name suggests it is a boolean when it isn't.  What do you suggest?

pupil-spa/src/app/services/answer/answer.service.ts
```diff
@@ -15,8 +18,24 @@ export class AnswerService {
       answers = [];
     }
 
+    // Check for duplicates - we can only store one answer for each question
+    const isAlreadyAnswered = answers.filter(
+      x => {
+        if (x.sequenceNumber === answer.sequenceNumber
```
GeorgeChatzigiannis: why not `x => x.sequenceNumber === answer.sequenceNumber && x.factor1 === answer.factor1 && x.factor2 === answer.factor2` ?

pupil-spa/src/app/services/answer/answer.service.ts
danwilliams: I agree that this sounds like a boolean. `alreadyAnswered` would work nicely.

jon-shipley:
@dan - Thanks for the comment.

Changes have already been added in commit https://github.com/DFEAGILEDEVOPS/MTC/pull/662/commits/8f5721cbcd5d0d9ca7f93839b1e4a1ea6e627544

```
+    if (previousAnswers.length > 0) {
 +      // Log this error
 +      this.auditService.addEntry(new DuplicateAnswerError(answer));
 +      return;
 +    }
```

pupil-spa/src/app/services/answer/answer.service.ts
```diff
@@ -15,8 +18,33 @@ export class AnswerService {
...
 
+  /**
+   * Find a previous answer in the answers array
+   * @param {Answer} answer - the current answer to be stored
+   * @param [Answer] answers - the list of previously stored answers
+   * @return [Answer?] - empty array or array of answers
+   */
+  private getPreviousAnswers(answer: Answer, answers) {
+    return answers.filter(
+      x => {
```
GeorgeChatzigiannis: Isn't it more readable like this? `x => x.sequenceNumber === answer.sequenceNumber && x.factor1 === answer.factor1 && x.factor2 === answer.factor2`
You are basically using an if to return boolean

jon-shipley: @GeorgeChatzigiannis - I thought it was less readable that way, sorry.

GeorgeChatzigiannis: I guess readability can be subjective

jon-shipley: @GeorgeChatzigiannis - happy to oblige if there is sufficient support for the change

GeorgeChatzigiannis: @jon-shipley it's a minor detail really if everyone else approves it as it stands I am happy.

PR #664, opened 2018-05-17
21204/rename cookie on admin app

GuyHarwood:
rename cookie to something more relevant


GuyHarwood:
@sinhaamit1980 @activemq will renaming the cookie break any of your tests?

PR #672, opened 2018-05-22
Bug/19402 fix test developer can remove form after check start date has passed

jon-shipley:
Add business logic 

* when showing the "Remove from Window" link in the UI
* when validating the request the request to unassign a checkForm from a checkWindow


GuyHarwood:
just one observation - do we need to be more descriptive about why they couldn't remove it?  as they could perceive it as being an error, rather than a valid business rule?

jon-shipley:
@GuyHarwood - the error message hasn't actually changed, and is a catch-all for all types of error conditions including DB error/network error.

PR #677, opened 2018-05-30
Bug/21337 Change tracking code snippet and allow google tag manager url on CSP

GeorgeChatzigiannis:
There might be a need to empty cache/hard reload to make the script appear


admin/app.js
```diff
@@ -133,9 +133,9 @@ if (config.Logging.Express.UseWinston === 'true') {
 
 app.use(cors())
 app.use(helmet())
-const scriptSources = ["'self'", "'unsafe-inline'", 'https://www.google-analytics.com']
+const scriptSources = ["'self'", "'unsafe-inline'", 'https://www.google-analytics.com', 'https://www.googletagmanager.com']
 const styleSources = ["'self'", "'unsafe-inline'"]
-const imgSources = ["'self'", 'https://www.google-analytics.com', 'data:']
+const imgSources = ["'self'", 'https://www.google-analytics.com', 'https://www.googletagmanager.com', 'data:']
```
GeorgeChatzigiannis: @GuyHarwood doubt we need it here.
Why was even analytics in the imgSources in the first place? 🤔

GuyHarwood: because it drops pixels https://developers.google.com/analytics/resources/concepts/gaConceptsTrackingOverview

admin/views/partials/_body_start.ejs
```diff
@@ -1,9 +1,9 @@
+<!-- Global Site Tag (gtag.js) - Google Analytics -->
+<script async src='https://www.googletagmanager.com/gtag/js?id=<%= googleTrackingId %>'></script>
 <script>
-    var googleTrackingId = '<%= googleTrackingId %>';
-    if (googleTrackingId) {
-      window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
-      ga('create', googleTrackingId, 'auto');
-      ga('send', 'pageview');
-    }
+  window.dataLayer = window.dataLayer || [];
```
GuyHarwood: you need the IF block to deal with situations where we have not added a tracking code and GA should not be enabled

GeorgeChatzigiannis: @GuyHarwood very true, it is updated

activemq:
screen shot 2018-05-31 at 13 53 37
screen shot 2018-05-31 at 13 52 56
LGTM 👍

PR #680, opened 2018-05-31
21238 Process data and bulk import to pupil table

GeorgeChatzigiannis:


admin/data/sql/migrations/090.do.add-bulk-permissions-pupil-table.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
```
GuyHarwood: why is there no allocation in the g sheet?

admin/services/data-access/pupil.data.service.js
```diff
@@ -2,9 +2,13 @@
...
 const { TYPES } = require('tedious')
 const R = require('ramda')
-const pupilDataService = {}
+const moment = require('moment')
+
 const table = '[pupil]'
+const config = require('../../config')
+const pupilDataService = {}
 const sqlService = require('./sql.service')
+const sqlPoolService = require('./sql.pool.service')
```
GuyHarwood: would this not be better in a dedicated service? its already 370~ lines long. 

how about pupilCensusImportService?

admin/data/sql/migrations/090.do.add-bulk-permissions-pupil-table.js
GeorgeChatzigiannis: Was about to allocate it. it's done.

admin/services/data-access/pupil.data.service.js
GeorgeChatzigiannis: @GuyHarwood I created pupilCensusImportDataService which contains now this functionality.
Let me know if you agree since it's data related.

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -0,0 +1,72 @@
+'use strict'
+
+const { TYPES } = require('tedious')
+const moment = require('moment')
+
+const table = '[pupil]'
```
GuyHarwood: can we please just include the table in the string.  it doesnt need to be configurable

GeorgeChatzigiannis: @GuyHarwood done

admin/services/pupil-census.service.js
```diff
@@ -39,7 +40,10 @@ pupilCensusService.upload = async (uploadFile) => {
       })
   })
   const blobResult = await pupilCensusService.uploadToBlobStorage(csvData)
```
GeorgeChatzigiannis: I am tempted to remove that as the blobstorage for the current implementation feels very unneeded.
I'll wait for @GuyHarwood comment on that one.

admin/data/sql/migrations/090.do.add-bulk-permissions-pupil-table.js
```diff
@@ -0,0 +1,7 @@
...
+
+const config = require('../../../config')
+
+module.exports.generateSql = function () {
+  return ` GRANT ALTER ON [mtc_admin].[pupil] TO ${config.Sql.Application.Username}`
```
jon-shipley: Why does the App user need `ALTER TABLE` perms?  Are we dynamically modifying the table during the application run?

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+  con.release()
+  return result
+}
+
+const bulkLoadData = (connection, pupilData, schools) => {
+  return new Promise((resolve, reject) => {
+    const bulkLoad = connection.newBulkLoad(`${config.Sql.Database}.${sqlService.adminSchema}.[pupil]`, function (error, rowCount) {
+      if (error) {
+        reject(error)
```
jon-shipley: If we reject here, we should also return... to prevent the subsequent resolve()

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -0,0 +1,69 @@
...
+    bulkLoad.addColumn('lastName', TYPES.NVarChar, {length: 'max', nullable: false})
+    bulkLoad.addColumn('foreName', TYPES.NVarChar, {length: 'max', nullable: false})
+    bulkLoad.addColumn('middleNames', TYPES.NVarChar, {length: 'max', nullable: true})
+    bulkLoad.addColumn('gender', TYPES.Char, {length: 1, nullable: false})
+    bulkLoad.addColumn('dateOfBirth', TYPES.DateTimeOffset, {nullable: false})
+
+    for (let index = 0; index < pupilData.length; index++) {
+      const csvRow = pupilData[index]
+      const dfeNumber = `${csvRow[0]}${csvRow[1]}`
+      const school = schools.find(s => s.dfeNumber === parseInt(dfeNumber))
```
jon-shipley: Should be a hash lookup, not a find.

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -0,0 +1,69 @@
...
+      const schoolId = school && school.id
+      if (!schoolId) {
+        reject(new Error(`School id not found for DfeNumber ${dfeNumber} for pupil on row number ${index + 1}`))
```
jon-shipley: Do you think we should return here?

admin/services/pupil-census-processing.service.js
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Processes and bulk imports pupil census
+ * @param csvData
+ * @return {Promise<void>}
+ */
+pupilCensusProcessingService.process = async(csvData) => {
+  // Fetch all school for pupil records
+  let schoolDfeNumbers = csvData.map(r => `${r[0]}${r[1]}`)
+  // filter duplicate entries
+  schoolDfeNumbers = schoolDfeNumbers.filter((item, pos, self) => self.indexOf(item) === pos)
```
jon-shipley: I dislike overwriting the variable you are filtering on.  It would be much cleaner to use a new array.

admin/services/pupil-census.service.js
```diff
@@ -57,20 +61,24 @@ pupilCensusService.uploadToBlobStorage = async (uploadFile) => {
...
  * @return {Object}
  */
-pupilCensusService.create = async (uploadFile, blobResult) => {
+pupilCensusService.create = async (uploadFile, blobResult, submissionResult) => {
   let dataInput = []
   const csvName = uploadFile.filename && uploadFile.filename.replace(/\.[^/.]+$/, '')
   const blobFileName = blobResult && blobResult.name
   dataInput.push(csvName, blobFileName)
   dataInput = JSON.stringify(dataInput.join(','))
+  const jobStatusCode = submissionResult.errorOutput ? 'CWR' : 'COM'
```
jon-shipley: It would be better imo to define these magic strings as constants with more meaningful names.

admin/data/sql/migrations/090.do.add-bulk-permissions-pupil-table.js
GeorgeChatzigiannis: Bulk insert requires Alter table permissions on pupil table with our current db user.
The query will fail if they are not supplied.

jon-shipley: Ah, of course.

admin/services/data-access/pupil-census-import.data.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/services/data-access/pupil-census-import.data.service.js
GeorgeChatzigiannis: @jon-shipley Updated since performance should be taken greatly into account for this lookup

admin/services/pupil-census-processing.service.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Processes and bulk imports pupil census
+ * @param csvData
+ * @return {Promise<void>}
+ */
+pupilCensusProcessingService.process = async(csvData) => {
+  // Fetch all school for pupil records
+  let schoolDfeNumbers = csvData.map(r => `${r[0]}${r[1]}`)
+  // filter duplicate entries
+  schoolDfeNumbers = schoolDfeNumbers.filter((item, pos, self) => {
```
jon-shipley: Do you think it would be easier and more obvious to filter by using a `Set`?  Or `R.uniq()`?

GeorgeChatzigiannis: @jon-shipley I think R.uniq would be the cleanest way. Updated

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -24,19 +23,17 @@ pupilCensusImportDataService.sqlBulkImport = async(pupilData, schools) => {
   } catch (error) {
     result.errorOutput = error
   }
+  con.release()
```
GuyHarwood: how come you are releasing this explicitly?  are you managing the connection yourself now?

GeorgeChatzigiannis: Don't we want to release the connection back to the pool eventually?
I checked all the other use cases of `getConnection()` and `release()` is being called.

GuyHarwood: forgot that you are requesting this yourself :)

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -0,0 +1,72 @@
...
+    bulkLoad.addColumn('school_id', TYPES.Int, {nullable: false})
+    bulkLoad.addColumn('upn', TYPES.Char, {length: 13, nullable: false})
+    bulkLoad.addColumn('lastName', TYPES.NVarChar, {length: 'max', nullable: false})
+    bulkLoad.addColumn('foreName', TYPES.NVarChar, {length: 'max', nullable: false})
+    bulkLoad.addColumn('middleNames', TYPES.NVarChar, {length: 'max', nullable: true})
+    bulkLoad.addColumn('gender', TYPES.Char, {length: 1, nullable: false})
+    bulkLoad.addColumn('dateOfBirth', TYPES.DateTimeOffset, {nullable: false})
+    bulkLoad.addColumn('job_id', TYPES.Int, {nullable: true})
+
+    for (let index = 0; index < pupilData.length; index++) {
```
jon-shipley: I'm not an SQL Server expert and haven't studied the internals of the underlying Tedious library but it seems optimistic to expect the entire file to be inserted into single bulk load insert.  Memory and DB performance may be adversely affected.

What if someone passes in a large file of 600K records?  Can it cope?  What does the memory profile look like?

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+  con.release()
+  return result
+}
+
+const bulkLoadData = (connection, pupilData, schools, jobId) => {
+  return new Promise((resolve, reject) => {
+    const bulkLoad = connection.newBulkLoad(`${config.Sql.Database}.${sqlService.adminSchema}.[pupil]`, function (error, rowCount) {
+      if (error) {
+        return reject(error)
```
jon-shipley: If we load 100 records in the file, and 1 record fails validation at the DB level, do the whole 100 records get rejected?  Or do 99 get inserted and 1 rejected?

Ideally, I would want to load 99 records that are good and log everything about the record that failed.  Does the service manager get enough information here to proceed even in the event of failures and incompatibilities?

I would expect to see a process that chunks up the input, which may be very large, into suitable chunks for batch processing as it simply cannot ingest an unlimited amount of data.

admin/services/pupil-census.service.js
```diff
@@ -38,45 +34,55 @@ pupilCensusService.upload = async (uploadFile) => {
...
+}
+
+/**
+ * Updates the output of a pupilCensus record
+ * @param {Number} jobId
+ * @param {Object} submissionResult
+ * @returns {Promise.<void>}
+ */
+pupilCensusService.updateJobOutput = async (jobId, submissionResult) => {
+  const jobStatusCode = submissionResult.errorOutput ? 'CWR' : 'COM'
```
jon-shipley: magic strings

GeorgeChatzigiannis: @jon-shipley this is the job status codes we are using in the job status table. 
CWR = Completed with Errors
COM = Completed

admin/services/data-access/pupil-census-import.data.service.js
GeorgeChatzigiannis: @jon-shipley In case of a validation error during a bulk insertion attempt all the records get rejected. I am not sure if this is something we can control as this is driven by tedious as far as I see it.

A process that chunks up the input wouldn't be required at this stage as we are dealing with a small dataset of maximum 6000 records.

admin/services/pupil-census.service.js
jon-shipley: Please replace with constants

admin/services/data-access/pupil-census-import.data.service.js
GeorgeChatzigiannis: @jon-shipley At this stage we are dealing with 6K records and bulk insertion has proven to be swift enough to handle it without any issues.

admin/services/data-access/pupil-census-import.data.service.js
jon-shipley: What message does the user get then?

admin/services/data-access/pupil-census-import.data.service.js
GuyHarwood: @GeorgeChatzigiannis that's an assumption based on local machine testing.  I suggest we get this on deployed to Azure asap and test it out with 6K file upload

PR #681, opened 2018-06-04
[21417] Extend the development  phase check window

jon-shipley:
DB update to extend the check window


admin/data/sql/migrations/091.do.extend-development-check-window.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE [mtc_admin].[checkWindow]
+SET checkEndDate = '2019-02-01'
+WHERE name = 'Development phase';
```
GeorgeChatzigiannis: EOF

PR #683, opened 2018-06-04
21411 - Added UI for download-pupil-check according to prototype

danwilliams:


danwilliams:
That is correct, it is 21433 😄

test/admin/features/download_pupil_check_data.feature
```diff
@@ -0,0 +1,17 @@
...
+  As a test developer
+  In order to verify pupil check
+  I want to download pupil check data
+
+
+  Background:
+    Given I am logged in with a test developer
+
+  Scenario: Download Pupil Check Data page has a heading
+    Given I am on the download pupil check data page
```
activemq: @sinhaamit1980 change to When as Given is taken in the background...

PR #689, opened 2018-06-07
feature/20766 introduce app insights as winston destination

GuyHarwood:
Sends winston logs to application insights, when enabled.

- Refactored logging setup to dedicated file under helpers directory
- Added formal config settings for app insights key 

screenshot 2018-06-07 14 00 40

### DIFF
--- a/admin/helpers/logger.js
+++ b/admin/helpers/logger.js
@@ -1,0 +1,65 @@
+'use strict'
+
+const winston = require('winston')
+const config = require('../config')
+const morgan = require('morgan')
+
+let initialised
+
+const init = (app) => {
+  if (initialised) return
+  /**
+ * Logging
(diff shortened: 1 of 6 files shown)


admin/helpers/logger.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const winston = require('winston')
+const config = require('../config')
+const morgan = require('morgan')
+
+let initialised
+
+const init = (app) => {
+  if (initialised) return
```
GeorgeChatzigiannis: Is that really required?
In which case would you expect `init` to be called again?

GuyHarwood: that's the point - if it is called again (either accidentally or some misunderstands it), it won't bother re-adding all the transports

PR #697, opened 2018-06-13
Feature/21354 disable form assign after check startdate

danwilliams:


test/admin/features/step_definitions/assign_check_window_steps.rb
```diff
@@ -14,6 +14,19 @@
...
+  assign_form_to_window_page.check_windows.rows.each do|row|
+    if row.name_of_window.text.include?(check_window_name)
+      check_found = true
+      break
+    else
+      check_found = false
+    end
+  end
+  expect(check_found).to be_falsy, "Expected check window : '#{check_window_name}' to not appear in Assign check window list. Actual its displayed in the list"
+end
```
activemq: @sinhaamit1980 Why not try something like this - 
expect(assign_form_to_window_page.check_windows.rows.find {|row| row.name_of_window.text.include? check_window_name}).to be_nil

Thoughts?

PR #712, opened 2018-06-19
Feature/21346 Update question and loading time defaults

GeorgeChatzigiannis:


GeorgeChatzigiannis:
@GuyHarwood For the time being, yes. 
These will be the default times and I was informed that they are unlikely to change .

PR #716, opened 2018-06-20
Bug/21668 teacher1 can edit teacher2 pupil data

jon-shipley:
Fix security concerns - out of role access to pupils


admin/services/pin.service.js
```diff
@@ -52,11 +52,15 @@ pinService.expirePupilPin = async (token, checkCode) => {
 /**
  * Check and expire pupil pins based on set of provided pupil Ids
  * @param pupilIds
+ * @param {Number} schoolId - `school.id` database ID
  * @returns {Promise}
  */
 
-pinService.expireMultiplePins = async (pupilIds) => {
-  const pupils = await pupilDataService.sqlFindByIds(pupilIds)
+pinService.expireMultiplePins = async (pupilIds, schoolId) => {
```
GeorgeChatzigiannis: @jon-shipley Do you think it would also makes sense here to check for the pupilIds existence?

admin/services/psychometrician-report.service.js
```diff
@@ -5,19 +5,19 @@ const R = require('ramda')
 const moment = require('moment')
 const uuidv4 = require('uuid/v4')
 
-const config = require('../config')
-// const checkWindowDataService = require('./data-access/check-window.data.service')
 const answerDataService = require('../services/data-access/answer.data.service')
 const azureFileDataService = require('./data-access/azure-file.data.service')
 const checkFormService = require('./check-form.service')
+// const checkWindowDataService = require('./data-access/check-window.data.service')
```
GeorgeChatzigiannis: do we need this commented code?

admin/services/pin.service.js
jon-shipley: Possibly, but it's not required as the underlying data-access method will throw if they are not there.

admin/services/psychometrician-report.service.js
jon-shipley: Removed

PR #720, opened 2018-06-21
Feature/21279 Add attendance number in psychometrician report

GeorgeChatzigiannis:
The number is based on the order field


admin/services/data-access/attendance-code.data.service.js
```diff
@@ -11,7 +11,7 @@ const attendanceCodeDataService = {
    * @returns {Promise<*>}
    */
   sqlFindAttendanceCodes: async () => {
-    const sql = `SELECT id, reason, code FROM ${sqlService.adminSchema}.[attendanceCode] ORDER BY [order]`
+    const sql = `SELECT id, reason, code, [order] FROM ${sqlService.adminSchema}.[attendanceCode] ORDER BY [order]`
```
jon-shipley: I really don't see the justification in using the `order` field for a business ID.   We already have the `code` column - why can't this be used?

GeorgeChatzigiannis: @jon
the AC has been updated to use the code column instead.

GeorgeChatzigiannis:
This is blocked until changes occur to link check with pupil attendance

PR #721, opened 2018-06-21
Bug/18778 upload multiple files

danwilliams:


admin/controllers/check-form.js
```diff
@@ -123,88 +124,111 @@ const saveCheckForm = async (req, res, next) => {
...
 
-  try {
-    fileName = await checkFormService.buildFormName(uploadFile.filename)
-    if (!fileName) {
-      req.flash('error', `Select a file with no more than 128 characters in name`)
-      return res.redirect('/test-developer/upload-new-form')
+    try {
+      fileName = await checkFormService.buildFormName(uploadFile[i].filename)
+      if (!fileName) {
+        req.flash('error', `Select a file with no more than 128 characters in name - ${uploadFile[i].filename.slice(0, -4)}`)
```
jon-shipley: I noticed last night that the DB only accepts 60 chars.   We should have a consistent length.

GeorgeChatzigiannis: @danwilliams aside from this comment around the length from @jon-shipley the PR looks good. Is this going to be changed in this PR?

PR #723, opened 2018-06-21
fix failures on master

activemq:


test/admin/features/question_time_limits.feature
```diff
@@ -8,13 +8,13 @@ Feature: Question time limit tests
   Background:
     Given I am on the admin page
 
-  Scenario: Question time limit has a default value of 5 seconds
+  Scenario: Question time limit has a default value of 6 seconds
```
GuyHarwood: i did wonder if this would break due to the recent changes.  
Could you make this dynamic and read it from the db upfront? 🤔

test/admin/features/support/hooks.rb
```diff
@@ -14,8 +14,8 @@
   step "I am logged in with a service manager"
   step 'I am on the admin page'
   step 'I am on the check settings page'
-  check_settings_page.update_question_time_limit(5)
-  check_settings_page.update_loading_time_limit(2)
+  check_settings_page.update_question_time_limit(6)
```
GuyHarwood: again, could this be dynamic?

test/admin/features/question_time_limits.feature
activemq: Good idea, ill put that in now👌

jon-shipley: At one point we weren't so sure that there would be a setting in the DB at all, so there is a config fallback, and possibly a fallback if the config is missing.

I'd say it should just be in the DB, and we fail to start if it's missing.  This is a bit tangential to this work, so if we agree we can raise another PR for it.

GuyHarwood: ah yes, of course.  So basically its embedded within the admin app.

Agree with what you suggest, and ill make a tech debt item for it.

activemq: Happy to do this as a separate piece of work...if guy agrees

GuyHarwood: yes, probably best.

activemq: Sweet awaiting reciew from amit and a green build

PR #726, opened 2018-06-22
Bug/18787 No forms uploaded indication

GeorgeChatzigiannis:
screen shot 2018-06-22 at 11 55 33 am


admin/views/test-developer/upload-and-view-forms.ejs
```diff
@@ -68,6 +68,9 @@
                 </tbody>
             </table>
         </div>
+        <% } else {%>
+        <br>
+        No forms uploaded.
```
jon-shipley: SHould be in a <p> tag?

PR #735, opened 2018-06-28
Feature/21905 speech disable reading header

danwilliams:


pupil-spa/src/app/header/header.component.html
```diff
@@ -1,4 +1,4 @@
-<header role="banner" id="global-header" class="with-proposition">
+<header role="banner" id="global-header" class="with-proposition speech-hidden">
```
jon-shipley: There is already a perfectly good *attribute* for hiding element from speech-readers: 'aria-hidden'.

PR #738, opened 2018-06-28
Bug/19365 Prevent errors on disabled fields

GeorgeChatzigiannis:
- Unified check window validation schemas 
- Replaced admin start date and check start date values with existing ones when these two fields are disabled in the UI to prevent validation errors


GeorgeChatzigiannis:
Closing this as it's being refactored into a new PR

PR #739, opened 2018-06-28
Spike/22076 exploratory work on changes to check

GuyHarwood:
explores use of Shared Access Signatures to submit from SPA directly to Azure Storage Queue, which are highly scalable and would negate the need for dedicated APIs for check-start, check-completion and feedback.

There are 2 apps under the `spike-storage-queue` folder..

- **client** client side app that pushes to a queue with a SAS token
- **server-issue-sas** node app that creates a SAS token.  This will be assigned during pin generation.

Also included are some SQL files for the emerging pupil app.

POC complete.  Publishing base64 encoded completed checks from client side application.

screenshot 2018-07-03 09 27 05


jon-shipley:
@GuyHarwood - can you clarify what access do the clients have to the queue?

GuyHarwood:
@jon-shipley that depends on the type of permissions requested when the server generates the SAS token.  If you look at the server project, you will see i am generating a queue client token only (no BLOB or file access) with add permissions only.  This means they can only push messages onto queues, nothing else.

[Full SAS explanation here](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1)

spike-storage-queue/server-issue-sas/index.js
```diff
@@ -0,0 +1,44 @@
...
+  const queueName = process.env.QUEUE_NAME || 'sastest'
+
+  var queueService = azure.createQueueService(storageConnection)
+
+  // Create a SAS token that expires in an hour
+  // Set start time to five minutes ago to avoid clock skew.
+  var startDate = new Date()
+  startDate.setMinutes(startDate.getMinutes() - 5)
+  var expiryDate = new Date(startDate)
+  expiryDate.setMinutes(startDate.getMinutes() + 60)
```
jon-shipley: For the actual work, we would probably want the token to live until the end of the check window - the same as the JWT token.

GeorgeChatzigiannis: I would assume we would want to rely on moment for dates to be consistent for the real work.

jon-shipley:
@GuyHarwood  As part of the spike - did you carry out any performance testing?

spike-storage-queue/server-issue-sas/index.js
GuyHarwood: absolutely on both points.  this is a very rough POC, just to get something working.

You guys can add all the finesse and polish 🥇

GuyHarwood:
@jon-shipley No, i haven't had time for that so far.  However, i did read through [this](https://docs.microsoft.com/en-us/azure/storage/common/storage-performance-checklist) yesterday, which summarises things well, and provides detail on performance limits...

> A single queue can process approximately 2,000 messages (1KB each) per second (each AddMessage, > GetMessage, and DeleteMessage count as a message here). If this is insufficient for your application, 
> you should use multiple queues and spread the messages across them.

PR #742, opened 2018-07-02
Bug/19289 pupil list remember checks

danwilliams:


admin/assets/javascripts/global-scripts.js
```diff
@@ -227,70 +242,107 @@ $(function () {
...
+     * Update sorting link to include active parameters - group / pupil ids.
+     * @param action
+     * @param groupId
+     */
+    updateSortingLink: function (action, param, paramId) {
+      var sortingLinks = $('.sortingLink')
+      sortingLinks.each(function () {
+        var sortingLink = $(this)
+        var insertParamId = paramId + ','
+        if (sortingLink.length > 0) {
```
GeorgeChatzigiannis: @danwilliams any chance we can simplify this a bit?
My personal view is that it's not really that readable.

admin/assets/javascripts/global-scripts.js
```diff
@@ -227,70 +242,107 @@ $(function () {
...
+          if (action === 'add') {
+            var paramIndex = sortingLink.attr('href').indexOf(param + 'Ids=')
+            if (paramIndex === -1) {
```
GeorgeChatzigiannis: Perhaps as part of simplification could you change the nested ifs?

activemq:
@danwilliams There is an issue -  if you choose a couple of pupils from the filtered pupils list and then click on select all, when sorting the pupils, the pupils which were selected via select all are unticked. Happy to demo the issue

admin/assets/javascripts/global-scripts.js
```diff
@@ -228,69 +265,163 @@ $(function () {
...
+        paramIdsArr.map(function (p) {
+          paramIds.push(p)
+          checkboxUtil.updateSortingLink('add', param, p)
+        })
+        checkboxUtil.checkCheckbox(param, paramIdsArr)
+
+        var countCheckedCheckboxes = inputStatus.countCheckedCheckboxes()
+        var countAllCheckboxes = inputStatus.countCheckboxes()
+
+        if (updateCountedCheckboxes) {
```
GeorgeChatzigiannis: Just a minor change here you could avoid the nested statement by returning if the updatedCountedCheckboxes value is not set.

PR #748, opened 2018-07-05
Bug/19365 Check window submission and validation refactor

GeorgeChatzigiannis:
1. Split check window validation based on:
- Action (add/edit)
- Field
2. Remove express validator and validator schema dependency
3. Throw validation error as exception back to controller level
4. Introduce urlSlug to prevent exposing check window id in the UI

### DIFF
--- a/pupil-spa/src/app/login-success/login-success.component.spec.ts
+++ b/pupil-spa/src/app/login-success/login-success.component.spec.ts
@@ -69,9 +69,10 @@ describe('LoginSuccessComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should be created', () => {
+  it('should be created and remove pupil data', () => {
     expect(component).toBeTruthy();
     expect(appUsageService.increment).toHaveBeenCalledTimes(1);
+    expect(storageService.setItem).toHaveBeenCalledTimes(1);
   });
 
   it('asks the user to confirm their details', () => {
@@ -79,10 +80,9 @@ describe('LoginSuccessComponent', () => {
     expect(compiled.querySelector('p.lede').textContent).toMatch(/If this is you, please confirm/);
   });
 
-  it('redirects to warm up introduction page and removes pupil data', () => {
+  it('redirects to warm up introduction page', () => {
     component.onClick();
     expect(mockRouter.navigate).toHaveBeenCalledWith(['check-start']);
-    expect(storageService.setItem).toHaveBeenCalledTimes(1);
   });
 
 });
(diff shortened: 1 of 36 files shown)


admin/controllers/service-manager.js
```diff
@@ -284,6 +283,74 @@ const controller = {
...
+   */
+
+  getCheckWindowEditForm: async (req, res, next) => {
+    req.breadcrumbs('Manage check windows', '/service-manager/check-windows')
+    res.locals.pageTitle = 'Edit check window'
+    let checkWindowData
+    try {
+      checkWindowData = await checkWindowService.getCheckWindowEditForm(req.params.id)
+    } catch (error) {
+      return next()
```
danwilliams: Can we do `next(error)` to pass the error to see it in the logs too? Or is this intended, to show the listing page for non-existing check windows ids?

admin/routes/service-manager.js
```diff
@@ -11,10 +11,12 @@ router.get('/home', isAuthenticated(rolesConfig.ROLE_SERVICE_MANAGER), (req, res
 router.get('/check-windows', isAuthenticated(rolesConfig.ROLE_SERVICE_MANAGER), (req, res, next) => serviceManagerController.getCheckWindows(req, res, next))
 router.get('/check-windows/remove/:checkWindowId', isAuthenticated(rolesConfig.ROLE_SERVICE_MANAGER), (req, res, next) => serviceManagerController.removeCheckWindow(req, res, next))
 router.get('/check-windows/add', isAuthenticated(rolesConfig.ROLE_SERVICE_MANAGER), (req, res) => serviceManagerController.getCheckWindowForm(req, res))
-router.get('/check-windows/edit/:id', isAuthenticated(rolesConfig.ROLE_SERVICE_MANAGER), (req, res, next) => serviceManagerController.getEditableCheckWindowForm(req, res, next))
+// router.get('/check-windows/edit/:id', isAuthenticated(rolesConfig.ROLE_SERVICE_MANAGER), (req, res, next) => serviceManagerController.getEditableCheckWindowForm(req, res, next))
```
danwilliams: Do we need this commented line? If not, it can be removed.

admin/spec/controllers/service-manager.spec.js
```diff
@@ -1,9 +1,12 @@
 /* global jasmine describe expect it beforeEach spyOn */
+/* global jasmine describe expect it beforeEach spyOn */
```
danwilliams: This duplicate line can be removed.

admin/controllers/service-manager.js
GeorgeChatzigiannis: @danwilliams it was mistakenly forgotten and now updated

admin/routes/service-manager.js
GeorgeChatzigiannis: @danwilliams it's removed now

admin/spec/controllers/service-manager.spec.js
GeorgeChatzigiannis: @danwilliams removed

PR #749, opened 2018-07-05
Feature/19548 spa inject env vars runtime

danwilliams:


GuyHarwood:
impressed that you found a way around this.   although it does feel like a lot more moving parts than our current approach. 🤔

danwilliams:
Yes, for sure! It may not be suitable. However I think it will still work with the system constraints... and the extra moving parts only have to be set up once 😄

GuyHarwood:
yes, good point.  when ive got the HLD out of the way we should organise a time to go through it, see it in action etc

danwilliams:
Awesome! 😄

PR #750, opened 2018-07-05
Bug/22117 Catch CSRF errors

danwilliams:


danwilliams:
@GuyHarwood that’s true, because they get redirected now, but we can add a custom error for it to be displayed in the error logs if needed.

GuyHarwood:
@danwilliams yes please.  i think it would be good to have a record of these happening.

PR #751, opened 2018-07-06
Bug/22133 Remove pupil localstorage after login

danwilliams:
Ensure the pupil details (name, dob) are wiped as they are rendered on the login confirmation screen.


GuyHarwood:
@sinhaamit1980 @activemq can we ensure we get tests written to validate this - it's crucial from a pupil data security perspective.

PR #757, opened 2018-07-16
Feature/22322 admin familiarisation pins

danwilliams:


admin/assets/javascripts/global-scripts.js
```diff
@@ -642,4 +642,28 @@ $(function () {
       $('#' + $(this).context.id).text('(' + totalPupils + ' pupil' + (totalPupils > 1 ? 's' : '') + ')')
     })
   }
+
+  /**
+   * Filtering pupils by name.
+   */
+  if ($('.filter-name').length > 0) {
```
GeorgeChatzigiannis: @danwilliams @jon-shipley @GuyHarwood I am worried global scripts is becoming quite large and not easily maintainable. Would it be a right time to start moving to separate files depending on purpose?

GuyHarwood: @GeorgeChatzigiannis yes, absolutely

admin/data/sql/migrations/106.do.add-familiarisation-pin.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE [mtc_admin].[pupil] ADD familiarisation_pin NVARCHAR(12) NULL
```
GuyHarwood: we definitely don't want 2 pin fields.  the bit field distinguishes this.  We went through this in fridays discussion (you may have joined after it was discussed?)
also, we are moving to a new table `checkFormAllocation`.  I mentioned to the BAs that this card was supposed to just cover the UI, and not actually do the plumbing, as that will be done by an azure function, triggered by a queue message.

admin/controllers/questions.js
```diff
@@ -53,7 +53,7 @@ const getQuestions = async (req, res) => {
         await storeLogonEvent(data.pupil.id, schoolPin, pupilPin, false, 500, errorMessage)
         return apiResponse.serverError(res)
       }
-      await checkStartService.prepareCheck([data.pupil.id], data.school.dfeNumber, data.school.id)
+      await checkStartService.prepareCheck([data.pupil.id], data.school.dfeNumber, data.school.id, 'live')
```
GuyHarwood: this should always be familiarisation mode for the test account.

admin/data/sql/migrations/106.do.add-familiarisation-pin.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TABLE [mtc_admin].[pupil] ADD familiarisation_pinExpiresAt [datetimeoffset](3) NULL
```
GuyHarwood: same as above.

admin/data/sql/migrations/106.do.add-familiarisation-pin.sql
danwilliams: Ah okay, I was under the impression that the code from e.g. this feature would be moved to the Azure function after it was working. Should I remove the "functionality" for pin generation and address it after the Azure function / queue message is implemented to do that?

GuyHarwood: you can stub the familiarisation mode out for now.  Jon has already picked up the ticket that covers that part of it i believe.

The UI side of things looks good, so we can just wire it up when Jons done the other ticket 😎

jon-shipley:
@danwilliams  Please resolve the conflicts.

PR #758, opened 2018-07-16
Bug/18734 check window error messages

GeorgeChatzigiannis:
screen shot 2018-07-16 at 12 10 14 pm


jon-shipley:
Shouldn't the unit tests need modifying too?

GeorgeChatzigiannis:
@jon-shipley unit tests are modified as `enterValidYear` is not needed as a message and the logic is covered by another if statement with a more appropriate message

GeorgeChatzigiannis:
@jon-shipley I included some additional tests

PR #760, opened 2018-07-17
Feature/22407 create azure storage tables & queues during deployment

GuyHarwood:
Creates storage tables & queues required by MTC applications in an idempotent manner.

Build related only, does not need Travis etc


GeorgeChatzigiannis:
that was a fast merge :D

PR #765, opened 2018-07-23
Feature/22405 view and print pins page

danwilliams:


admin/views/pupil-pin/view-and-print-pins.ejs
```diff
@@ -23,72 +29,53 @@
...
-                    Pupil PINs and school password generated
-                </strong>
-                <br>
-                Valid only until 4pm, <%= date %>
-            </div>
-        </div>
-
         <p class="lede">
-            Please distribute pupil Personal Identification Numbers (PINs) and school password to your pupils.
+            Personal Identification Number (PIN) have been generated for selected pupils. These expire at 4pm daily.
```
GeorgeChatzigiannis: @danwilliams looks good just a question about this.
I can see on the other PR 4pm was changed to end of the day as text description.
Should this apply here as well?

danwilliams: It's from the prototype... perhaps that should be raised as a BA/UI/UX issue?

PR #766, opened 2018-07-23
Feature/22406 print preview page

danwilliams:


danwilliams:
@GeorgeChatzigiannis good shout - there are some rows that don’t get fully printed on the same page.

PR #767, opened 2018-07-24
Feature/22408 generate pupil pins check page

danwilliams:


admin/views/pupil-pin/generate-pins-overview.ejs
```diff
@@ -70,16 +70,50 @@
...
+              </tr>
+              </thead>
+              <tbody>
+              <tr>
+                  <td>
+                      <label>
+                          PINs have been generated for <%= pupils.length %> pupils
+                      </label>
+                      <div class="font-greyed-out font-xsmall">
+                          Expires 4pm today
```
GeorgeChatzigiannis: @danwilliams same question as the other PR.
Is this supposed to be 4pm or end of the day as description to the user?

danwilliams: Again this is down to the prototypes.

GuyHarwood: @danwilliams can you elaborate what you mean with regard to prototypes?

PR #768, opened 2018-07-24
Feature/22544 access arrangements list

GeorgeChatzigiannis:
- HTML scaffolding
- New table that includes access arrangement options


GeorgeChatzigiannis:
@jon-shipley @GuyHarwood I added feature toggling on both UI and router level.

PR #769, opened 2018-07-25
Feature/22397 check prep as azure function with table storage

jon-shipley:
Add new Azure Function `prepare-check` that reads from a queue and writes to table storage for fast pupil authentication

* adds function validation
* function unit tests / circle-ci hookup
* injects on to the prepare-check queue (disabled at present)
* creates unique JWT for each pupil


.circleci/config.yml
```diff
@@ -18,12 +18,16 @@ jobs:
           key: v1-node-pupil-spa-deps-{{ checksum "pupil-spa/yarn.lock" }}
       - restore_cache:
           key: v1-node-pupil-api-deps-{{ checksum "pupil-api/yarn.lock" }}
-      - run: cd admin; yarn install
-      - run: cd pupil-spa; yarn install
-      - run: cd pupil-api; yarn install
+      - restore_cache:
```
GuyHarwood: 🎖

functions/prepare-check/readme.md
```diff
@@ -0,0 +1,11 @@
+# QueueTrigger - JavaScript
+
+The `QueueTrigger` makes it incredibly easy to react to new Queues inside of Azure Queue Storage. This sample demonstrates a simple use case of processing data from a given Queue using C#.
```
GuyHarwood: this mentions C#, probably needs updating

GuyHarwood:
Message queueing now works.  Added base64 encoding of payload.
However - the message structure looks incorrect. It contains a single value, when it should contain the JSON structure created in `checkStartService.prepareCheckQueueMessages`

jon-shipley:
@GuyHarwood - fixed - the correct message is now placed on the queue.  Not sure what your point is re Pin Generation. It is working on this branch.  Pupil logins are working as they are in `master`

GuyHarwood:
Removed CI Enabled tag due to travis env var issues.  Will re-enable once we have that separate issue under control

GeorgeChatzigiannis:
@GuyHarwood @jon-shipley  got some conflicts on this PR.

GuyHarwood:
@GeorgeChatzigiannis merge issues resolved

admin/services/sas-token.service.js
```diff
@@ -0,0 +1,59 @@
...
+
+require('dotenv').config()
+
+const addPermissions = azure.QueueUtilities.SharedAccessPermissions.ADD
+const storageConnection = process.env.AZURE_STORAGE_CONNECTION_STRING
+
+if (!storageConnection) {
+  throw new Error('An AZURE_STORAGE_CONNECTION_STRING is a required environment variable.')
+}
+const azureQueueService = azure.createQueueService(storageConnection)
```
GuyHarwood: this is what is causing travis to fail, as it runs as soon as the app starts.  We need to explore whether we should do this once on start up, or each time it is used.  
We can add new encrypted vars next week to the travis config file.

GuyHarwood:
Closing this PR until @jon-shipley returns.  The remaining work will be in new stories, but should be picked up from this branch.  We cannot merge this in right now due to the storage connection string dependency.

GuyHarwood:
re-opened as moved storage service initialisation next to usage in method.  commented out `prepareCheck2` calls so that it never runs.  This should now pass through travis so the code can be merged, and then work commence on @jon-shipley return

admin/controllers/pupil-pin.js
```diff
@@ -95,7 +95,9 @@ const postGeneratePins = async (req, res, next) => {
   }
   let school
   try {
+    // OLD code - writes to check table
     await checkStartService.prepareCheck(pupilsList, req.user.School, req.user.schoolId, pinEnv)
```
danwilliams: The comments suggest that the old code is still running? Is this correct?

admin/controllers/pupil-pin.js
```diff
@@ -104,6 +106,11 @@ const postGeneratePins = async (req, res, next) => {
     if (update) {
       await schoolDataService.sqlUpdate(R.assoc('id', school.id, update))
     }
+
+    // New code - writes to allocateCheckFormTable, depends on school pin being ready
+    // disabled as not yet working correctly in travis.
+    // await checkStartService.prepareCheck2(pupilsList, req.user.School, req.user.schoolId, pinEnv === 'live')
```
danwilliams: Should the new code not be running now, for this feature to be complete? I was expecting this to be uncommented and replace the old code above on line 98?

admin/services/check-start.service.js
```diff
@@ -61,24 +67,83 @@ checkStartService.prepareCheck = async function (pupilIds, dfeNumber, schoolId,
...
+    const c = await checkStartService.initialisePupilCheck(pid, checkWindow, allForms, usedFormIds, pinEnv === 'live')
+    delete c.isLiveCheck // this not used in the check table (deprecated); but by the checkFormAllocation table
     checks.push(c)
   }
   await checkDataService.sqlCreateBatch(checks)
 }
 
+// PrepareCheck 2: Write an entry to the checkFormAllocationTable, and place a message on the queue
+// for writing to the data store used by pupil authentication.
+checkStartService.prepareCheck2 = async function (pupilIds, dfeNumber, schoolId, isLiveCheck) {
```
danwilliams: The new function has been introduced but the old function is still present. Should we not be removing the old function? Should the new function not be replacing it wholesale? I am not keen on the appending of "2" - is there a need to keep the old function around... there is an implication for future work to clean up the references to the "2" function and rename this to drop the "2"?

The new function duplicates a lot of the code in the original function, which I find concerning. I feel there should be a better way to achieve this.

The new function uses the Azure queue, but I would have thought that removal of unused code and/or migration to the new version would be included, without this, maintaining which function is actually used is quite difficult.

admin/services/check-start.service.js
```diff
@@ -122,4 +195,54 @@ checkStartService.pupilLogin = async function (pupilId) {
...
+    throw new Error('checkFormAllocationIds must be an array')
+  }
+
+  const messages = []
+  const checkFormAllocations = await checkFormAllocationDataService.sqlFindByIdsHydrated(checkFormAllocationIds)
+  const sasExpiryDate = moment().add(config.Tokens.sasTimeOutHours, 'hours')
+  const sasToken = sasTokenService.generateSasToken(sasTokenService.queueNames.CHECK_COMPLETE, sasExpiryDate)
+
+  for (let o of checkFormAllocations) {
+    const config = await configService.getConfig({id: o.pupil_id}) // ToDo: performance note: this does 2 sql lookups per pupil. Optimise!
```
danwilliams: It seems like there could be some optimisations made by requesting all/batches of pupil data at once instead of synchronously waiting for each pupil's config to load inside a loop. Also, is there a limit to the number of messages that can be input into Azure queue at any time, because this could potentially be rejected if there is a limit?

I note there is also a "todo" here - I feel that ideally we should not be merging in todos? Could this not be dealt with now?

admin/services/sas-token.service.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const azure = require('azure-storage')
+const moment = require('moment')
+const config = require('../config')
+
+const addPermissions = azure.QueueUtilities.SharedAccessPermissions.ADD
+const storageConnection = process.env.AZURE_STORAGE_CONNECTION_STRING
```
danwilliams: It would be nice to get a default in the config file for this, otherwise you wouldn't know what env variable to set.

admin/spec/back-end/controllers/pupil-pin.spec.js
```diff
@@ -232,6 +232,7 @@ describe('pupilPin controller:', () => {
         const controller = require('../../../controllers/pupil-pin.js').postGeneratePins
 
         spyOn(checkStartService, 'prepareCheck')
+        spyOn(checkStartService, 'prepareCheck2')
```
danwilliams: Tests seem to refer to and spy on the new version of the function, but don't actually cover any test cases.

admin/spec/back-end/controllers/questions.spec.js
```diff
@@ -132,11 +135,14 @@ describe('Questions controller', () => {
       expect(getConfigSpy).toHaveBeenCalledTimes(1)
       expect(jwtSpy).toHaveBeenCalledTimes(1)
       expect(prepareQuestionDataSpy).toHaveBeenCalledTimes(1)
-      done()
     })
   })
 
   describe('unhappy paths', () => {
+    beforeEach(() => {
+      spyOn(winston, 'error') // ignore winston messages from unhappy outcomes
```
danwilliams: This bit sounds risky - if an error is thrown by winston then something unexpected must have happened?

admin/services/check-start.service.js
```diff
@@ -61,24 +67,83 @@ checkStartService.prepareCheck = async function (pupilIds, dfeNumber, schoolId,
...
-    const c = await checkStartService.initialisePupilCheck(pid, checkWindow, allForms, usedFormIds, pinEnv)
+    const c = await checkStartService.initialisePupilCheck(pid, checkWindow, allForms, usedFormIds, pinEnv === 'live')
+    delete c.isLiveCheck // this not used in the check table (deprecated); but by the checkFormAllocation table
     checks.push(c)
   }
   await checkDataService.sqlCreateBatch(checks)
 }
 
+// PrepareCheck 2: Write an entry to the checkFormAllocationTable, and place a message on the queue
+// for writing to the data store used by pupil authentication.
```
danwilliams: Code comments are not standard jsdoc, hinting of parameters is not included.

functions/prepare-check/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+function validateMessage (message) {
+  const check = (prop) => { if (R.not(R.has(prop, message))) { throw new Error(`Message failed validation check: missing field: ${prop}`) } }
+  const topLevelProperties = ['schoolPin', 'pupilPin', 'pupil', 'school', 'tokens', 'config', 'questions']
+  topLevelProperties.map(check)
+  validatePupil(message.pupil)
+  validateQuestions(message.questions)
+  validateSchool(message.school)
+  validateConfig(message.config)
+  return true
```
danwilliams: Returns are pointless as they are never checked; it runs until an exception is thrown.

functions/prepare-check/index.js
```diff
@@ -0,0 +1,99 @@
+const R = require('ramda')
+
+/**
+ * Write to Table Storage for fast pupil authentication
+ * Reads from a queue
+ * @param context
+ * @param {} prepareCheckMessage
+ */
+module.exports = function (context, prepareCheckMessage) {
```
danwilliams: For the queue runner I can see performance implications if there is a singular message being processed at once with the round trip time. Since it's a polled solution it may be worthwhile to obtain batches at a time and process them concurrently if possible, to improve performance.

admin/config.js
```diff
@@ -99,5 +99,10 @@ module.exports = {
   },
   Cors: {
     Whitelist: process.env.CORS_WHITELIST || 'http://localhost:4200' // for development
+  },
+  Tokens: {
+    // 12 hours default expiry
+    jwtTimeOutHours: process.env.JWT_TIMEOUT_HOURS || 12,
+    sasTimeOutHours: process.env.SAS_TIMEOUT_HOURS || 12,
```
danwilliams: I personally like, and agree with, trailing commas, but this is raising a linting error. Should we update the linting rules? I do agree with this line.

admin/services/jwt.service.js
```diff
@@ -4,6 +4,7 @@ const moment = require('moment')
 const crypto = Promise.promisifyAll(require('crypto'))
 const jwt = Promise.promisifyAll(require('jsonwebtoken'))
 const uuidv4 = require('uuid/v4')
+const config = require('../config')
```
danwilliams: Linting error: 'config' is assigned a value but never used.

admin/controllers/questions.js
```diff
@@ -3,14 +3,18 @@
...
 const jwtService = require('../services/jwt.service')
+const moment = require('moment')
 const pupilAuthenticationService = require('../services/pupil-authentication.service')
+const pupilDataService = require('../services/data-access/pupil.data.service')
 const pupilLogonEventService = require('../services/pupil-logon-event.service')
-const checkWindowService = require('../services/check-window.service')
 const R = require('ramda')
 const winston = require('winston')
 
+
```
danwilliams: Linting error: More than 1 blank line not allowed.

admin/controllers/pupil-pin.js
GuyHarwood: yes

admin/controllers/pupil-pin.js
GuyHarwood: see above

admin/services/check-start.service.js
GuyHarwood: again, see above

admin/services/check-start.service.js
GuyHarwood: this is in a separate card along with some other optimisations

admin/services/sas-token.service.js
GuyHarwood: Jon has changed this so it is no longer required, for the benefit of unit testing.
As we progress with the overall implementation we can re-assess this.

functions/prepare-check/index.js
GuyHarwood: this should be set to `context.done()`

admin/spec/back-end/controllers/questions.spec.js
GuyHarwood: this is covered in PBI 23255

functions/prepare-check/index.js
GuyHarwood: the listener is abstract to the function implementation, we just declare the binding.  We can configure some processing behaviours, but we can explore that once we have the solution wired up and working.

admin/controllers/questions.js
GuyHarwood: This should now be fixed as i have re-enabled codacy and standardJS should enforce this by default.

admin/services/jwt.service.js
GuyHarwood: This should now be fixed as i have re-enabled codacy and standardJS should enforce this by default.

admin/config.js
jon-shipley: GDS pretty specify that standard is the linter of choice.   `standard` is not configurable, so I will update this trailing comma in another branch.

admin/services/sas-token.service.js
jon-shipley: Good spot, I'll fix this on the next PR.

admin/spec/back-end/controllers/questions.spec.js
jon-shipley: Not really.  Winston is only od a logger, and this prevents spurious console output during the test run.   Is there a better method?

functions/prepare-check/index.js
jon-shipley: Actually the `return true` was written for the unit test.  Happy to reconsider.  If an error is thrown then the top level function will handle it and call `context.done(error)`

PR #770, opened 2018-07-27
Feature/22409 custom print page

danwilliams:


admin/controllers/pupil-pin.js
```diff
@@ -152,6 +152,51 @@ const getViewAndPrintPins = async (req, res, next) => {
...
+  let groups
+  let school
+  let error
+  const date = dateService.formatDayAndDate(new Date())
+  try {
+    pupils = await pinService.getPupilsWithActivePins(req.user.School, pinEnv)
+    groupsForPupils = await groupService.getGroupsAsArray(req.user.schoolId)
+    school = await pinService.getActiveSchool(req.user.School)
+    error = await checkWindowSanityCheckService.check()
+    if (pupils.length > 0) {
```
GeorgeChatzigiannis: Could avoid the nested if here with a guard clause.
Also I would think this logic could be part of a service?

admin/controllers/pupil-pin.js
```diff
@@ -191,10 +236,66 @@ const getPrintPins = async (req, res, next) => {
...
+  if (!Array.isArray(pupilsList) || pupilsList.length === 0) {
+    return res.redirect(`/pupil-pin/view-and-print-${pinEnv}-pins-list`)
+  }
+  res.locals.pinEnv = pinEnv
+  res.locals.pageTitle = 'Print pupils'
+  let groups
+  let pupils
+  let school
+  let qrDataURL
+  const date = dateService.formatDayAndDate(new Date())
```
GeorgeChatzigiannis: Same as the comments above around passing a date to this method.

PR #772, opened 2018-07-30
Feature/22576 timeout notification minute countdown

danwilliams:


GuyHarwood:
@danwilliams @GeorgeChatzigiannis given that George recently added front end test framework support, it would be best to ensure new code follows suit.

GeorgeChatzigiannis:
@danwilliams looks good but has some conflicts based on the last merge

PR #773, opened 2018-07-30
Fixed the lint errors on master

danwilliams:


admin/app.js
```diff
@@ -73,6 +72,7 @@ const app = express()
 
 setupBrowserSecurity(app)
 setupLogging(app)
+// const checkConfigWhitelist = require('./helpers/whitelist-dev')
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: Actually wrong question, if this is not being used I suspect the commented code can be deleted.
Any particular reason to keep it?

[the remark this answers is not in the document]
danwilliams: I thought it was "temporarily disabled" similar to the verifySignature in `admin/services/nca-tools-auth.service.js`, that’s why I didn’t remove it completely. I wasn’t sure, but I've deleted it now :smile: It's probably right to do so.

admin/services/nca-tools-auth.service.js
```diff
@@ -58,7 +58,7 @@ const service = {
 module.exports = service
 
 /**
- * Verify the provided signature is valid.
+ * Verify the provided signature is valid. - commented out at the moment, see the check on line 25
```
GuyHarwood: we do need to come back to this and discuss with 3rd party (we have a backlog item for it).

admin/app.js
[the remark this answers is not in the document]
GuyHarwood: ironically we are about to introduce a required env var for azure storage.

However, we should deal with that separately.  let's get the code removed for now.

admin/app.js
```diff
@@ -17,7 +17,6 @@ const TediousSessionStore = require('connect-tedious')(session)
 const breadcrumbs = require('express-breadcrumbs')
 const flash = require('connect-flash')
 const config = require('./config')
-const checkConfigWhitelist = require('./helpers/whitelist-dev')
```
jon-shipley: Why is this commented out?  Isn't it still required?

admin/app.js
```diff
@@ -73,7 +72,6 @@ const app = express()
 
 setupBrowserSecurity(app)
 setupLogging(app)
-// checkConfigWhitelist(app)
```
jon-shipley: Why is this removed?  Isn't it still required?

danwilliams: @jon-shipley my understanding from @GuyHarwood is that this is no longer required.

GuyHarwood:
@danwilliams correct.  once you have a green build on Travis you should be able to merge this PR

PR #776, opened 2018-08-01
Access arrangements list typeahead component

GeorgeChatzigiannis:
- Reusable typeahead component based on GDS [accessible autocomplete component](https://github.com/alphagov/accessible-autocomplete)
- Applied use of case of the component on access arrangements list form

Example found here:
https://gyazo.com/050f4635d0215fd1d9dd1bb955ce57df


admin/assets/javascripts/typeahead.js
```diff
@@ -0,0 +1,29 @@
...
+ * Typeahead component
+ */
+
+/* global $ */
+$(function () {
+  'use strict'
+  if (!window.GOVUK) {
+    window.GOVUK = {}
+  }
+  window.GOVUK.typeahead = {
```
GeorgeChatzigiannis: How do you feel about this naming convention?
Shall I rather change it to autoComplete to match gds naming?

admin/assets/stylesheets/elements/_accessible-autocomplete.scss
```diff
@@ -0,0 +1,134 @@
+.autocomplete__wrapper {
```
GeorgeChatzigiannis: This is the css code that govuk accessible component uses which unfortunately is not by default offered as a scss module.

admin/services/pupil.service.js
```diff
@@ -99,4 +99,20 @@ pupilService.sortByGroup = (pupilsList, sortDirection) => {
   })
 }
 
+/**
```
GeorgeChatzigiannis: Would you prefer this to be a separete service with a separete method since it's only utilised so far in access arrangements?

admin/assets/javascripts/typeahead.js
GuyHarwood: can do.  don't mind either way, they are both decent names

GuyHarwood: i notice in createComponent below, you are using autocomplete as var names.  so maybe use that

GeorgeChatzigiannis: @GuyHarwood thought so as well for naming consistency, it's done.

GeorgeChatzigiannis:
Codacy fails only on the minified library file not much that can be done other than ignoring it.

PR #777, opened 2018-08-03
Feature/22545 Access arrangements selection

GeorgeChatzigiannis:
- UI changes and behavior based on user interaction


admin/data/sql/migrations/110.do.create-question-reader-reasons.sql
```diff
@@ -0,0 +1,18 @@
...
+  CONSTRAINT [PK_questionReaderReasons] PRIMARY KEY CLUSTERED ([id] ASC)
+    WITH (
+      PAD_INDEX = OFF,
+      STATISTICS_NORECOMPUTE = OFF,
+      IGNORE_DUP_KEY = OFF,
+      ALLOW_ROW_LOCKS = ON,
+      ALLOW_PAGE_LOCKS = ON
+    ),
+  CONSTRAINT [questionReaderReasons_code_uindex] UNIQUE([code])
+)
```
jon-shipley: Should we have a convention that require's semi-colons to terminate SQL statements?  Not having one looks odd to me, and I wonder if it could cause trouble.

admin/data/sql/migrations/112.do.create-pupil-access-arrangements.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE [mtc_admin].[pupilAccessArrangements] (
+  id int IDENTITY (1,1) NOT NULL,
+  pupil_id int NOT NULL,
+  accessArrangements_id int NOT NULL,
+  questionReaderReasons_id int NULL,
+  createdAt datetimeoffset(3) NOT NULL DEFAULT GETUTCDATE(),
+  updatedAt datetimeoffset(3) NOT NULL DEFAULT GETUTCDATE(),
+  version rowversion,
+  recordedByUser_id int NOT NULL,
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: We are using the same convention on `pupilRestart`.
If we are to change it here it should also change across other tables with similar conventions.
I'll wait for your conclusion.

admin/data/sql/migrations/110.do.create-question-reader-reasons.sql
GuyHarwood: commas are correct here, as its part of the create table statement.
very different to how i would usually create constraints, which would be separate statements.

maybe Georges SQL tool does it this way

admin/data/sql/migrations/112.do.create-pupil-access-arrangements.sql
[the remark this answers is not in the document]
GuyHarwood: if i recall correctly, our usual convention is `recordedBy_userId` and the underscore denotes the separation of purpose from linked column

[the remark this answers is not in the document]
jon-shipley: Let's keep our current convention. 👍

admin/spec/front-end/pupil-access-arrangements-selection.spec.js
```diff
@@ -0,0 +1,188 @@
...
+    '        </div>' +
+    '        <div class="multiple-choice-mtc">' +
+    '            <input id="accessArrangement-6" name="accessArrangement[6]" type="checkbox" value="RON"' +
+    '                   aria-label="Tick accessArrangement Remove on-screen number pad." aria-checked="false"' +
+    '                   role="checkbox">' +
+    '            <div></div>' +
+    '        </div>' +
+    '    </li>')
+
+  $accessArrangementsList.append($accessArrangementsListItems)
```
GuyHarwood: if you were to make the `pupil-access-arrangements-selection.js` injectable so that `window.GOVUK` could be mocked, would all this html be necessary?  also, could you not build up the DOM programatically instead of having all the mark up?

GeorgeChatzigiannis: @GuyHarwood field type is change to `recordedBy_user_id ` and test's DOM is being generated instead of statically declared.

GeorgeChatzigiannis:
@danwilliams that is correct submission is currently WIP separately.

PR #781, opened 2018-08-13
23258 - Add new services to prefix azure tables and queues

jon-shipley:
# WIP

### DIFF
--- a/admin/services/queue-name-service.js
+++ b/admin/services/queue-name-service.js
@@ -1,0 +1,36 @@
+'use strict'
+const config = require('../config')
+
+const NAMES = {
+  PREPARE_CHECK: 'prepare-check',
+  CHECK_COMPLETE: 'check-complete',
+  PUPIL_FEEDBACK: 'pupil-feedback'
+}
+
+Object.freeze(NAMES)
+
+const nameService = {
+  /**
+   * Returns the correct queue name for the environment based on a key.
+   * E.g.
+   * @param qName - the basic queue name
+   * @return {string}
+   */
+  getName: function (qName) {
+    const prefix = config.Azure.queuePrefix
+    if (!NAMES.hasOwnProperty(qName.toUpperCase().replace('-', '_'))) {
+      throw new Error('Queue not found: ' + qName)
+    }
+
+    // If the config prefix is an empty string - there is nothing to do, so return the queue name as it is.
+    if (!prefix) {
+      return qName
+    }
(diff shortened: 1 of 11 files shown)


admin/services/queue-name-service.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+const config = require('../config')
+
+const NAMES = {
+  PREPARE_CHECK: 'prepare-check',
+  CHECK_COMPLETE: 'check-complete',
+  PUPIL_FEEDBACK: 'pupil-feedback'
+}
+
+Object.freeze(NAMES)
```
GeorgeChatzigiannis: why freezing it? is there a chance this can be mutated somehow?

admin/services/queue-name-service.js
```diff
@@ -0,0 +1,36 @@
...
+
+    // If the config prefix is an empty string - there is nothing to do, so return the queue name as it is.
+    if (!prefix) {
+      return qName
+    }
+    // Queue naming info: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-queues-and-metadata
+    return `${prefix}-${qName}`
+  },
+
+  NAMES
```
GeorgeChatzigiannis: is that line intentional to be this way?

admin/services/table-name-service.js
```diff
@@ -0,0 +1,51 @@
...
+      throw new Error('Table name not found: ' + tableName)
+    }
+
+    // If the config prefix is an empty string - there is nothing to do, so return the queue name as it is.
+    if (!prefix) {
+      return tableName
+    }
+
+    // Queue naming info: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-queues-and-metadata
+    const stem = capitalise(tableName)
```
GeorgeChatzigiannis: I read the doc and if I understand correctly on number 3 it states 'all letters in a queue name must be lowercase'.
In your capitalise function your first param will be uppercased and the rest will be lower cased.
Am I missing something?

admin/services/queue-name-service.js
GeorgeChatzigiannis: It's useful indeed, I am wondering what would be the scenario where this would be somehow modified.

admin/services/queue-name-service.js
GeorgeChatzigiannis: yea it makes sense thanks.

admin/services/queue-name-service.js
jon-shipley: It's exported.   This prevents meddling.

PR #783, opened 2018-08-14
Feature/23410 pupil api prep and tidy up

jon-shipley:
Delete files that aren't needed for the pupil-api now that it will only be handling auth via table storage


pupil-api/package.json
```diff
@@ -18,24 +16,16 @@
...
     "morgan": "^1.9.0",
-    "mssql": "^4.1.0",
-    "postgrator": "^3.6.0",
-    "ramda": "^0.25.0",
-    "random-number-csprng": "^1.0.2",
-    "tedious": "^2.3.1",
-    "tedious-connection-pool": "^1.0.5",
     "to-bool": "0.0.1",
-    "uuid": "^3.1.0",
+    "uuid": "^3.3.2",
```
GuyHarwood: you will also need `azure-storage` package

pupil-api/src/config.ts
```diff
@@ -14,43 +11,13 @@ export = {
...
     }
   },
   Environment: getEnvironment(),
-  JwtSecret: process.env.JWT_SECRET || 'dev-token'.padEnd(64, '-'),
   Endpoints: {
-    Auth: process.env.API_AUTH || false,
-    CheckStart: process.env.API_CHECK_START || false,
-    CheckSubmit: process.env.API_CHECK_SUBMIT || false,
-    Feedback: process.env.API_FEEDBACK || false
+    Auth: process.env.API_AUTH || false
```
GuyHarwood: you could probably just ditch the whole thing now, as it will only ever have one endpoint

pupil-api/package.json
jon-shipley: That won't be needed until the PR that implements the lookup though, which is planned as a subsequent PR.  This PR is just a tidy up.

GuyHarwood: good point

pupil-api/src/config.ts
jon-shipley: Yes, good point.  I'll push an update.

danwilliams:
I see that the migrations are deleted instead of reverted, were they not applied? Or does it basically not matter, seeing as the `mtc_pupil` database will be dropped too, now that we’re not using it anymore?

pupil-api/src/services/jwt.service.ts
```diff
@@ -1,82 +0,0 @@
-'use strict'
```
danwilliams: My understanding is that this endpoint is not responsible for checking the pupil’s JWT token validity anymore, as part of the table storage changes.

jon-shipley: That's correct.  The SQL database is replaced by a high-performance data-store.

jon-shipley:
@danwilliams 

The migrations are no longer needed as the whole SQL database is replaced by a simple lookup in Table Storage.

Thanks for the review 👍

PR #786, opened 2018-08-15
Feature/23141 Familiarisation exclude left-school pupils

danwilliams:
Updated the valid attendance to check for left school on familiarisation


admin/services/attendance.service.js
```diff
@@ -64,6 +64,19 @@ const attendanceService = {
    */
   getAttendanceCodes: async () => {
     return attendanceCodeDataService.sqlFindAttendanceCodes()
+  },
+
+  /**
+   * Check if pupil has valid attendance
```
jon-shipley: What is a valid attendance?

admin/services/attendance.service.js
```diff
@@ -64,6 +64,19 @@ const attendanceService = {
...
+   * @param pupilId
+   * @param pinEnv
+   * @return {boolean}
+   */
+  hasValidAttendance: async (pupilId, pinEnv) => {
+    const pupilAttendance = await pupilAttendanceDataService.findOneByPupilId(pupilId)
+    return pinEnv === 'live'
+      ? pupilAttendance && pupilAttendance.id
```
jon-shipley: This returns either the `id` or the `attendanceCode_id` rather than the bool suggested in the jsdoc won't it?

admin/services/attendance.service.js
danwilliams: You're right - "valid" might be a bit confusing, so I have changed it to `hasAttendance` instead of `hasValidAttendance`.

admin/services/attendance.service.js
danwilliams: I've added some more comments in the jsdoc to explain what it does (plus fixed that bool thing).

admin/services/attendance.service.js
```diff
@@ -64,6 +64,24 @@ const attendanceService = {
...
+   *
+   * All attendance entries are considered for live checks, while for
+   * familiarisation checks, only `left school` entries are considered
+   *
+   * @param pupilId
+   * @param pinEnv
+   * @return {boolean}
+   */
+  hasAttendance: async (pupilId, pinEnv) => {
+    const pupilAttendance = await pupilAttendanceDataService.findOneByPupilId(pupilId)
```
jon-shipley: Which roles would be performing this action?

admin/spec/back-end/service/attendance-service.spec.js
```diff
@@ -162,4 +162,40 @@ describe('attendanceService', () => {
       done()
     })
   })
+
+  describe('#hasAttendance', () => {
+    describe('for live env', () => {
+      it('returns valid if pupil has any attendance', async () => {
+        spyOn(pupilAttendanceDataService, 'findOneByPupilId').and.returnValue({ id: 'id', attendanceCode_id: 100 })
+        const result = await service.hasAttendance('id', 'live')
+        expect(result).toBeTruthy()
```
jon-shipley: I would prefer this test to actually test for the boolean that is returned from `hasAttendance()`.

admin/services/attendance.service.js
```diff
@@ -64,6 +64,24 @@ const attendanceService = {
...
+    return pinEnv === 'live'
+      ? pupilAttendance && pupilAttendance.attendanceCode_id !== undefined
+      : pupilAttendance && pupilAttendance.attendanceCode_id === 2 // left school
```
GeorgeChatzigiannis: Is it intentional to only take into account `left school` attendance id (2) as attendance for non live environments?

admin/services/attendance.service.js
danwilliams: If you mean user roles, this will be called when teachers get the list of pupils to generate pins for.

admin/spec/back-end/service/attendance-service.spec.js
danwilliams: Good catch - updated.

admin/services/attendance.service.js
danwilliams: Yes... but, thinking about it, it is a magic number, which is not great. There's no constant in the codebase, but I've updated the query that checks the user code to also do a join and get the code, and changed the codebase to check the code instead of the ID.

admin/services/attendance.service.js
danwilliams: That’s not directly called by the teacher (by the controller), it’s used by `pinGenerationService.isValid` which is, as you mention, called by `pinGenerationService.getPupils(dfeNumber, pinEnv)` with `dfeNumber`.

The idea is that `getPupils` will get the pupils list using the `dfeNumber`, then will check if the pupils are valid using `isValid` (which also calls `hasAttendance` for a specific pupil) — this functionality was previously in `isValid`. I thought it makes more sense to put it in that separate function to keep the `isValid` one cleaner, and write unit tests for `hasAttendance` in particular 😄

jon-shipley: That seems reasonable to me.  👍

PR #787, opened 2018-08-15
Feature/23432 Submission & validation on pupil access arrangements

GeorgeChatzigiannis:
- [x] Submission
- [x]  Validation

### DIFF
--- a/admin/assets/stylesheets/elements/_flash-messages.scss
+++ b/admin/assets/stylesheets/elements/_flash-messages.scss
@@ -13,20 +13,8 @@
   a:hover {
     color: $green;
   }
-}
-
-.info-message-clickable {
-  border: 5px solid $green;
-  padding: 1em;
-  margin-top: 40px;
-  color: $green;
-  font-weight: bold;
-  text-decoration: underline;
-  a,
-  a:link,
-  a:visited,
-  a:active,
-  a:hover {
+  &.clickable {
+    text-decoration: underline;
     color: $green;
   }
 }
--- a/admin/assets/stylesheets/elements/_flash-messages.scss
+++ b/admin/assets/stylesheets/elements/_flash-messages.scss
@@ -15,6 +15,22 @@
   }
 }
 
+.info-message-clickable {
+  border: 5px solid $green;
+  padding: 1em;
+  margin-top: 40px;
+  color: $green;
+  font-weight: bold;
+  text-decoration: underline;
+  a,
+  a:link,
+  a:visited,
+  a:active,
+  a:hover {
+    color: $green;
+  }
+}
+
 // Green border to highlight table cell content
 .highlight-item {
   display: flex;
--- a/admin/lib/errors/access-arrangements.js
+++ b/admin/lib/errors/access-arrangements.js
@@ -1,8 +1,9 @@
 'use strict'
 
 module.exports = {
-  missingPupilName: '"Enter a pupil name',
-  missingAccessArrangements: 'Select an access arrangement',
-  missingExplanation: 'Enter an explanation',
+  missingPupilName: 'Enter a pupil name',
+  missingAccessArrangements: 'Select at least one access arrangement',
+  missingInputAssistanceExplanation: 'Enter an explanation for input assistance',
+  missingScreenReaderExplanation: 'Enter an explanation for inbuilt screen reader',
   missingQuestionReaderReason: 'Select a reason'
 }
--- a/admin/lib/errors/access-arrangements.js
+++ b/admin/lib/errors/access-arrangements.js
@@ -1,0 +1,8 @@
+'use strict'
+
+module.exports = {
+  missingPupilName: '"Enter a pupil name',
+  missingAccessArrangements: 'Select an access arrangement',
+  missingExplanation: 'Enter an explanation',
+  missingQuestionReaderReason: 'Select a reason'
+}
--- a/admin/lib/errors/access-arrangements.js
+++ b/admin/lib/errors/access-arrangements.js
@@ -4,6 +4,6 @@ module.exports = {
   missingPupilName: 'Enter a pupil name',
   missingAccessArrangements: 'Select at least one access arrangement',
   missingInputAssistanceExplanation: 'Enter an explanation for input assistance',
-  missingScreenReaderExplanation: 'Enter an explanation for inbuilt screen reader',
+  missingScreenReaderExplanation: 'Enter an explanation for question reader',
   missingQuestionReaderReason: 'Select a reason'
 }
--- a/admin/services/access-arrangements.service.js
+++ b/admin/services/access-arrangements.service.js
@@ -42,20 +42,20 @@ accessArrangementsService.submit = async (requestData, dfeNumber, userId) => {
  * @returns {Object}
  */
 accessArrangementsService.process = async (requestData, pupil, dfeNumber, userId) => {
-  const { accessArrangements: accessArrangementsCodes, questionReaderReason: questionReaderReasonCode } = requestData
+  const { accessArrangements: accessArrangementsCodes, questionReaderReason } = requestData
   const pupilAccessArrangements = R.clone(requestData)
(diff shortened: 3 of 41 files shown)


admin/lib/errors/access-arrangements.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+module.exports = {
+  missingPupilName: '"Enter a pupil name',
```
jon-shipley: Is `'"` a typo?

admin/services/access-arrangements.service.js
```diff
@@ -10,4 +16,65 @@ accessArrangementsService.getAccessArrangements = async () => {
...
+ * @param {Number} userId
+ * @returns {Object}
+ */
+accessArrangementsService.submit = async (requestData, dfeNumber, userId) => {
+  const validationError = accessArrangementsValidator.validate(requestData)
+  if (validationError.hasError()) {
+    throw validationError
+  }
+  const { pupilUrlSlug, accessArrangements: accessArrangementsCodes, questionReaderReason: questionReaderReasonCode } = requestData
+  let pupilAccessArrangements = R.clone(requestData)
```
jon-shipley: It's surprising to see `let` being used here when care is being taken to clone the input data.  Could this be const?

admin/services/access-arrangements.service.js
```diff
@@ -10,4 +16,65 @@ accessArrangementsService.getAccessArrangements = async () => {
...
+  const accessArrangements = await accessArrangementsDataService.sqlFindAccessArrangementsByCodes(accessArrangementsCodes)
```
jon-shipley: I don't think there is any plan to support changing access codes during runtime, so this call could be `memoized()`

admin/services/access-arrangements.service.js
```diff
@@ -10,4 +16,65 @@ accessArrangementsService.getAccessArrangements = async () => {
...
+  }
+  pupilAccessArrangements['accessArrangements_ids'] = JSON.stringify(accessArrangements.map(aa => aa.id))
+  const pupil = await pupilDataService.sqlFindOneBySlugAndSchool(pupilUrlSlug, dfeNumber)
+  if (!pupil) {
+    throw new Error('Pupil url slug does not match a pupil record')
+  }
+  pupilAccessArrangements['pupil_id'] = pupil.id
+  pupilAccessArrangements['recordedBy_user_id'] = userId
+  if (!pupilAccessArrangements.accessArrangements.includes('ITA')) {
+    pupilAccessArrangements = R.omit(['inputAssistanceInformation'], pupilAccessArrangements)
```
jon-shipley: IMHO this is just weird.  You are using `omit` to get a new data structure, and then proceed to re-assign the reference to it.   This is incongruous; if we are using immutable objects then please be consistent.

admin/services/access-arrangements.service.js
```diff
@@ -10,4 +16,65 @@ accessArrangementsService.getAccessArrangements = async () => {
...
+  }
+  if (pupilAccessArrangements.questionReaderReason !== 'OTH') {
+    pupilAccessArrangements = R.omit(['questionReaderOtherInformation'], pupilAccessArrangements)
+  }
+  if (pupilAccessArrangements.accessArrangements.includes('QNR')) {
+    questionReaderReason = await questionReaderReasonsDataService.sqlFindQuestionReaderReasonByCode(questionReaderReasonCode)
```
jon-shipley: Another one that could be `memoized()`?

admin/services/access-arrangements.service.js
```diff
@@ -10,4 +16,65 @@ accessArrangementsService.getAccessArrangements = async () => {
...
+accessArrangementsService.save = async (pupilAccessArrangements, pupil) => {
+  const pupilAccessArrangement = await pupilAccessArrangementsDataService.sqlFindPupilAccessArrangementsByPupilId(pupil.id)
+  if (pupilAccessArrangement && pupilAccessArrangement['pupil_id']) {
+    // update
+    await pupilAccessArrangementsDataService.sqlUpdate(pupilAccessArrangements)
+  } else {
+    // create
+    await pupilAccessArrangementsDataService.sqlCreate(pupilAccessArrangements)
+  }
+  return pupil
```
jon-shipley: Is this return actually useful?

admin/services/data-access/access-arrangements.data.service.js
```diff
@@ -17,4 +20,39 @@ accessArrangementsDataService.sqlFindAccessArrangements = async function () {
...
+ * @returns {Promise<Array>}
+ */
+accessArrangementsDataService.sqlFindAccessArrangementByPupilId = async function () {
+  const sql = `
+  SELECT 
+    id, 
+    code, 
+    description
+  FROM ${sqlService.adminSchema}.[accessArrangements]
+  WHERE`
```
jon-shipley: unfinished?

admin/lib/errors/access-arrangements.js
GeorgeChatzigiannis: @jon-shipley yea i fixed it earlier. :)

admin/services/access-arrangements.service.js
GeorgeChatzigiannis: @jon-shipley this pattern is now updated

admin/services/access-arrangements.service.js
GeorgeChatzigiannis: @jon-shipley updated as per our conversation

admin/services/access-arrangements.service.js
GeorgeChatzigiannis: @jon-shipley this is now removed and the pattern has changed as per our conversation

admin/services/access-arrangements.service.js
GeorgeChatzigiannis: @jon-shipley I need to use on controller level the id, foreName and lastName of the pupil for the highlight message.
Perhaps I could return a simplified object with just these three properties.
Any objections?

admin/services/data-access/access-arrangements.data.service.js
GeorgeChatzigiannis: @jon-shipley it's removed since it was never used.

admin/assets/stylesheets/elements/_flash-messages.scss
```diff
@@ -15,6 +15,22 @@
...
+  text-decoration: underline;
+  a,
+  a:link,
+  a:visited,
+  a:active,
+  a:hover {
+    color: $green;
+  }
+}
+
```
danwilliams: Hmmmm, does this have to be explicitly underlined? I thought `.info-message` div texts will be already underlined when there’s a link inside them. If there's text (not a link) that's underlined, won't that be confusing to the users?

...if that should have the exact info-message properties with the addition that it has to be explicitly underlined, maybe an inner `&.clickable { text-decoration: underline; color: $green; }` would be better than duplicating the props (and have the `<div class="info-message clickable">`)? 🤔

GeorgeChatzigiannis: @danwilliams good catch I tried to do that but it did not work for some weird reason and I ended up duplicating which indeed is ugly.
I updated it just now and it works like a charm.
Many thanks!

admin/services/data-access/pupil-access-arrangements.data.service.js
```diff
@@ -0,0 +1,110 @@
...
+      accessArrangements_id,
+      questionReaderReasons_id,
+      inputAssistanceInformation,
+      questionReaderOtherInformation
+      ) VALUES`
+
+  queries.push([insertSql, inserts.join(', \n')].join(' '))
+  let sql
+  if (queries.length > 1) {
+    sql = queries.join('\n')
```
jon-shipley: `queries.join("\n")` is suitable for 1 or more queries, so I don't see any need for special handling of the singular case.

PR #793, opened 2018-08-19
Feature/22557 access arrangements list overview

GeorgeChatzigiannis:
screen shot 2018-08-22 at 9 13 07 pm


admin/data/sql/migrations/113.do.add-access-arrangements-id-field.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE [mtc_admin].pupilAccessArrangements ADD accessArrangements_id int NOT NULL
+ALTER TABLE [mtc_admin].pupilAccessArrangements
+ADD CONSTRAINT FK_pupilAccessArrangements__accessArrangements_id
```
jon-shipley: Is this a double-underscore - in which case is this a typo?

GeorgeChatzigiannis: @jon-shipley yes very good catch. Webstorm generated it this way. I am updating this.

GeorgeChatzigiannis: @jon-shipley unfortunately this is now merged since it was part of #787.
I'll generate an additional migration which renames this one.

admin/views/access-arrangements/overview.ejs
```diff
@@ -55,7 +55,45 @@
...
+                        </div>
+                    </div>
+                </td>
+                <td class="pupil-access-arrangements">
+                    <% pupil.arrangements && pupil.arrangements.forEach((pupilAccessArrangement, i) => { %>
+                        <%if (pupilAccessArrangement.length - 1 !== i && pupilAccessArrangement.length > 1) { %>
+                            <span><%= pupilAccessArrangement %></span><br />
+                        <% } else { %>
+                            <span><%= pupilAccessArrangement %></span>
+                        <% } %>
```
danwilliams: I think this will be better if it’s just a `<div><%= pupilAccessArrangement %></div>` instead of the conditional, since the div has `display: block` (so, line breaks) by default — also, I think you meant to use `pupil.arrangements` instead of `pupilAccessArrangement` in the if statement? (since now it compares the access arrangement length instead of the list length)

Other than that, I think it looks okay 👍

PR #795, opened 2018-08-21
Feature/22546 edit access arrangements

GeorgeChatzigiannis:
screen shot 2018-08-22 at 2 57 32 pm


GuyHarwood:
@GeorgeChatzigiannis this looks exactly the same as PR #793

admin/services/data-access/pupil-access-arrangements.data.service.js
```diff
@@ -129,4 +129,32 @@ pupilAccessArrangementsDataService.sqFindPupilsWithAccessArrangements = async (d
...
+    p.foreName,
+    p.lastName,
+    paa.inputAssistanceInformation,
+    paa.questionReaderOtherInformation,
+    aa.code as accessArrangementCode,
+    qrr.code as questionReaderReasonCode 
+    FROM ${sqlService.adminSchema}.pupilAccessArrangements paa
+    INNER JOIN ${sqlService.adminSchema}.pupil p
+      ON p.id = paa.pupil_id
+    INNER JOIN ${sqlService.adminSchema}.accessArrangements aa
```
GuyHarwood: is there always going to be a matching record between `pupilAccessArrangements` and `accessArrangements`?

if not, an inner join will omit records without a match

admin/services/pupil-access-arrangements-edit.service.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Get pupil access arrangements edit data
+ * @returns {Promise<Array>}
+ */
+pupilAccessArrangementsEditService.getEditData = async (requestData, pupilUrlSlug, dfeNumber) => {
+  if (!pupilUrlSlug) {
+    throw new Error('Pupil url slug not provided')
+  }
+  const reqBody = R.clone(requestData)
```
GuyHarwood: are you passing in a raw http request?  if so, i would consider parsing that in the controller and declare more explicit parameters on the service method

admin/services/data-access/pupil-access-arrangements.data.service.js
GeorgeChatzigiannis: There should always be at least 1 matching record of `accessArrangements` for  `pupilAccessArrangements ` since this is being called from the edit controller method.

admin/services/pupil-access-arrangements-edit.service.js
GeorgeChatzigiannis: @GuyHarwood updated so on both submission and editing controller areas an explicit object is being passed down to the corresponding services.

admin/services/data-access/pupil-access-arrangements.data.service.js
GuyHarwood: 👍

admin/services/pupil-access-arrangements.service.js
```diff
@@ -30,4 +30,28 @@ pupilAccessArrangementsService.getPupils = async (dfeNumber) => {
...
+    throw new Error('Pupil url slug is not provided')
+  }
+  const pupilAccessArrangementsData = await pupilAccessArrangementsDataService.sqlFindAccessArrangementsByUrlSlug(urlSlug)
+  const formData = pupilAccessArrangementsData.reduce((acc, item) => {
+    acc.pupilUrlSlug = !acc.pupilUrlSlug ? item.urlSlug : acc.pupilUrlSlug
+    acc.foreName = !acc.foreName ? item.foreName : acc.foreName
+    acc.lastName = !acc.lastName ? item.lastName : acc.lastName
+    acc.inputAssistanceInformation = !acc.inputAssistanceInformation ? item.inputAssistanceInformation : acc.inputAssistanceInformation
+    acc.questionReaderOtherInformation = !acc.questionReaderOtherInformation ? item.questionReaderOtherInformation : acc.questionReaderOtherInformation
+    acc.questionReaderReason = !acc.questionReaderReason ? item.questionReaderReasonCode : acc.questionReaderReason
```
danwilliams: Can these be something like `acc.pupilUrlSlug = acc.pupilUrlSlug || item.urlSlug` instead, if I got that right? i.e if `acc.pupilUrlSlug` is undefined, it will be `item.urlSlug`, and if that is undefined too, the final result is undefined (although I don’t think that will happen?) 🤔

GeorgeChatzigiannis: @danwilliams valid point the ternary is completely unnecessary.
It's updated now and yes it can't be go into undefined.

PR #796, opened 2018-08-21
Feature/22398 pupil api lookup table storage

jon-shipley:
Add:
* `pupil-api` authorisation against an Azure Table Storage
* extra tokens in the payload for the SPA to communicate back to the queues


pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -0,0 +1,92 @@
...
+            })
+          }
+        })
+      }
+      tableService = azureTableService
+    }
+
+    const { result } = await tableService.retrieveEntityAsync(authTable, schoolPin, pupilPin)
+    if (result.isCollected._ === true) {
+      throw new Error('Repeat attempts to collect data are not allowed')
```
jon-shipley: If we bail out here the pupil will only be able to authenticate once and once only.
@GuyHarwood - thoughts?

GuyHarwood: good point.  we need a retry mechanism that is capped.  i will add a card. needs discussion with BAs/PO

functions/prepare-check/function.json
```diff
@@ -9,7 +9,7 @@
       "name": "queueMessage",
       "type": "queueTrigger",
       "direction": "in",
-      "queueName": "prepare-check",
+      "queueName": "jon-prepare-check",
```
GuyHarwood: this can be standardised back to `prepare-check` now we have everything in place

pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // Prepare the pupil data for use by the SPA
+    const data = this.preparePupilData(result)
+
+    // Mark the data as having been collected
+    try {
+      await this.markAsCollected(tableService, clone(result))
```
GuyHarwood: i imagine the difference in performance could be significant if we don't make this call.

We could move away from this by invalidating/deleting the record when processing the `check-started` queue message

jon-shipley: Possibly there would be a slight difference.  I didn't see any difference when running a single sample of each, on the contrary, the one without the update took exactly the same time.  Looks like it would be worth running a performance test to see if the update call if a hotspot.

pupil-api/src/config.ts
```diff
@@ -9,12 +9,11 @@ const getEnvironment = () => {
 
 export = {
   PORT: process.env.PORT || '3003',
-  // autoMark true | false - Automatically mark the check data when we receive it: boolean
-  autoMark: process.env.hasOwnProperty('AUTO_MARK') ? toBool(process.env.AUTO_MARK) : true,
   Logging: {
     Express: {
-      UseWinston: process.env.EXPRESS_LOGGING_WINSTON || false
+      UseWinston: process.env.hasOwnProperty('EXPRESS_LOGGING_WINSTON') ? toBool(process.env.EXPRESS_LOGGING_WINSTON) : false
```
GeorgeChatzigiannis: I think simplifying this ternary to: `process.env.hasOwnProperty('EXPRESS_LOGGING_WINSTON') && toBool(process.env.EXPRESS_LOGGING_WINSTON)`

pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -0,0 +1,92 @@
...
+      school: JSON.parse(result.school['_']),
+      config: JSON.parse(result.config['_']),
+      tokens: JSON.parse(result.tokens['_'])
+    }
+    return pupilData
+  },
+
+  markAsCollected: async function (tableService, update): Promise<void> {
+    const entGen = azureStorage.TableUtilities.entityGenerator
+    const now = entGen.DateTime(new Date()) // this works, column gets DateTime type
```
GeorgeChatzigiannis: I am concerned of inconsistency of data between those who are stored using the entGen.Datetime and the ones that are still based on moment.
Ultimately it should not be different but would that be worth investigating since we are making this change here?

pupil-api/src/config.ts
jon-shipley: Yes, that's okay if the default is to be false, but how would you default to true in the future?  I prefer leaving it as is, as it allows the default setting to be changed more easily.

GeorgeChatzigiannis: If the default value is to be changed in the future then yes leave it as it is.

pupil-api/src/App.ts
```diff
@@ -16,8 +16,9 @@ const config = require('./config')
 import authRoutes from './routes/auth'
 import pingRoute from './routes/ping'
 
+let logLevel = 'debug'
 if (process.env.NODE_ENV !== 'production') {
-  winston.level = 'debug'
+  logLevel = 'error'
```
danwilliams: Is this not the other way around? If `NODE_ENV` is production, `logLevel` would be error, and debug if in dev?

pupil-api/src/controllers/auth.controller.ts
```diff
@@ -1,9 +1,27 @@
...
+import * as winston from 'winston'
+
+import * as apiResponse from './api-response'
+import { pupilAuthenticationService } from '../services/pupil-authentication.service'
 
 class AuthController {
   async postAuth (req: Request, res: Response) {
-    return res.status(501).send('Not yet implemented')
+    const contentType = req.get('Content-Type')
+    if (!(contentType === 'application/json' || contentType === 'application/json; charset=utf-8')) {
```
danwilliams: Maybe you can use `req.is(‘application/json’)` which catches both (including other charsets, although that most likely won’t happen)?

pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -0,0 +1,89 @@
...
+                  resolve({ result, response })
+                })
+              } catch (error) {
+                reject(error)
+              }
+            })
+          }
+        })
+      }
+      tableService = azureTableService
```
danwilliams: This might look nicer/readable in a separate `initTableService()` function or similar?

pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    // Prepare the pupil data for use by the SPA
+    const data = this.preparePupilData(result)
+
+    // Mark the data as having been collected
+    try {
+      await this.markAsCollected(tableService, clone(result))
+    } catch (error) {
+      winston.error(error.message)
+      throw error
```
danwilliams: A quick note for `markAsCollected`, there is no other way for this to mark the result as corrected and not return the data (which would lead to login failing but being marked as collected and the pupil won’t be able to login anymore), right? Seems like that to me 👍 or we might need to retry somehow?

GuyHarwood: this piece of code may well be removed Dan and passed off to a queue message and corresponding function bound to the queue.  I would prefer the login API to just serve data, nothing more.  However, this was part of the original design so completely understand why Jon has put it in.  I will add the retry policy to the upcoming card that has the refinements in it.

jon-shipley: This behaviour has now been relaxed - repeat logins are now allowed.

PR #799, opened 2018-08-23
Feature/21908 spa configurable numpad removal

danwilliams:


GuyHarwood:
just so everyone is aware - travis has not built this as its target branch is an unprotected branch

admin/services/config.service.js
```diff
@@ -59,9 +60,10 @@ const configService = {
       winston.error('Failed to get access arrangements: ' + error.message)
     }
 
-    if (accessArrangementsCodes.includes('ATA')) {
-      checkOptions.audibleSounds = true
-    }
+    accessArrangementsCodes.forEach(code => {
+      if (code === 'ATA') checkOptions.audibleSounds = true
```
jon-shipley: I'd prefer not testing against magic strings.   I think testing against a constant makes more sense, at least I would know what it meant.

danwilliams: Definitely agree - updated to use what George has done recently now that it's in master.

admin/services/data-access/access-arrangements.data.service.js
```diff
@@ -5,6 +5,19 @@ const sqlService = require('./sql.service')
 const accessArrangementsDataService = {}
 const accessArrangementCodes = {}
 
+/**
+ * Constants to describe the access arrangements codes
+ */
+accessArrangementsDataService.CODES = Object.freeze({
```
GeorgeChatzigiannis: That's indeed very useful.
@jon-shipley @danwilliams  we should follow this pattern in the other areas as well.

PR #804, opened 2018-08-30
Feature/22856 pupil logs in via dedicated pupil api

jon-shipley:
Pupil SPA - pupil authenticates using pupil-auth microservice API

* Hooks up High-performance architecture
* Hidden by feature toggle which defaults to off - controlled by the environment variable `FEATURE_USE_HPA`
* Authentication *only* - future PRs will address check started, check completed and feedback end-points

NB - HPA returned auth payload is slightly different to the old payload as it has many more tokens.


pupil-spa/src/app/services/user/user.service.ts
```diff
@@ -14,25 +14,19 @@ export class UserService {
...
-        { schoolPin, pupilPin },
-        requestArgs)
+      await this.http.post(`${APP_CONFIG.authURL}`, { schoolPin, pupilPin }, headers)
         .toPromise()
-        .then((response) => {
-          if (response.status !== 200) {
-            return reject(new Error('Login Error:' + response.status + ':' + response.statusText));
-          }
-          const data = response.json();
+        .then(data => {
```
GeorgeChatzigiannis: how does the app handle rejection with this change?
Was that block unnecessary in the first place?

pupil-spa/src/app/services/user/user.service.spec.ts
```diff
@@ -1,87 +1,72 @@
...
+          expect(storageService.setItem)
+            .toHaveBeenCalledWith(questionsDataKey, mockLoginResponseBody[questionsDataKey]);
+          expect(storageService.setItem)
+            .toHaveBeenCalledWith(configDataKey, mockLoginResponseBody[configDataKey]);
+        },
         (err) => {
           shouldNotExecute();
         });
+
+      const req = httpTestingController.expectOne(`${APP_CONFIG.authURL}`);
```
GeorgeChatzigiannis: What's the reason behind adding the variable in a string template?
does it not work with just `expectOne(APP_CONFIG.authURL)` ?

pupil-spa/src/app/services/user/user.service.ts
jon-shipley: StatusCode is no longer provided when using httpClient - so it a 401 / 500 will be rejected.

pupil-spa/src/app/services/user/user.service.spec.ts
jon-shipley: That was the existing pattern so I kept it consistent.  I suppose the property could be set to a non-string value?

pupil-spa/src/app/services/user/user.service.ts
GeorgeChatzigiannis: I see, will the rejection be handled then fine?
What will the user view in the frontend when it occurs?

pupil-spa/src/app/services/user/user.service.spec.ts
GeorgeChatzigiannis: I would expect not but it is indeed safeguard if that scenario can potentially happen.

pupil-spa/gen_config.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash
-apiUrl=${API_URL:-"http://localhost:3001"}
-authUrl=${AUTH_URL:-"http://localhost:3001/api/questions"}
-checkStartedUrl=${CHECK_STARTED_URL:-"http://localhost:3001/api/check-started"}
-checkSubmissionUrl=${CHECK_SUBMISSION_URL:-"http://localhost:3001/api/completed-check"}
+featureUseHpa=${FEATURE_USE_HPA:-"false"}
+if [[ "$featureUseHpa" == "true" ]]; then
```
GeorgeChatzigiannis: @jon-shipley if `featureUseHpa` is true then aside from `authUrl` remain undefined.
Is that intentional?
I would assume this would break the app at this stage right on check start.

PR #807, opened 2018-08-30
Feature/23823 ict survey landing page

danwilliams:


danwilliams:
For the footer links, we currently use the `API_URL` to point to the admin version of the pages (privacy, cookies, contact). Will we need to have a version of them in the SPA as well?

jon-shipley:
@danwilliams - sounds like a question for the PO

jon-shipley:
@danwilliams  - I am surprised the PR is going into master. I would have thought this would be best as it's own support branch, as seems like a standalone project to me.

@GuyHarwood - thoughts?

pupil-spa/src/app/survey-welcome/survey-welcome.component.html
```diff
@@ -0,0 +1,58 @@
...
+        </p>
+      </div>
+      <p>
+        Following the connection test, please give us feedback on ICT issues you may have come across. 
+      </p>
+      <p> 
+        You will be able to preview the multiplication tables check system after the connections test. 
+      </p>
+      <details>
+        <summary (click)="toggleDetails()"><span class="summary">{{ showDetails ? 'Less details' : 'More details' }}</span></summary>
```
jon-shipley: Surely this should be "Fewer details"?    Details can be counted accurately.  https://en.oxforddictionaries.com/usage/less-or-fewer

If we want to stick with "less", then perhaps "Less Detail" and "More detail" make sense.

SineadFluent: In the Axure, we have 'More detail' and 'Less detail'. Some places in gov.uk design system they don't change the text on the link https://design-system.service.gov.uk/components/details/ but in this case, it makes more sense to describe what the user will be doing by clicking on that link. I've raised it in the design systems github to see if anyone has opinions/research about this. For now let's keep it as 'More detail'/ 'Less detail'

PR #809, opened 2018-09-04
Feature/24014 check windows hub page

GeorgeChatzigiannis:
New check window hub page and view changes on service manager home page.
Feature toggle that activates the routing for the new check window functionality has been implemented until the feature is complete.


admin/views/service-manager/service-manager-home.ejs
```diff
@@ -11,13 +11,17 @@
         <%- partial('partials/_session_expiry') %>
 
         <ol class="list">
+            <!--<li>-->
```
jon-shipley: Is this being left in for a reason?

GeorgeChatzigiannis: Actually yes. I want to keep the existing functionality until the new check window changes are fully implemented.
I could uncomment the existing link and keep both until the 2nd one is done.
@GuyHarwood  @jon-shipley what would be the best course of action?

jon-shipley: That's fine, as long as there is a reason.

GuyHarwood: @GeorgeChatzigiannis as discussed - please use a feature toggle that defaults to the old implementation until such time that the new one is usable.  Ensure the new version is accessible on via direct URL so it can be viewed if necessary.

GeorgeChatzigiannis: @GuyHarwood @jon-shipley updated with feature toggle as discussed

PR #810, opened 2018-09-04
Feature/23412 check started function persist to pupilevent table storage

jon-shipley:
Azure function to process the "Check Started" message:

* updates master SQL DB using `less-tedious` and connection-pooling
* deletes the authentication/question entity from the `preparedCheck` table storage
* logs raw queue message to the queue event log


functions/check-started/index.js
```diff
@@ -1,11 +1,46 @@
...
+    context.log('SUCCESS: Admin DB updated')
+  } catch (error) {
+    context.log.error(`ERROR: unable to update admin db for [${checkStartMessage.checkCode}]`)
+    throw error
+  }
+
+  // Delete the row in the preparedCheck table - prevent pupils logging in again.
+  try {
+    await deleteFromPreparedCheckTableStorage(checkStartMessage.checkCode, context.log)
+    context.log('SUCCESS: Table storage cache deleted from preparedCheck table')
```
GuyHarwood: can we change this message to `SUCCESS: pupil check row deleted from preparedCheck table` please

functions/check-started/index.js
```diff
@@ -1,11 +1,46 @@
...
+  } catch (error) {
+    context.log.error(`ERROR: unable to delete from table storage for [${checkStartMessage.checkCode}]`)
+    throw error
+  }
+
+  // Store the raw message to an audit log
   context.bindings.pupilEventsTable = []
   const entity = {
     PartitionKey: checkStartMessage.checkCode,
```
GuyHarwood: you will need rowKey also i believe, as that is the unique identifier of the row in the partition.  you can just add a UUID, as checkCode is good for partition.

jon-shipley: This code is unchanged, so not in the diff, and it includes the RowKey 😄 :

```
const entity = {
    PartitionKey: checkStartMessage.checkCode,
    RowKey: uuid(),
    eventType: 'check-started',
    payload: JSON.stringify(checkStartMessage),
    processedAt: moment().toDate()
  }
```

functions/check-started/index.js
jon-shipley: Yes, that works for me

jon-shipley:
@GeorgeChatzigiannis 

There isn't a spec file, because there are a lot of side-effects just from 'requiring' the code - It initialises azure table storage and the SQL connection pool in the hope that these will be done once, and not repeated on every call.

functions/check-started/index.js
GuyHarwood: ahhh!

GuyHarwood:
@jon-shipley you are welcome to override the snyk failure, as its a low priority one.

PR #814, opened 2018-09-06
Fix `ng build --prod`

jon-shipley:
Fix an issue that was preventing Docker from building the pupil-spa.   The `config` var is referenced inside HTML, so needs to be public.

Let me know your thoughts.


GuyHarwood:
Passes VSTS build so all good!

PR #817, opened 2018-09-07
remove logdna from admin app

GuyHarwood:
now we are solely using app insights we no longer need the package, or require the initialisation code

### DIFF
--- a/admin/helpers/logger.js
+++ b/admin/helpers/logger.js
@@ -8,20 +8,6 @@ let initialised
 
 const init = (app) => {
   if (initialised) return
-  /**
- * Logging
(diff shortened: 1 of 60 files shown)


admin/helpers/logger.js
```diff
@@ -12,7 +12,7 @@ const init = (app) => {
  * Logging
  * use LogDNA transport for winston if configuration setting available
  */
-  if (config.Logging.LogDna.key) {
+/*   if (config.Logging.LogDna.key) {
```
jon-shipley: Why not remove it.

GuyHarwood: it should have been.  that was temporary while i checked for other references.

removing...

PR #819, opened 2018-09-12
Feature/23411 complete check function persist to pupilevent table storage

jon-shipley:
New complete check function to handle messages on the `check-complete`  queue put there by the Pupil SPA.

* sends the payload to the Admin DB
* updates the check status
* stores a copy of the message in `pupilEvents`
* deletes from the `preparedCheck` if not already deleted


admin/services/check-start.service.js
```diff
@@ -236,10 +236,12 @@ checkStartService.prepareCheckQueueMessages = async function (checkFormAllocatio
       schoolPin: o.school_pin,
       pupilPin: o.pupil_pin,
       pupil: {
+        id: o.pupil_id,
```
GuyHarwood: nice. i will remove this from the other story it was added to.

functions/lib/lib.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
+
```
GuyHarwood: can we call this file something more meaningful like `azure-storage-helper.js`?

in `index.js` above it is not clear what the intention is from requiring `lib/lib`

functions/completed-checks/index.js
```diff
@@ -1,11 +1,60 @@
...
+// SQL server table
+const checkResultTable = '[checkResult]'
+const checkFormAllocationTable = '[checkFormAllocation]'
+const schema = ['mtc_admin']
+const checkStatusTable = '[checkStatus]'
+const checkTable = '[checkFormAllocation]'
 
-module.exports = function (context, completedCheckMessage) {
+// Table Storage
+const azureTableService = getPromisifiedAzureTableService()
```
GeorgeChatzigiannis: @jon-shipley I still think promisifying the whole library for 2 method calls along with any potential internal dependency in other methods is too much.
The time it takes for the library 'promisification' to complete might be small but I believe it's unnecessary for the minimal usage you have.

functions/lib/lib.js
jon-shipley: Seems fair enough.  👍

functions/lib/lib.js
```diff
@@ -0,0 +1,79 @@
...
+    logger.error(error.message)
+    throw error
+  }
+}
+
+/**
+ * Promisify and cache the azureTableService library as it still lacks Promise support
+ */
+let azureTableService
+module.exports.getPromisifiedAzureTableService = function getPromisifiedAzureTableService () {
```
GeorgeChatzigiannis: Perhaps here we could restrict to the methods that we want the loop to go through with regards to promisifying?

functions/completed-checks/index.js
GeorgeChatzigiannis: I am resolving this based on the answer given by Guy on #821

PR #820, opened 2018-09-12
Feature/23826 connection test

danwilliams:


danwilliams:
For the processing speed, I'm doing fibonacci(N) x `fibonacciInteractions` times and checking how long it took to compute it, where:

```private fibonacciN = 10000;
private fibonacciIterations = 50000;```

For the bandwidth check, I'm currently requesting a large file (at the moment, the largest image in the pupil-spa, but the link is configurable through an env variable) and doing:

```this.connectionSpeed = ((requestSize * 8) / ((endTime - startTime) / 1000)) / 1024;```

...and for the device data, I'm using the device service that was already implemented before for the pupil check and gets the window / navigator / screen properties that we want.

Also, for testing, I’ve created a `connection-tests` queue, created an ADD only token, and used the env variables to set them, as in:

```TEST_SAS_QUEUE='connection-tests' TEST_SAS_URL='url' TEST_SAS_TOKEN='token' npm start```

**Comment by GuyHarwood** [comment]
@danwilliams Jon solved the speed test issue by generating a set of files of various sizes that were available from cloud storage.  Feels like we are re-solving the same problem.

Have you seen the source code from last years survey?

File: pupil-spa/src/app/services/connection-test/connection-test.service.ts [hunk 1]
```diff
@@ -0,0 +1,128 @@
...
+  constructor(private http: HttpClient,
+              private router: Router,
+              private storageService: StorageService,
+              private deviceService: DeviceService) {}
+
+  async startTest() {
+    await Promise.all([
+      this.benchmarkProcessing(),
+      this.benchmarkConnection()
+    ]);
```
jon-shipley: 👍 - nice to see it all happening at once

pupil-spa/src/app/services/connection-test/connection-test.service.ts
```diff
@@ -0,0 +1,128 @@
...
+        navigator: this.deviceService.getNavigatorProperties(),
+        networkConnection: this.deviceService.getNetworkInformation(),
+        screen: this.deviceService.getScreenProperties(),
+      },
+      processingTime: this.processingTime,
+      connectionSpeed: this.connectionSpeed,
+    };
+  }
+
+  public benchmarkProcessing(): Promise<void> {
```
jon-shipley: How long do you expect this to take?

pupil-spa/src/app/services/connection-test/connection-test.service.ts
```diff
@@ -0,0 +1,128 @@
...
+      }
+
+      this.processingTime = Date.now() - startTime;
+      resolve();
+    });
+  }
+
+  public benchmarkConnection(): Promise<void> {
+    return new Promise(resolve => {
+      const connectionTestFile = APP_CONFIG.connectionTestFile;
```
jon-shipley: This might be fit for purpose, however, I have some questions that might be best asked offline.

danwilliams:
@GuyHarwood I believe the principle is the same... I have left it open as to what file could be used, and indeed we could use more than one. I've not seen the code from last year - I understood from conversation that it was not kept? I think what is here is a good starting point - there are some things that could make it more sophisticated, but I wasn't sure how far to take it.

pupil-spa/src/app/connection-test/connection-test.component.ts
```diff
@@ -1,12 +1,25 @@
-import { Component } from '@angular/core';
+import { Component, OnInit} from '@angular/core';
+import { StorageService } from '../services/storage/storage.service';
+import { ConnectionTestService } from '../services/connection-test/connection-test.service';
+import { WindowRefService } from '../services/window-ref/window-ref.service';
```
GeorgeChatzigiannis: @danwilliams is that windowrefservice reference being used?

GeorgeChatzigiannis:
@danwilliams some conflicts here

pupil-spa/src/app/services/connection-test/connection-test.service.ts
```diff
@@ -0,0 +1,174 @@
...
+        resolve();
+      });
+    });
+  }
+
+  private getQueueService() {
+    return AzureStorage.Queue.createQueueServiceWithSas(
+      APP_CONFIG.testSasUrl,
+      APP_CONFIG.testSasToken
+    ).withFilter(new AzureStorage.Queue.ExponentialRetryPolicyFilter());
```
GeorgeChatzigiannis: @danwilliams is there a reason you chose exponential instead of linear policy filter?

danwilliams: Mmmm, it just seemed best to me... I guess it's somewhat arbitrary?

GuyHarwood: @danwilliams @GeorgeChatzigiannis https://azure.github.io/azure-storage-node/common_filters_exponentialretrypolicyfilter.js.html

pupil-spa/src/app/services/connection-test/connection-test.service.ts
jon-shipley: Seems reasonable

GuyHarwood:
fix works for docker build issues.

signing off.

PR #821, opened 2018-09-12
feature/24184 - skeleton pupil status function

GuyHarwood:
Skeleton function and queue for pupil status processing.


functions/pupil-status/index.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Promisify the azureStorage library as it still lacks Promise support
+ */
+function initAzureTableService () {
+  if (!azureTableService) {
+    azureTableService = azureStorage.createTableService()
+    bluebird.promisifyAll(azureTableService, {
+      promisifier: (originalFunction) => function (...args) {
```
GeorgeChatzigiannis: Similar comment I made to Jon's PR, why not promisifying what is required for the implementation?
The time needed for the whole 'promisification' might be small but the library consists of many methods that we are promisifying for no reason as they will not be used.

GuyHarwood: this happens once when the functions app (containing ALL of the functions) is loaded.  
it makes sense to have one piece of code that promisifies the entire library, rather than boilerplate for each method.
However - could we extract this out to a file in the functions root folder, and then it is loaded once for ALL functions instead of in each function.  This would be similar to how the sql connection pooling works i believe.  @jon-shipley thoughts?

jon-shipley: Do you mean this?
https://github.com/DFEAGILEDEVOPS/MTC/pull/819/files#diff-e43640ff0b6d9faccb1709042885f09c

GuyHarwood: thats what i was thinking, perfect 👍  
@GeorgeChatzigiannis i will implement this now

GeorgeChatzigiannis: 👍

GuyHarwood:
waiting on #819 so i can reference the azure storage helper

PR #825, opened 2018-09-18
Bug/24021 Detect key presses only when modal is visible

GeorgeChatzigiannis:


admin/assets/javascripts/jquery-modal.js
```diff
@@ -17,7 +17,7 @@ $(function () {
     $('#js-modal-overlay').on('click', function (e) {
       toggleShowHideModal(e)
     })
-    $('body').on('keydown', function (e) {
+    $('#modal-wrapper').on('keydown', function (e) {
```
jon-shipley: I think that placing the `keydown` event handler on the `#modal-wrapper`  means that if another element outside of `#modal-wrapper` has the focus when the key is pressed then the `#modal-wrapper` will not receive the event.

jon-shipley: Okay - I can see from the code that this does not apply to the modal dialog as it is removed from the screen when the focus is lost.

jon-shipley: 👍

PR #826, opened 2018-09-18
Feature/23415 feedback function persist to pupil event table storage

jon-shipley:
* Saves the pupil feedback message to the Admin DB - with one lookup to get the FK
* New FK `checkFormAllocation_id` added to `pupilFeedback` table that references the `checkFormAllocation` table rather than the `check` table which is deprecated.  This allows feedback to be provided by both old and new (HPA) systems.


functions/feedback/index.js
```diff
@@ -1,19 +1,75 @@
...
+const sqlService = require('less-tedious')
+const { TYPES } = require('tedious')
+
+const sqlUtil = require('../lib/sql-helper')
+const schema = '[mtc_admin]'
+const feedbackTable = '[pupilFeedback]'
+
+module.exports = async function (context, feedbackMessage) {
+  if (typeof feedbackMessage !== 'object') {
+    throw new Error('Badly formed message')
```
GuyHarwood: 🎖

functions/feedback/sample.data.json
```diff
@@ -0,0 +1,6 @@
+{
```
GuyHarwood: this is a good idea 💯 
we should do this across all of them

GuyHarwood: card added to board for above.

functions/lib/sql-helper.js
```diff
@@ -0,0 +1,26 @@
...
+const sqlService = require('less-tedious')
+
+const checkFormAllocationTable = '[checkFormAllocation]'
+const { TYPES } = require('tedious')
+const schema = '[mtc_admin]'
+
+/**
+ * Retrieve the checkFormAllocation data from the db
+ * @param checkCode
+ * @return {Promise<void>}
```
GeorgeChatzigiannis: @jon-shipley JSdoc return type is wrong.
You are returning an object(R.head) as far as I see.

PR #833, opened 2018-09-21
Feature/22166 Submit feedback to queue

GeorgeChatzigiannis:
- Include config toggle variable to perform either call to queue or previous API post call to submit the feedback
- Introduce feedback api max attempts and delay variables that are used in the retry config

### DIFF
--- a/pupil-spa/src/app/services/feedback/feedback.service.ts
+++ b/pupil-spa/src/app/services/feedback/feedback.service.ts
@@ -63,11 +63,7 @@ export class FeedbackService {
       errorDelay: this.feedbackAPIErrorDelay,
       errorMaxAttempts: this.feedbackAPIErrorMaxAttempts
     };
-    try {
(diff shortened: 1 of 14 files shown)


pupil-spa/src/app/services/feedback/feedback.service.ts
```diff
@@ -2,44 +2,91 @@ import { Injectable } from '@angular/core';
...
+    const { url, token } = this.tokenService.getToken('pupilFeedback');
+    // Create a model for the payload
+    const retryConfig = {
+      errorDelay: this.feedbackAPIErrorDelay,
+      errorMaxAttempts: this.feedbackAPIErrorMaxAttempts
+    };
+    try {
+      await this.azureQueueService.addMessage(queueName, url, token, payload, retryConfig);
+    } catch (error) {
+      return new Error(error);
```
jon-shipley: Why catch the original error and throw a new one, and lose the stack trace?

PR #837, opened 2018-09-21
removing empty related section from privacy pagew

activemq:


pupil-spa/src/app/privacy/privacy.component.html
```diff
@@ -175,10 +175,7 @@ <h2 class="heading-small heading-contents" id="info">Further information</h2>
       </p>
     </div>
 
-    <div class="column-one-third related">
```
jon-shipley: If you remove the entire column then the LHS will spill over into this column.  Suggest you leave it in, just remove the header.

activemq: Ok sounds good

activemq:
@jon-shipley Can you re-review please

PR #838, opened 2018-09-24
Feature/24340 sample queue messages

GuyHarwood:
Documentation on queues / messaging and its processing.


functions/completed-checks/sample.data.json
```diff
@@ -0,0 +1,75 @@
+{
```
GeorgeChatzigiannis: @GuyHarwood the following data are not accessible via local storage: `schoolPin `, `pupilPin `, `firstName `, `lastName `, `dob `
Also it seems `audit`, `answers`, `inputs` are missing

GuyHarwood: good catch.  pupil details removed, will add missing data.

PR #840, opened 2018-09-25
Bug/24469 view details not working in safari

danwilliams:


GeorgeChatzigiannis:
@danwilliams Assuming all browsers are compatible with this, approved.

danwilliams:
@GeorgeChatzigiannis The polyfill I added is an updated version of the original one - that had problems with the click events; this one works directly, without ‘changing’ the code as had to be done for the previous one.

The polyfill checks if the browser supports the details element; if not (which includes IE/Edge), it inserts the needed CSS, and adds code with the functionality as in the original spec introduced by html5.

PR #845, opened 2018-09-27
Feature/24638 pin gen v2 data service

GuyHarwood:
Data service for 2nd gen pin service


admin/data/sql/migrations/123.do.pupil-eligible-for-pin-gen.view.sql
```diff
@@ -0,0 +1,29 @@
+CREATE VIEW [mtc_admin].vewPupilsEligibleForPinGeneration AS
+SELECT
+ p.id,
+ p.forename,
+ p.lastName,
+ p.dateOfBirth,
+ p.urlSlug,
+ p.school_id,
+ COUNT(chk.id) AS checkCount,
+ IIF(COUNT(chk.id) < 3, 1, 0) AS isRestartAllowed
```
GeorgeChatzigiannis: @GuyHarwood non blocker: Would it be possible to declare 3 as a variable to make this more readable cause otherwise it seems like a magic number.

PR #848, opened 2018-09-28
Fix migrations

jon-shipley:
Edit migration 118

Now takes into account existing data.  Tested on a local DB dump of the dev environment.


GuyHarwood:
@jon-shipley Quite a lot of specs are failing, some of them surrounding the restarts feature.

Is this due to the included change?

PR #850, opened 2018-10-03
Feature/24376 Prevent submission in practice mode

GeorgeChatzigiannis:
- Prevent check submission on pupil spa when practice toggle (config) is enabled
- Prevent check complete sas token generation if the batch of checks generated are non-live (familiarisation)

### DIFF
--- a/admin/services/check-start.service.js
+++ b/admin/services/check-start.service.js
@@ -279,7 +279,7 @@ checkStartService.prepareCheckQueueMessages = async function (checkIds) {
       questions: checkFormService.prepareQuestionData(JSON.parse(o.checkForm_formData)),
       config: config
     }
-    if (hasLiveChecks) {
+    if (o.check_isLiveCheck) {
       message.tokens.checkComplete = {
         token: checkCompleteSasToken.token,
         url: checkCompleteSasToken.url
--- a/admin/services/check-start.service.js
+++ b/admin/services/check-start.service.js
@@ -211,7 +211,7 @@ checkStartService.pupilLogin = async function (pupilId) {
 /**
  * Query the DB and put the info into messages suitable for placing on the prepare-check queue
  * The message needs to contain everything the pupil needs to login and take the check
- * @param CheckIds
+ * @param checkIds
  * @return {Promise<Array>}
  */
 checkStartService.prepareCheckQueueMessages = async function (checkIds) {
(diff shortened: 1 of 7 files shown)


pupil-spa/src/app/services/check-complete/check-complete.service.spec.ts
```diff
@@ -22,6 +22,13 @@ let submissionService: SubmissionService;
 let tokenService: TokenService;
 
 describe('CheckCompleteService', () => {
+  const getItemMock = (arg, isPractice) => {
+    if (arg === 'config') {
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @GuyHarwood this in case the check complete service includes in the future different `storageService` `getItem` calls.
For the particular one we only want to return a fake config object if the caller's argument was `config`.

admin/services/check-start.service.js
```diff
@@ -227,9 +227,14 @@ checkStartService.prepareCheckQueueMessages = async function (checkIds) {
   const checks = await checkFormAllocationDataService.sqlFindByIdsHydrated(checkIds)
   const sasExpiryDate = moment().add(config.Tokens.sasTimeOutHours, 'hours')
 
+  const hasLiveChecks = R.all(c => R.equals(c.check_isLiveCheck, true))(checks)
```
GuyHarwood: my only concern here would be if there is a situation where there could be a mix of live/practice checks?  although my understanding is that a batch will always be all practice or all live?

GeorgeChatzigiannis: @GuyHarwood as far as I see this is being called when familiarisation pins are generated and when live pins are generated.
It appears that with the current implementation we will never have a mixture of live/practice checks at that level.

pupil-spa/src/app/services/check-complete/check-complete.service.spec.ts
```diff
@@ -67,24 +75,34 @@ describe('CheckCompleteService', () => {
...
       expect(addEntrySpy.calls.all()[0].args[0].type).toEqual('CheckSubmissionApiCalled');
       expect(submissionService.submitData).toHaveBeenCalledTimes(1);
     });
+    it('submit should return if the app is configured to run in practice mode', async () => {
+      const addEntrySpy = spyOn(auditService, 'addEntry');
+      spyOn(storageService, 'getItem').and.callFake(arg => getItemMock(arg, true));
+      spyOn(submissionService, 'submitData');
+      await checkCompleteService.submit(Date.now());
+      expect(addEntrySpy).toHaveBeenCalledTimes(0);
+      expect(submissionService.submitData).toHaveBeenCalledTimes(0);
```
danwilliams: I'm just trying to decide whether `.toHaveBeenCalledTimes(0);` would look better as `.not.toHaveBeenCalled()` - but that's a pretty minor comment really.

PR #852, opened 2018-10-04
Feature/24490 Preserve preparedCheck entry in practice mode

GeorgeChatzigiannis:
- On pupil app practice config variable is passed to the check-started payload 
- `check-started` function can identify if the check is running on familiarisation and prevent deletion of preparedCheck

### DIFF
--- a/functions/check-started/index.js
+++ b/functions/check-started/index.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const moment = require('moment')
-const R = require('ramda')
 const winston = require('winston')
 const uuid = require('uuid/v4')
 const { TYPES } = require('tedious')
--- a/functions/check-started/index.js
+++ b/functions/check-started/index.js
@@ -32,9 +32,11 @@ module.exports = async function (context, checkStartMessage) {
     throw error
   }
 
-  // Delete the row in the preparedCheck table - prevent pupils logging in again.
+  // Delete the row in the preparedCheck table for live checks only - prevent pupils logging in again.
   try {
-    await deleteFromPreparedCheckTableStorage(azureTableService, checkStartMessage.checkCode, context.log)
+    if (!checkStartMessage.practice) {
+      await deleteFromPreparedCheckTableStorage(azureTableService, checkStartMessage.checkCode, context.log)
+    }
     context.log('SUCCESS: pupil check row deleted from preparedCheck table')
   } catch (error) {
     context.log.error(`ERROR: unable to delete from table storage for [${checkStartMessage.checkCode}]`)
--- a/functions/check-started/index.js
+++ b/functions/check-started/index.js
@@ -11,6 +11,7 @@ const config = require('../config')
 winston.level = 'error'
 sqlService.initialise(config)
 const { deleteFromPreparedCheckTableStorage, getPromisifiedAzureTableService } = require('../lib/azure-storage-helper')
+const sqlUtil = require('../lib/sql-helper')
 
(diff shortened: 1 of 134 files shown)


functions/check-started/index.js
```diff
@@ -32,9 +32,11 @@ module.exports = async function (context, checkStartMessage) {
     throw error
   }
 
-  // Delete the row in the preparedCheck table - prevent pupils logging in again.
+  // Delete the row in the preparedCheck table for live checks only - prevent pupils logging in again.
   try {
```
GuyHarwood: the problem with this is that you are trusting the client payload to be correct.  You should be looking this up in the main database via checkCode to determine how the flag is set.

functions/check-started/index.js
```diff
@@ -32,10 +33,14 @@ module.exports = async function (context, checkStartMessage) {
     throw error
   }
 
-  // Delete the row in the preparedCheck table - prevent pupils logging in again.
+  // Delete the row in the preparedCheck table for live checks only - prevent pupils logging in again.
   try {
-    await deleteFromPreparedCheckTableStorage(azureTableService, checkStartMessage.checkCode, context.log)
-    context.log('SUCCESS: pupil check row deleted from preparedCheck table')
+    const check = await sqlFindCheckByCheckCode(checkStartMessage.checkCode, context.log)
+    context.log('check: ', check)
```
jon-shipley: This check var looks like debug. Perhaps it should be removed?

functions/check-started/index.js
```diff
@@ -32,10 +33,14 @@ module.exports = async function (context, checkStartMessage) {
...
 
-  // Delete the row in the preparedCheck table - prevent pupils logging in again.
+  // Delete the row in the preparedCheck table for live checks only - prevent pupils logging in again.
   try {
-    await deleteFromPreparedCheckTableStorage(azureTableService, checkStartMessage.checkCode, context.log)
-    context.log('SUCCESS: pupil check row deleted from preparedCheck table')
+    const check = await sqlFindCheckByCheckCode(checkStartMessage.checkCode, context.log)
+    context.log('check: ', check)
+    if (check.isLiveCheck) {
+      await deleteFromPreparedCheckTableStorage(azureTableService, checkStartMessage.checkCode, context.log)
```
jon-shipley: 💯

functions/check-started/index.js
```diff
@@ -91,3 +97,26 @@ async function updateAdminDatabaseForCheckStarted (checkCode, startedAt, logger)
...
   }
 }
+
+/**
+ * Fetch check record based on check code
+ * @param {String} checkCode - the unique GUID that identifies the check in the admin DB
+ * @param {Function} logger
+ * @return {Object} check record
+ */
+async function sqlFindCheckByCheckCode (checkCode, logger) {
```
jon-shipley: You don't need to write this function again - just import '../lib/sql-helper' into the function.  There is an example usage like this in completed-check.

functions/check-started/index.js
GuyHarwood: i would agree.  if the object is sizeable then it creates a lot of noise in the logs

GeorgeChatzigiannis: @jon-shipley @GuyHarwood this is removed.

functions/check-started/index.js
GeorgeChatzigiannis: thanks @jon-shipley, didn't spot that helper. I have updated based on that recommendation.

PR #858, opened 2018-10-05
Feature/24788 last beep after webspeech

danwilliams:


danwilliams:
@GeorgeChatzigiannis Good point, yes; I agree 😄 Added!

PR #859, opened 2018-10-05
Feature/24789 spa webspeech leave active state button

danwilliams:


pupil-spa/src/app/services/speech/speech.service.ts
```diff
@@ -107,6 +109,8 @@ export class SpeechService implements OnDestroy {
 
     this.utterancesGC.push(utterance);
     this.synth.speak(sayThis);
+
+    return new Promise((resolve) => resolve());
```
GeorgeChatzigiannis: @danwilliams what is the purpose of returning a Promise which all it does is resolve?
Can you elaborate on that one please?
I would expect this method to either be void or returned on the line above.

danwilliams: I’ve used that to await it in the other components before setting the ‘active’ state on the buttons, but you’re right with it being redundant... as, the function being async will return a promise anyway, so I’ve removed it 😄 👍

PR #866, opened 2018-10-11
Display stored check window list on manage check window screen

GeorgeChatzigiannis:
- Created db view with check window names and statuses
- Updated manage check window UI to display check window list

### DIFF
--- a/admin/data/sql/migrations/143.do.create-check-window-with-status-view.sql
+++ b/admin/data/sql/migrations/143.do.create-check-window-with-status-view.sql
@@ -1,0 +1,9 @@
+CREATE VIEW [mtc_admin].vewCheckWindowWithStatus
+AS SELECT cw.name, CAST(
+    CASE
+        WHEN GETUTCDATE() < cw.checkStartDate THEN 'Inactive'
+        WHEN GETUTCDATE() >= cw.checkStartDate AND GETUTCDATE() <= cw.checkEndDate THEN 'Active'
+        ELSE 'Past'
+    END AS NVARCHAR(50)
+   ) AS [status]
+FROM mtc.mtc_admin.checkWindow cw
--- a/admin/services/check-window-v2.service.js
+++ b/admin/services/check-window-v2.service.js
@@ -1,0 +1,18 @@
+'use strict'
+
+const checkWindowDataService = require('./data-access/check-window.data.service')
+const monitor = require('../helpers/monitor')
+
+const checkWindowV2Service = {}
+
+/**
+ * Get all check windows recorded in the database with their status
+ * @returns {Array} List of check windows
+ */
+checkWindowV2Service.getCheckWindows = async () => {
+  const checkWindows = await checkWindowDataService.sqlFindCheckWindowsWithStatus()
+  checkWindows.forEach(cw => (cw.canRemove = cw.status === 'Inactive'))
+  return checkWindows
+}
+
+module.exports = monitor('check-window-v2.service', checkWindowV2Service)
--- a/admin/services/data-access/check-window.data.service.js
+++ b/admin/services/data-access/check-window.data.service.js
@@ -314,6 +314,15 @@ const checkWindowDataService = {
     ]
     const rows = await sqlService.query(sql, params)
     return R.head(rows)
+  },
+
+  /**
+   * Fetch all check windows with status
+   * @return {Promise<Array>}
+   */
+  sqlFindCheckWindowsWithStatus: async () => {
+    const sql = `SE LECT * FROM ${sqlService.adminSchema}.[vewCheckWindowWithStatus]`
+    return sqlService.query(sql)
   }
 }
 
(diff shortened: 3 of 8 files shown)


admin/data/sql/migrations/143.do.create-check-window-with-status-view.sql
```diff
@@ -0,0 +1,9 @@
+CREATE VIEW [mtc_admin].vewCheckWindowWithStatus
+AS SELECT cw.urlSlug, cw.name, cw.isDeleted, CAST(
+    CASE
+        WHEN GETUTCDATE() < cw.adminStartDate THEN 'Inactive'
+        WHEN GETUTCDATE() >= cw.adminStartDate AND GETUTCDATE() <= cw.adminEndDate THEN 'Active'
+        ELSE 'Past'
+    END AS NVARCHAR(50)
+   ) AS [status]
+FROM mtc.mtc_admin.checkWindow cw
```
jon-shipley: Don't include the database name `mtc` please

admin/services/check-window-v2.service.js
```diff
@@ -0,0 +1,43 @@
...
+  const checkWindow = await checkWindowDataService.sqlFindOneByUrlSlug(urlSlug)
+  if (!checkWindow) {
+    throw new Error('Check window not found')
+  }
+  if (checkWindow.adminEndDate.isBefore(currentUTCTimestamp)) {
+    throw new Error('Deleting an past check window is not permitted')
+  }
+  if (checkWindow.adminStartDate.isSameOrBefore(currentUTCTimestamp) && checkWindow.adminEndDate.isSameOrAfter(currentUTCTimestamp)) {
+    throw new Error('Deleting an active check window is not permitted')
+  } else {
```
jon-shipley: This else is not part of the guard condition.  Consider moving it into its own block for clarity.

admin/services/check-window-v2.service.js
```diff
@@ -0,0 +1,43 @@
...
+    await checkWindowDataService.sqlDeleteCheckWindow(checkWindow.id)
+    return {
```
jon-shipley: I am dubious that this return object is suitable for a service.  It looks a bit like this will passed directly to the view.  The needs of the view of largely irrelevant to the service.

admin/services/data-access/check-window.data.service.js
```diff
@@ -314,6 +314,18 @@ const checkWindowDataService = {
...
     return R.head(rows)
+  },
+
+  /**
+   * Fetch all check windows with status
+   * @return {Promise<Array>}
+   */
+  sqlFindCheckWindowsWithStatus: async () => {
+    const sql = `SELECT *
+    FROM ${sqlService.adminSchema}.[vewCheckWindowWithStatus]
```
jon-shipley: 💯 Nice to see this as a view

admin/services/check-window-v2.service.js
```diff
@@ -0,0 +1,43 @@
...
+  checkWindows.forEach(cw => (cw.canRemove = cw.status === 'Inactive'))
+  return checkWindows
+}
+
+/**
+ * Mark check window as deleted based on url slug
+ * @returns {Object|Error} Either a successful message or throws an exception
+ */
+checkWindowV2Service.markDeleted = async (urlSlug) => {
+  const currentUTCTimestamp = moment.utc()
```
GuyHarwood: feels like there should be validation here to ensure that `urlSlug` is actually a valid UUID.

admin/services/check-window-v2.service.js
GuyHarwood: agreed. just execute the delete and return

admin/data/sql/migrations/143.do.create-check-window-with-status-view.sql
GeorgeChatzigiannis: @jon-shipley good spot, updated.

admin/services/check-window-v2.service.js
GeorgeChatzigiannis: @jon-shipley this has now been modified.

admin/services/check-window-v2.service.js
GeorgeChatzigiannis: @jon-shipley @GuyHarwood it's updated.

admin/services/check-window-v2.service.js
GeorgeChatzigiannis: @GuyHarwood introduced validation but I am guessing we could extend this validation in other areas where we need to validate `urlSlug` fields.

GuyHarwood: i did think that earlier.  it will no doubt need adding to the pupil edit area, and maybe some others.

admin/data/sql/migrations/143.do.create-check-window-with-status-view.sql
```diff
@@ -6,4 +6,4 @@ AS SELECT cw.urlSlug, cw.name, cw.isDeleted, CAST(
         ELSE 'Past'
     END AS NVARCHAR(50)
    ) AS [status]
-FROM mtc.mtc_admin.checkWindow cw
+FROM [mtc_admin].checkWindow cw
```
GuyHarwood: 👍  the database name is different across environments, and should never appear statically in the code

PR #873, opened 2018-10-18
Feature/24062 Edit check window

GeorgeChatzigiannis:
- Edit check window call and UI changes
- Edited check window submission


admin/controllers/check-window.js
```diff
@@ -60,9 +62,15 @@ const controller = {
    */
   submitCheckWindow: async (req, res, next) => {
     const requestData = req.body
+    const { checkWindowUrlSlug } = requestData
     let flashMessage
     try {
-      flashMessage = await checkWindowV2AddService.submit(requestData)
+      if (checkWindowUrlSlug) {
+        flashMessage = await checkWindowV2UpdateService.submit(requestData)
```
jon-shipley: This seems odd to me.  Why is a service returning a message that seems to be targeted at the GUI?

admin/controllers/check-window.js
```diff
@@ -99,6 +107,35 @@ const controller = {
...
+
+  getCheckWindowEditForm: async (req, res, next) => {
+    req.breadcrumbs('Manage check windows', '/check-window/manage-check-windows')
+    res.locals.pageTitle = 'Edit check window'
+    req.breadcrumbs(res.locals.pageTitle)
+    let checkWindowData
+    let checkWindowViewData
+    try {
+      checkWindowData = await checkWindowV2Service.getCheckWindow(req.params.checkWindowUrlSlug)
+      checkWindowViewData = checkWindowHelper.getEditViewData(checkWindowData)
```
jon-shipley: It's not all that clear to me from the code that:

1) `checkWindowHelper` is a Presenter and that 
2) it is returning a view model

@danwilliams @GuyHarwood - thoughts?

admin/helpers/check-window.js
```diff
@@ -0,0 +1,110 @@
...
+  return checkWindowViewData
+}
+
+/**
+ * Prepare validated data for submission
+ * @param {Object} requestData
+ * @param {Number} checkWindowId
+ * @returns {Object} Formatted check window data
+ */
+checkWindowHelper.prepareSubmissionData = (requestData, checkWindowId = null) => {
```
jon-shipley: Is this business logic?  There is a hard-coded check end date at the bottom of the function.

admin/services/check-window-v2-update.service.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Object} requestData
+ * @returns {String} flash message for successful db insertion
+ */
+checkWindowV2UpdateService.submit = async (requestData) => {
+  const checkWindow = await checkWindowV2Service.getCheckWindow(requestData.checkWindowUrlSlug)
+  const validationError = checkWindowAddValidator.validate(requestData)
+  if (validationError.hasError()) {
+    throw validationError
+  }
+  const checkWindowData = checkWindowHelper.prepareSubmissionData(requestData, checkWindow.id)
```
jon-shipley: This is odd. The Presentation helper is now being called by a service.  Which tier of the application is the Helper actually in?  Either Presentation or Business logic service.

admin/views/check-window/create-check-window.ejs
```diff
@@ -454,11 +508,20 @@
...
+            <% if (!checkWindowData.pastCheckWindow) { %>
             <div class="form-buttons">
+                <input type="hidden" name="checkWindowUrlSlug" value="<%= checkWindowData.checkWindowUrlSlug %>">
                 <input class="button" type="submit" value="Save">
                 <a href="/check-window/manage-check-windows" class="button button-secondary">Cancel</a>
             </div>
+            <% } %>
         </form>
     </div>
 </div>
```
jon-shipley: tidy 👌

admin/helpers/check-window.js
GeorgeChatzigiannis: @jon-shipley The familiarisation and live check end dates are actually required to be set at the end time of that dates specified by the user. However you got he thinking about relocating this method elsewhere since its more of business logic.

admin/controllers/check-window.js
GuyHarwood: @GeorgeChatzigiannis is this helper function preparing the data model to be shown in the view?  if so, call it a presenter.

admin/controllers/check-window.js
GuyHarwood: @GeorgeChatzigiannis might be better to just fail/succeed the service via a try/catch call and curate an appropriate flash message in the controller

admin/helpers/check-window.js
GuyHarwood: @GeorgeChatzigiannis if this is structuring the raw data service output that would be useless to anyone without the structuring, then you may want to consider this as business logic that belongs in the service.  However, if this is a formatting exercise just for the web view, put it in a presenter.

admin/helpers/check-window.js
```diff
@@ -0,0 +1,110 @@
...
+      requestData['liveCheckStartYear']
+    )
+  checkWindowData.checkEndDate =
+    dateService.createUTCFromDayMonthYear(
+      requestData['liveCheckEndDay'],
+      requestData['liveCheckEndMonth'],
+      requestData['liveCheckEndYear']
+    )
+  checkWindowData.checkEndDate.set({ hour: 22, minute: 59, second: 59 })
+  checkWindowData.familiarisationCheckEndDate.set({ hour: 22, minute: 59, second: 59 })
```
GuyHarwood: whats the significance of 1 second before 11pm?  maybe put in a variable with a descriptive name.  i'm curious as to what this is for!

admin/controllers/check-window.js
GeorgeChatzigiannis: @GuyHarwood @jon-shipley this is now updated

admin/controllers/check-window.js
GeorgeChatzigiannis: @GuyHarwood @jon-shipley this is updated based on feedback

admin/helpers/check-window.js
GeorgeChatzigiannis: @GuyHarwood I moved this to the service level since it is indeed business logic.

admin/services/check-window-v2-update.service.js
GeorgeChatzigiannis: @jon-shipley the method is moved under check window v2 service.

admin/controllers/check-window.js
```diff
@@ -60,18 +62,30 @@ const controller = {
...
+        flashMessage = `${requestData.checkWindowName} has been created`
+      }
     } catch (error) {
       if (error.name === 'ValidationError') {
-        res.locals.pageTitle = 'Create check window'
+        res.locals.pageTitle = checkWindowUrlSlug ? 'Edit check window' : 'Create check window'
+        let checkWindowViewData = requestData
+        if (checkWindowUrlSlug) {
+          const checkWindowData = await checkWindowV2Service.getCheckWindow(checkWindowUrlSlug)
+          checkWindowViewData = checkWindowPresenter.getViewModelData(checkWindowData, requestData)
```
jon-shipley: 💯

GeorgeChatzigiannis:
waiting for another review from @GuyHarwood on this

admin/helpers/check-window.js
GeorgeChatzigiannis: Originally this was set at 23:59:59 as the last second of the day to cover the full duration of last check window date.
I noticed with BST time change that db was storing it as 00:59:59 of the next day.
I added this to prevent this overlap.

GuyHarwood:
@GeorgeChatzigiannis doesn't that mean its effectively wrong for roughly half the year though?
we baseline in UTC, so can accommodate summertime differences

admin/helpers/check-window.js
GuyHarwood: @GeorgeChatzigiannis doesn't that mean its effectively wrong for roughly half the year though?
we baseline in UTC, so can accommodate summertime differences

GeorgeChatzigiannis: @GuyHarwood I noticed that there was no issue with the last second of the day when storing on db level so I made that change.
To answer your question the previous implementation was also moving the time to the last hour of the day so that the last day would be included in the check window period.

GeorgeChatzigiannis: @GuyHarwood I noticed that there was no issue with the last second of the day when storing on db level so I made that change.
To answer your question the previous implementation was also moving the time to the last hour of the day so that the last day would be included in the check window period.

PR #886, opened 2018-10-24
Feature/25352 school staff generates pupil familiarisation pins

jon-shipley:
This PR adds discrimination between familiarisation checks and live checks.  It also adds Pupil status and an associated pupil status Azure Function to re-compute status for the HPA.  It is anticipated that this retention of state for pupils will dramatically increase the speed of the pupil register on the next PR.

* add new pupil status field
* add a view to show pupils eligible to generate familiarisation pins (uses pupil status)
* update code to use the familiarisation view or the live view depending on the pinEnv
* adds pupil status function to re-compute the status of a pupil

To update the pupil status, send a request to the Azure Queue Service for that pupil to be updated.  Context is not needed, and the status will be determined based on the current state of the pupil, by inspecting pupil checks, restarts, and attendance code.


admin/data/sql/migrations/157.do.backfill-pupil-status.sql
```diff
@@ -0,0 +1,84 @@
+declare @statusId int;
```
GuyHarwood: 💯

admin/data/sql/migrations/166.do.create-view-pupils-eligible-for-familiarisation-pin.sql
```diff
@@ -0,0 +1,45 @@
...
+         WHERE
+                 -- include all pupils except those who are marked as not taking check because they left school
+                (ac.id IS NULL OR ac.code <> 'LEFTT')
+         AND    ps.code IN ('UNALLOC',
+                            'ALLOC',
+                            'LOGGED_IN',
+                            'NOT_TAKING_CHECK')
+
+
+         EXCEPT
```
jon-shipley: If we want to be able to generate a new familiarisation check _at_ _any_ _time_ e.g. because the access arrangements were changed, then we need to lose this `except` and add in a way to expire old familiarisation pins when generating a new familiarisation check.

admin/controllers/pupil-pin.js
```diff
@@ -24,7 +24,7 @@ const getGeneratePinsOverview = async (req, res, next) => {
   let pupils
   try {
     if (featureToggles.isFeatureEnabled('prepareCheckMessaging')) {
-      pupils = await pinGenerationV2Service.getPupilsWithActivePins(req.user.schoolId, pinEnv)
+      pupils = await pinGenerationV2Service.getPupilsWithActivePins(req.user.schoolId, pinEnv === 'live')
```
GuyHarwood: why are we hardcoding this to live?

jon-shipley: It's an expression that returns true or false, as the parameter is a boolean.

😀 I assume this is a pre-coffee review 😀

GuyHarwood: brewing now 🤦‍♂️

functions/pupil-status/pupil-status-analysis.service.js
```diff
@@ -0,0 +1,59 @@
...
+ * Data Format
+ [ { pupil_id: 12,
+   pupilStatusCode: 'UNALLOC',
+   check_id: 1,
+   checkStatusCode: 'NEW' }
+ ]
+ */
+
+module.exports = {
+  analysePupilData: function analysePupilData (data) {
```
GeorgeChatzigiannis: @jon-shipley would it be possible to provide a jsdoc description about this method?

restart.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
GeorgeChatzigiannis: 💯

admin/data/sql/migrations/160.do.populate-pupil-status-table.sql
```diff
@@ -0,0 +1,8 @@
+insert into [mtc_admin].[pupilStatus] (code, description)
+values
+       ('UNALLOC', 'The pupil has not been allocated a check'),
```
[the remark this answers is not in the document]
GuyHarwood: @GeorgeChatzigiannis i think that will just make them more difficult to decipher.  There is no real need to limit them to 3 chars, so i vote they are left as-is.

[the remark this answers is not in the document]
jon-shipley: @GeorgeChatzigiannis - they are more than 3 chars as I found the other codes in other tables harder to use than I expected.   As @GuyHarwood mentioned, the performance difference is negligible, and I think the clarity is very much improved.

functions/pupil-status/pupil-status-analysis.service.js
jon-shipley: Good spot!  JSDoc improved.

admin/controllers/pupil-pin.js
```diff
@@ -54,6 +54,7 @@ const getGeneratePinsOverview = async (req, res, next) => {
  */
 const getGeneratePinsList = async (req, res, next) => {
   const pinEnv = (req.params && req.params.pinEnv === 'live') ? 'live' : 'familiarisation'
+  const isLiveCheck = (req.params && req.params.pinEnv === 'live') ? true : false;
```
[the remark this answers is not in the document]
jon-shipley: I don't think so.

```
> node -i
> const req = { paramsMissing: 1 }
undefined
> req.params && req.params.pinEnv === 'live'
undefined
> (req.params && req.params.pinEnv === 'live') ? true : false
false
> 
```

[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley I went with this approach on #895 
```suggestion
  const isLiveCheck = (req.params && req.params.pinEnv === 'live') ? true : false;
```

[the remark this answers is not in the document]
jon-shipley: @GeorgeChatzigiannis - thanks for the suggestion but what is there seems clearer to me at least.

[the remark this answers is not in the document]
GeorgeChatzigiannis: Agreed with your final approach and I will be making changes on #895 to comply.

admin/services/pin-generation-v2.service.js
```diff
@@ -30,12 +30,12 @@ const serviceToExport = {
   /**
    * Return a list of pupils who have an active pin - one that can be used to take a check
    * TODO: add pinEnv
```
danwilliams: I think that can be updated, as the ‘isLiveCheck’ param was added...?

admin/spec/back-end/controllers/restart.spec.js
```diff
@@ -1,11 +1,14 @@
 'use strict'
-/* global describe it expect beforeEach jasmine spyOn fail */
+/* global describe it expect beforeEach jasmine spyOn fail xit */
```
danwilliams: Why are the tests added commented out with `xit`, is it because of the feature toggle? 🤔 If so, can that be mocked and the tests ran?

jon-shipley: Yes, these tests are part of the HPA code paths.  

We may need to run unit tests in legacy mode, and also in HPA mode?  If so we could feature toggle the tests.

Perhaps we need a new ticket though?

admin/services/pin-generation-v2.service.js
jon-shipley: Good spot! Now removed

admin/spec/back-end/controllers/restart.spec.js
danwilliams: New ticket sounds sensible 👍

PR #889, opened 2018-10-25
messaging standards doc for functions and versioning directory for each function

GuyHarwood:


GuyHarwood:
@GeorgeChatzigiannis that's a larger piece of work that spreads across the solution, but it would be 'preferable' for sure.  it would help with debugging

PR #894, opened 2018-11-01
getting a e2e check test working

activemq:
A e2e check test has been tagged for a smoke test run. 
Still to do - speak with Amit to figure out which admin tests we can run as part of the smoke test that have already been refactored


GuyHarwood:
@activemq closing this until you come back to it.

PR #895, opened 2018-11-02
Feature/24366 Inaccessible pin generation links 

GeorgeChatzigiannis:
- [x] Behaviour for pin generation eligibility and reflected UI work on school page including override
- [x] Prevent/allow route redirection based on pin generation eligibility
- [x] UI update on view and print pins page
- [x] Behaviour to prevent pin generation links under related sections


admin/views/school/school-home.ejs
```diff
@@ -38,13 +38,43 @@
             </li>
 
             <li>
+                <% if (pinGenerationEligibilityData.isFamiliarisationPinGenerationAllowed && !pinGenerationEligibilityData.isUnavailableTime) { %>
```
jon-shipley: Why not just have one flag to check?   It should just be `isFamiliarisationPinGenerationAllowed`?  Why complicate matters?

admin/views/school/school-home.ejs
```diff
@@ -38,13 +38,43 @@
...
                 <h2 class="heading-small"><a href="/pupil-pin/generate-familiarisation-pins-overview">Generate pupil PINs - familiarisation check</a></h2>
                 <p>Generate personal identification numbers (PINs) for pupils and view school password for the familiarisation check</p>
+                <% } %>
+                <% if (pinGenerationEligibilityData.isUnavailableTime || !pinGenerationEligibilityData.isFamiliarisationPinGenerationAllowed) { %>
```
jon-shipley: Again, should this be controlled by a single flag?

admin/views/school/school-home.ejs
```diff
@@ -38,13 +38,43 @@
...
+                <h2 class="heading-small"><a class="font-greyed-out disabled-link" href="#">Generate pupil PINs - familiarisation check</a></h2>
+                <span>Generate personal identification numbers (PINs) for pupils and view school password for the familiarisation check</span>
+                <p>
+                    <strong class="disabled-banner">UNAVAILABLE</strong>
```
jon-shipley: Does the UI ask for all uppercase?  It's generally considered to be less readable.

admin/services/data-access/check-window.data.service.js
```diff
@@ -326,6 +326,18 @@ const checkWindowDataService = {
...
     ORDER BY name ASC`
     return sqlService.query(sql)
+  },
+
+  /**
+   * Fetch active check window
+   * @return {Object}
+   */
+  sqlFindActiveCheckWindow: async () => {
+    const sql = `SELECT *
```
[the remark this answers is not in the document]
GuyHarwood: yep, you could end up with the wrong record quite easily here, needs to be strict.

admin/services/pin-generation-eligibility.service.js
```diff
@@ -0,0 +1,37 @@
...
+const monitor = require('../helpers/monitor')
+
+const pinGenerationEligibilityService = {}
+
+/**
+ * Return pin generation availability
+ * @param pinEnv live or familiarisation
+ * @returns {Boolean | Error} live pin generation allowance
+ */
+pinGenerationEligibilityService.isPinGenerationAllowed = async (pinEnv) => {
```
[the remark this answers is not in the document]
GuyHarwood: seconded.  use a bool.

admin/helpers/school-home-pin-generation-eligibility-presenter.js
```diff
@@ -0,0 +1,33 @@
...
+    familiarisationCheckStartDate: checkWindow.familiarisationCheckStartDate,
+    familiarisationCheckEndDate: checkWindow.familiarisationCheckEndDate,
+    liveCheckStartDate: checkWindow.checkStartDate,
+    liveCheckEndDate: checkWindow.checkEndDate,
+    isFamiliarisationPeriodActive: currentDate.isAfter(checkWindow.familiarisationCheckStartDate) && currentDate.isBefore(checkWindow.familiarisationCheckEndDate),
+    isFamiliarisationPinGenerationAllowed: (currentDate.isAfter(checkWindow.adminStartDate) && currentDate.isBefore(checkWindow.familiarisationCheckStartDate)) || overridePinGenerationEligibility,
+    isLivePeriodActive: currentDate.isAfter(checkWindow.checkStartDate) && currentDate.isBefore(checkWindow.checkEndDate),
+    isLivePinGenerationAllowed: (currentDate.isAfter(checkWindow.adminStartDate) && currentDate.isBefore(checkWindow.checkStartDate)) || overridePinGenerationEligibility,
+    isPinGenerationInTheFuture: currentDate.isBefore(checkWindow.adminStartDate),
+    isUnavailableTime: currentDate.hour() < 8 || currentDate.hour() > 16
```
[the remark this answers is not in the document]
GuyHarwood: would be easier to understand if you broke it up into variable assignments that are named into the rule, then just return the object at the end.  Shorthand often just makes it more difficult to follow.

admin/controllers/pupil-pin.js
```diff
@@ -23,6 +24,7 @@ const getGeneratePinsOverview = async (req, res, next) => {
   const helplineNumber = config.Data.helplineNumber
   let pupils
   try {
+    await pinGenerationEligibilityService.determinePinGenerationEligibility(pinEnv)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley added.

admin/services/pin-generation-eligibility.service.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @GuyHarwood @jon-shipley this is now covered as asked.

admin/services/pin-generation-eligibility.service.js
```diff
@@ -0,0 +1,37 @@
...
+
+const schoolHomePinGenerationEligibilityPresenter = require('../helpers/school-home-pin-generation-eligibility-presenter')
+const monitor = require('../helpers/monitor')
+
+const pinGenerationEligibilityService = {}
+
+/**
+ * Return pin generation availability
+ * @param pinEnv live or familiarisation
+ * @returns {Boolean | Error} live pin generation allowance
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley I replaced that with the `@throws statement`.

admin/views/school/school-home.ejs
GeorgeChatzigiannis: @jon-shipley the requirement is indeed asking for uppercase.

admin/spec/back-end/service/pin-generation-eligibility.service.spec.js
```diff
@@ -15,38 +15,38 @@ describe('pinGenerationEligibilityService', () => {
       })
     })
     it('should return true if live pin generation is allowed', async () => {
-      const result = await pinGenerationEligibilityService.isPinGenerationAllowed('live')
+      const result = await pinGenerationEligibilityService.isPinGenerationAllowed(true)
```
GuyHarwood: just a note in terms of readability - it would be better to have a bool called `liveCheckMode` set to true here.  currently it looks like you are saying 'yes, pin generation is allowed'

admin/controllers/school.js
```diff
@@ -1,28 +1,32 @@
...
  * @returns {Promise<*>}
  */
-const getSchoolLandingPage = async (req, res, next) => {
+controller.getSchoolLandingPage = async (req, res, next) => {
   res.locals.pageTitle = 'School Homepage'
+  let pinGenerationEligibilityData
   let schoolName = ''
-
   try {
+    pinGenerationEligibilityData = await schoolHomePinGenerationEligibilityPresenter.getEligibilityData()
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley I added a comment above and corrected the method name.
Essentially what it does is fetching flags to determine the UI behaviour.
Let me know your thoughts.

admin/helpers/school-home-pin-generation-eligibility-presenter.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley @GuyHarwood I have reimplemented this method since it was not matching the business expectations.

admin/services/data-access/check-window.data.service.js
```diff
@@ -326,6 +326,18 @@ const checkWindowDataService = {
...
     ORDER BY name ASC`
     return sqlService.query(sql)
+  },
+
+  /**
+   * Fetch active check window
+   * @return {Object}
+   */
+  sqlFindActiveCheckWindow: async () => {
+    const sql = `SELECT TOP 1 *
```
jon-shipley: This is better in that it will only return one or zero rows, but it's still possible for this query to return the wrong rows by omission if there are multiple check windows active at the same time.

It ought to be possible to close this off at the database level by ensuring that any new `checkWindow` must NOT overlap any existing `checkWindow`.

This could be done by a trigger to detect overlaps perhaps?

```
select
	original.name,
	original.checkStartDate,
	original.checkEndDate,
	overlap.name,
	overlap.checkStartDate,
	overlap.checkEndDate
	
from
	[mtc_admin].[checkWindow] as original 
	inner join [mtc_admin].[checkWindow] as overlap ON (original.id <> overlap.id)
 
where

  -- overlapping, not just touching
  original.checkEndDate > overlap.checkStartDate AND
  overlap.checkEndDate > original.checkStartDate AND

  -- not exact duplicate
  original.checkEndDate <> overlap.checkEndDate AND
  original.checkStartDate <> overlap.checkStartDate
 
 group by 
 	original.name,
	original.checkStartDate,
	original.checkEndDate,
	overlap.name,
	overlap.checkStartDate,
	overlap.checkEndDate
;
```

GuyHarwood: I would prefer this as a business rule in code.

George - if i recall correctly there are existing rules around this?

GeorgeChatzigiannis: Not existing but there is a set of rules to-be-implemented to cover this from the service manager perspective.

GuyHarwood: yes, as agreed on standup.  Thanks George

admin/helpers/school-home-pin-generation-eligibility-presenter.js
```diff
@@ -0,0 +1,39 @@
...
+const monitor = require('../helpers/monitor')
+
+const schoolHomePinGenerationEligibilityPresenter = {}
+
+/**
+ * Fetch data for familiarisation and live pin generation eligibility
+ * @returns {Object} Eligibility data including flags and relevant datetimes
+ */
+schoolHomePinGenerationEligibilityPresenter.getPresentationData = async () => {
+  const checkWindow = await checkWindowV2Service.getActiveCheckWindow()
```
jon-shipley: What happens if there isn't an active check window?

admin/views/access-arrangements/overview.ejs
```diff
@@ -38,10 +38,14 @@
                         <a href="/pupil-register/pupils-list">Pupil register</a>
                     </li>
                     <li>
+                        <% if (pinGenerationEligibilityData.isFamiliarisationPinGenerationAllowed) { %>
```
jon-shipley: Nice test 🏅

admin/helpers/school-home-pin-generation-eligibility-presenter.js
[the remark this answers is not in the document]
GuyHarwood: ideally the presenter should be called from the controller, which passes the data to the presenter, which it obtained from the service.

GeorgeChatzigiannis: @jon-shipley we shouldn't reach that case ever since the expectation would be for the teacher to be unable to login on the teacher admin panel when there is no active check window

admin/helpers/school-home-pin-generation-eligibility-presenter.js
```diff
@@ -0,0 +1,39 @@
...
+  const overridePinGenerationEligibility = config.OverridePinExpiry
+  const currentDate = moment.utc()
+  const pinGenerationEligibilityData = {}
+  const isWithinRestrictedHours = currentDate.hour() <= 8 || currentDate.hour() >= 16
+  const isWithinFamiliarisationPeriod = currentDate.isAfter(checkWindow.familiarisationCheckStartDate) && currentDate.isBefore(checkWindow.familiarisationCheckEndDate)
+  const isWithinLivePeriod = currentDate.isAfter(checkWindow.checkStartDate) && currentDate.isBefore(checkWindow.checkEndDate)
+  // Familiarisation data
+  pinGenerationEligibilityData.familiarisationCheckStartDate = checkWindow.familiarisationCheckStartDate
+  pinGenerationEligibilityData.familiarisationCheckEndDate = checkWindow.familiarisationCheckEndDate
+  pinGenerationEligibilityData.isFamiliarisationPinGenerationAllowed = (isWithinFamiliarisationPeriod && !isWithinRestrictedHours) || overridePinGenerationEligibility
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley there are cases below where `isWithinRestrictedHours` is used because the negative case is needed.

admin/helpers/school-home-pin-generation-eligibility-presenter.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @GuyHarwood @jon-shipley agreed and updated.

admin/helpers/school-home-pin-generation-eligibility-presenter.js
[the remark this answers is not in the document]
jon-shipley: @GeorgeChatzigiannis - that's not idiomatic English though.  We would usually say 'inside opening hours' or 'outside opening hours' for the hours for a shop for example.  Saying 'within' is similar to 'inside', the 'within' bit sets up an expectation in the English speaker that the hours are the usual hours.  There is not much use of 'restricted hours' expect perhaps short trading days like Sunday.

Various style guides usually suggest naming variables for the positive case only.

@GuyHarwood @danwilliams - can we take a vote on this in case it's just me?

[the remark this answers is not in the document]
danwilliams: After discussion, I would vote for `isWithinOpeningHours`, or similar. This may be slightly better than `isWithinRestrictedHours`, and more obvious in cases like this than `!isWithinRestrictedHours`, which I think is the core point here. However, I don't have strong feelings on this, and if George feels the word/concept "restricted" needs additional specific focus, then I'm also not against leaving it as it is.

admin/controllers/pupil-pin.js
```diff
@@ -116,6 +133,7 @@ const postGeneratePins = async (req, res, next) => {
   }
   let school
   try {
+    await pinGenerationEligibilityService.determinePinGenerationEligibility(isLiveCheck)
```
jon-shipley: This is essentially a validator isn't it? I wonder if we should call it that?

admin/helpers/school-home-pin-generation-eligibility-presenter.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Fetch data for familiarisation and live pin generation eligibility
+ * @param checkWindowData
+ * @returns {Object} Eligibility data including flags and relevant datetimes
+ */
+schoolHomePinGenerationEligibilityPresenter.getPresentationData = async (checkWindowData) => {
+  const overridePinGenerationEligibility = config.OverridePinExpiry
+  const currentDate = moment.utc()
+  const pinGenerationEligibilityData = {}
+  const isWithinOpeningHours = currentDate.hour() > 8 && currentDate.hour() < 16
```
jon-shipley: Thanks very much for the change.  This reads very nicely now!

admin/controllers/pupil-pin.js
GeorgeChatzigiannis: @jon-shipley renamed to `businessAvailabilityService` as agreed

PR #896, opened 2018-11-02
adding test commit

activemq:
Something has broken a lot of travis tests on a couple of PR's. The tests that fail on travis work fine locally, I want to use this PR to figure out which commit broke them on travis and see how we can resolve the issue


GuyHarwood:
@activemq i suggest we close this until we need to come back to it.

PR #897, opened 2018-11-07
Feature/25648 colour fonts applied

danwilliams:
Please note, this depends upon 25630, and the commit history currently contains that. I will be rebasing once 25630 is in master. Hence I'm putting this up for early review but the review is for commits from 25 October onwards. This will be clearer after the rebase, but I would prefer not to wait, and open this up for feedback now.


pupil-spa/src/assets/shared-styles/elements/_colour-contrast.scss
```diff
@@ -0,0 +1,140 @@
...
+
+  label {
+    color: $copy-colour !important;
+  }
+
+  .button-start {
+    background: none;
+  }
+
+  // Fill eyes when YOB
```
jon-shipley: What does this mean?  Is there a screenshot?

danwilliams: Heheh 😄 it's referring to the SVG, which is being styled according to what's needed.

PR #901, opened 2018-11-07
Bug/26009 hpa pupil register page is throwing 500

jon-shipley:
* Prevent error being thrown by ensuring var is defined
* Backfill check started date with audit data if the data is not yet there when the completed check message is received
* prevent duplicate check submissions


functions/completed-checks/index.js
```diff
@@ -79,6 +79,25 @@ async function savePayloadToAdminDatabase (completedCheckMessage, logger) {
...
 
+  // Don't process any checks more than once
+  if (check.receivedByServerAt) {
+    const msg = `completed-check: ERROR: payload re-submission is banned for check ${checkData.checkCode}`
+    logger.error(msg)
+    throw new Error(msg)
+  }
+
+
+  if (!checkData.startedAt) {
```
GeorgeChatzigiannis: @jon-shipley under which circumstances the `startedAt` field would not be populated?

GeorgeChatzigiannis:
@jon-shipley looks good just 1 comment

functions/completed-checks/index.js
GeorgeChatzigiannis: is it only in case check-started function fails?

jon-shipley: @GeorgeChatzigiannis 

1.  The client browser had networking issues
2. The completed check message was received before the check started message (busy system?)

GeorgeChatzigiannis: Ok that makes sense to me, thanks.

PR #902, opened 2018-11-08
Remove monitor calls

jon-shipley:
Remove monitor calls in favour of code editor navigation


GeorgeChatzigiannis:
@jon-shipley one question, do we need to keep monitor.js if it is not being used in the files anymore?

PR #904, opened 2018-11-13
Feature/26376 hpa pupil login date not being recorded in the check table

jon-shipley:
Save pupil login date to the database when the pupil logs in (HPA mode)

* new pupil login function processor
* emit login date to the queue when logging-in to the pupil-spa


functions/pupil-login/index.js
```diff
@@ -1,8 +1,25 @@
...
+
 module.exports = async function (context, message) {
   context.log(
     `processing pupil-login message version:${message.version} for check:${message.checkCode}`
   )
-  context.done()
+
+  validateMessage(message)
+
+  switch (message.version) {
```
GeorgeChatzigiannis: @jon-shipley I think the switch is overcomplicating things here and there's only 1 case.
Why not a simple if?

functions/pupil-login/message-validator.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
```
GeorgeChatzigiannis: clean 🤘

functions/pupil-login/v1.spec.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+/* global describe expect it spyOn */
+
+const sqlService = require('less-tedious')
+const v1 = require('./v1')
+
+describe('v1-process', () => {
+  it('sqlService.modify', async () => {
```
GeorgeChatzigiannis: @jon-shipley can you please provide a better description for that test?
Doesn't seem very readable to me.

pupil-api/src/services/azure-queue.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!queueService) {
+    // If we have not been provided with a queueService, assume we want an Azure one.
+    if (!azureQueueService) {
+      azureQueueService = azure.createQueueService()
+    }
+    queueService = azureQueueService
+  }
+
+  const message = JSON.stringify(payload)
```
GeorgeChatzigiannis: @jon-shipley what happens if the payload is null|empty|undefined?

pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -29,26 +33,53 @@ function initAzureTableService (): void {
...
     // set tableService to azureTableService if not provided, but only instantiate it once.
     if (!tableService) {
       initAzureTableService()
       tableService = azureTableService
     }
 
+    // This will throw if the pupil is not found
     const { result } = await tableService.retrieveEntityAsync(authTable, schoolPin, pupilPin)
 
+    if (!path(['pinExpiresAt', '_'], result)) {
```
GeorgeChatzigiannis: is the `_` the direct value in these cases?

functions/pupil-login/index.js
jon-shipley: @GeorgeChatzigiannis - this is a versioned message format - this pattern can be extended easily to allow future versions.

If we had an if clause we would still require the else part to cover the default case of a rejecting a message it can't process.

functions/pupil-login/v1.spec.js
jon-shipley: Me Neither ☹️ I'll fix it.

pupil-api/src/services/azure-queue.service.ts
jon-shipley: I don't think this service needs to care.  In any case in the flow in the pupil-api the message is guaranteed:

```
    // Emit a successful login to the queue
    const pupilLoginMessage = {
      checkCode: result.checkCode._,
      loginAt: new Date(),
      version: 1
    }
```


And `JSON.stringify` will handle those cases you mentioned without throwing an error:

```
> JSON.stringify(null)
'null'
> JSON.stringify()
undefined
> JSON.stringify(undefined)
undefined
> 
```

pupil-api/src/services/pupil-authentication.service.ts
jon-shipley: Yes.  Blame Azure Table Storage.

PR #908, opened 2018-11-19
Feature/26630 prepared check sync prerequisites 

GeorgeChatzigiannis:
Added necessary config, migration changes and queues for prerequisites


functions/prepaped-check-sync/message-schema/message.v1.json
```diff
@@ -0,0 +1,78 @@
+{
```
GuyHarwood: is it the responsibility of the message sender to provide the updated payload?  or should it just indicate the checkCode of the check that needs resyncing?  the latter seems feels like a better choice

GuyHarwood:
@GeorgeChatzigiannis added CI label, as you are adding DB changes that need regression testing

functions/prepaped-check-sync/message-schema/message.v1.json
GeorgeChatzigiannis: @GuyHarwood I would think the prepared-check-sync function will fetch the relevant `PartitionKey`, `RowKey `,  and `config` then update the latter with the new settings and push to table storage?
In that case the message would not need the checkCode but only these 3 properties that I have highlighted.

Thoughts? 🤔

admin/data/sql/migrations/183.do.index-pupil-urlslug-check-checkcode.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX check_checkCode_index ON [mtc_admin].[check] (checkCode)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley @GuyHarwood updated.

PR #909, opened 2018-11-20
Feature/25939 sql migrations

danwilliams:


danwilliams:
Thanks @GuyHarwood! 😄

admin/data/sql/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+Generated migrations have the filename yyyymmddhhmmss.<do|undo>.name.<sql|js>
+
+
+## Running migrations
+
+- The migration script will execute *all* migrations up or down to a given version
+
+```
+node sql/migrate-sql.js
```
GeorgeChatzigiannis: What is the way to undo all the migrations?
With the existing implementation we could perform that by running `yarn migrate-sql 0`.
Is this already supported?

GuyHarwood: @GeorgeChatzigiannis that's exactly what it says George.  it will execute all migrations up or down to a given version

GuyHarwood:
@danwilliams i think travis may be failing because the seeds are executed separately?  if so, we will need to add a command to `travis.yml` for that

PR #912, opened 2018-11-21
Feature/26672 pupil views with font size

danwilliams:


pupil-spa/src/app/warmup-complete/warmup-complete.component.html
```diff
@@ -1,14 +1,24 @@
...
+      <div class="grid-row">
+        <br />
+        <div class="column-two-thirds">
+          <button type="button" class="button aa-button-size aa-contrast-button" id="start-now-button" (click)="onClick()">
+            Next
+          </button>
+        </div>
+      </div>
+    </div>
+    <div *ngIf="shouldShowMore" class="column-one-third more-links aa-copy-size">
```
GeorgeChatzigiannis: @danwilliams question here, I can see `shouldShowMore` relies on being in a familiarisation check.
How come we are showing different elements for the two types on warmup complete?
I was under the impression the content should be identical.

GeorgeChatzigiannis: @danwilliams I am approving this under the speculation that it only applies on familiarisation checks.

PR #915, opened 2018-11-27
Bug/fix migrations

jon-shipley:
Update the seeder data so that pupils are generated for the teacher1, teacher2, teacher3, teacher4, teacher5 test logins.


activemq:
LGTM 👍 Generated a pin and logged in

PR #916, opened 2018-11-27
Feature/26671 pupil views with question reader

danwilliams:


pupil-spa/src/app/services/speech/speech.service.ts
```diff
@@ -199,16 +211,16 @@ export class SpeechService implements OnDestroy {
   speakFocusedElement(nativeElement): void {
     const { id, nodeName, parentNode } = nativeElement;
 
-    let toSpeak = nativeElement;
+    let toSpeak = nativeElement.cloneNode(true);
```
GeorgeChatzigiannis: @danwilliams out of curiosity why did you need to clone?

pupil-spa/src/app/services/speech/speech.service.ts
```diff
@@ -147,13 +147,11 @@ export class SpeechService implements OnDestroy {
...
-   * Parse the source of a NativeElement and speak the text
+   * Removes aria-hidden child elements
    * @param nativeElement
+   * @returns {clonedElement}
    */
-  speakElement(nativeElement): Promise<{}> {
-    this.focusInterruptedPageSpeech = false;
-    const elementsToSpeak = 'h1, h2, h3, h4, h5, h6, p, li, button, a, span, fieldset';
-
+  removeUnspokenElements(nativeElement): HTMLElement {
```
GeorgeChatzigiannis: @danwilliams I see the `speakFocusedElement` already clones the element.
Does this function still need to clone the already cloned node?

PR #917, opened 2018-11-28
Feature/23751 pupil api improvements

danwilliams:


pupil-api/src/App.ts
```diff
@@ -12,6 +12,8 @@ import * as winston from 'winston'
 import * as expressWinston from 'express-winston'
 import * as azure from './azure'
 const config = require('./config')
+const corsOptions = require('../../admin/helpers/cors-options')
```
GuyHarwood: how do we set the permitted domains for CORS rules?

pupil-api/readme.md
```diff
@@ -7,44 +7,6 @@ Required environment variables
...
-#### Body
-```
-{
-	"pupilPin": "<some pin>",
-	"schoolPin": "<some pin>"
-}
-```
+## Swagger API Documentation
 
+When running in development the api documentation can be found at http://devserver:3003/api-docs
```
jon-shipley: `devserver` seems arbitrary.  Wouldn't `localhost` be better,  as it is more common,  here?

pupil-api/src/App.ts
jon-shipley: I am not sure that we should rely on this behaviour - that it requires a file from another project, in this case, `admin`.    This file also includes the admin `config` file which seems potentially dangerous.

It makes packaging less obvious.  Let's not require files outside of the current application's directory.

pupil-api/src/App.ts
```diff
@@ -12,6 +12,8 @@ import * as winston from 'winston'
...
+const setupLogging = require('../../admin/helpers/logger')
```
jon-shipley: Same comment as above here.

pupil-api/src/App.ts
GuyHarwood: @jon-shipley agreed

pupil-api/readme.md
GuyHarwood: agreed.  @danwilliams please change docs.

pupil-api/src/App.ts
GuyHarwood: agree on this too.  please include in this project and also allow for setup via env vars.  IIRC thats how the admin app does it.

pupil-api/readme.md
danwilliams: I am unsure that I agree with changing it to localhost... I could note that localhost is a possibility, but it can critically be any hostname, and I think it's dangerous to rely too much on localhost as a default?

danwilliams: `devserver` is a generic name to illustrate a concept. My dev machines are more interestingly-named 😄 The point of what I was saying is that localhost is not the only host to use, and should not really be the only host we work with or imply. (I don't think this is the place for me to go into all the reasons not to use localhost.) In any case, the change was made - I thought I'd just make this further note.

GuyHarwood: point acknowledged :) 
i think it makes it obvious to the reader when it says `localhost`

PR #918, opened 2018-11-28
feature/26631 Pupil prefs queue submission to prepared-check-sync

GeorgeChatzigiannis:
- [x] Submit message to the prepared-check-sync queue after pupil preferences have been updated
- [x] Restructure pupil prefs function
- [x] Backfill missing tests from previous implementation


functions/pupil-prefs/v1.js
```diff
@@ -0,0 +1,105 @@
...
+  const { checkCode, preferences } = pupilPrefsMessage
+  const { fontSizeCode, colourContrastCode } = preferences
+  context.log('pupil-prefs: message received', checkCode, fontSizeCode, colourContrastCode)
+
+  // Update pupil access arrangements preference based on code presence in the payload
+  try {
+    if (fontSizeCode) {
+      await v1.updatePupilAccessArrangementsPreference(checkCode, 'pupilfontSizes_id', pupilFontSizesTable, fontSizeCode, 'FTS')
+      context.log(`pupil-prefs: SUCCESS: pupil access arrangement row updated for checkCode ${checkCode} and font size code ${fontSizeCode}`)
+    }
```
jon-shipley: Should this be an else clause, rather than sequential `if`s?   

If I see two `if`s then I assume that both can be true in the same run, whereas, if there is an if/else construct then I know the values are exclusive.

functions/pupil-prefs/v1.js
```diff
@@ -0,0 +1,105 @@
...
+ */
+v1.updatePupilAccessArrangementsPreference = async function (checkCode, prefField, prefTable, prefCode, accessArrangementCode) {
+  const sql = `UPDATE ${schema}.${pupilAccessArrangementsTable}
+               SET ${prefField} = (
+                SELECT id FROM ${schema}.${prefTable} 
+                WHERE code = @prefCode
+               )                
+               WHERE pupil_Id = (
+                  SELECT p.id FROM ${schema}.${pupilTable} p
+                  LEFT OUTER JOIN ${schema}.${checkTable} chk
```
jon-shipley: If we have a checkCode then shouldn't the join be guaranteed?

functions/pupil-prefs/v1.js
GeorgeChatzigiannis: @jon-shipley updated

functions/pupil-prefs/v1.js
GeorgeChatzigiannis: @jon-shipley changed to `INNER JOIN`

functions/pupil-prefs/v1.js
GeorgeChatzigiannis: As discussed I am undoing this since the SPA will always submit both colour contrast and font size even with their default values

functions/lib/sql-helper.js
```diff
@@ -26,3 +26,25 @@ module.exports.sqlFindCheckByCheckCode = async function (checkCode) {
...
   return R.head(res)
 }
+
+/**
+ * Retrieve all pupil checks based on check code
+ * @param checkCode
+ * @return {Promise<object>}
+ */
+module.exports.sqlFindChecksByCheckCode = async function (checkCode) {
+  const sql = `SELECT * from mtc.mtc_admin.[check]
```
jon-shipley: There is no need to explicitly state the database name, that's determined by someone else.

GeorgeChatzigiannis: this is fixed please ignore :)

functions/lib/sql-helper.js
```diff
@@ -26,3 +26,25 @@ module.exports.sqlFindCheckByCheckCode = async function (checkCode) {
   const res = await sqlService.query(sql, params)
   return R.head(res)
 }
+
+/**
+ * Retrieve all pupil checks based on check code
```
jon-shipley: More sense to me: 'Find all checks for a particular pupil based on one of the pupil's `checkCodes`'

functions/prepared-check-sync/v1.js
```diff
@@ -5,13 +5,30 @@ const uuid = require('uuid/v4')
...
 const preparedCheckSchemaValidator = require('../lib/prepared-check-schema-validator')
 
 const v1 = {}
 
 v1.process = async function process (context, preparedCheckSyncMessage) {
   const { checkCode } = preparedCheckSyncMessage
   context.log('prepared-check-sync: message received', checkCode)
+  const checks = await sqlUtils.sqlFindChecksByCheckCode(checkCode)
+  await Promise.all(checks.map(async (check) => {
+    await v1.updatePreparedChecks(context, check.checkCode)
```
jon-shipley: Why `await` in the `map` as well as on the `Promise.all`?

functions/lib/sql-helper.js
GeorgeChatzigiannis: @jon-shipley well said, updated.

functions/prepared-check-sync/v1.js
GeorgeChatzigiannis: @jon-shipley both that await and the Promise.all seem redundant to me so I removed them.

PR #919, opened 2018-11-29
27017 - Updated AppUsageService to load counter from local storage

danwilliams:
- Load the device data from local storage in the AppUsageService constructor
- Added a store() method to AppUsageService
- Call store() method in SubmissionService
- Added tests


pupil-spa/src/app/services/app-usage/app-usage.service.ts
```diff
@@ -1,12 +1,22 @@
...
+
 
 @Injectable()
 export class AppUsageService {
 
-  private appUsageCounter;
+  private appUsageCounter: number;
+
+  constructor(private storageService: StorageService) {
+    const deviceData = storageService.getItem(DeviceService.storageKey);
```
GuyHarwood: we decided earlier this week not to use localStorage, as it seemed overkill.  are you doing this because you are unable to persist the counter across page reloads / re-logins?

GuyHarwood:
@activemq this needs to come out of testing.  @GeorgeChatzigiannis has raised a valid point that i overlooked in my earlier review

activemq:
@guy

> @activemq this needs to come out of testing. @GeorgeChatzigiannis has raised a valid point that i overlooked in my earlier review

@GuyHarwood agreed, just proved this is not working atm

pupil-spa/src/app/services/app-usage/app-usage.service.ts
```diff
@@ -1,19 +1,29 @@
...
+
+  constructor(private storageService: StorageService) {
+  }
 
-  constructor() {
-    this.appUsageCounter = 0;
+  store(): void {
+    // Store the appUsageCounter in the device data so it is
+    // picked up later by storageService.getAllItems()
+    const deviceData = this.storageService.getItem(DeviceService.storageKey) || {};
```
jon-shipley: The entire point of the `appUsageCounter` is to not persist across reloads.  It is intended for the anomaly report so that we can see the number of checks completed since the app was reloaded.

Note that we do *not* currently test the SPA app with multiple tests taken sequentially.

The idea is that if we see a large number in the app usage counter, and some unexplained anomaly in the returned data, we might then have a clue as to the behaviour of the app on that device / OS / with x number of checks.

So there is no point storing it in local storage.  The ideal behaviour would be an in-memory counter that is initialised to zero on page load and injected into the payload on submission.

GeorgeChatzigiannis: Valid point I completely missed that on my review

danwilliams: Hmmmm... I think you are looking at local storage inside app usage counter and assuming that we’re still storing the value of the counter (which we’re not). We’re using local storage in app usage counter to add the counter to the device data so it's included in `localstorage.getAllItems()`. `localstorage.getAllItems()` is used in the submission-service and check-complete-service to build a submission payload. If we don’t add counter to the device data in app-usage-service we will have code duplication in the submission-service and check-complete-service, adding the counter to the device data _after_ its been loaded from storage.

Happy to change it if you still disagree with this approach? Just wanted to ensure it is understood.

danwilliams: ...the point is that it’s not stored for _persistence_ in localStorage, that’s only used as an intermediary place to keep the usage counter to avoid the duplication for hpa / legacy code when sending the final payload.

jon-shipley: @danwilliams - thanks for the update, that does clarify things further.  I'll approve the PR, however, I would prefer an approach where the submission service calls something to gather and orchestrate the data it needs to submit, rather than assuming that the storage layout in localStorage is what is required to be sent.  The current approach is no fault of this PR, but seems brittle to me.

Having to write an in-memory counter to localStorage to facilitate sending to the back-end is not a clean approach in my view.

GuyHarwood: i also noticed this when i did a previous PR on the app usage counter - its not explicit enough that it requires that value for submission.  i wrote some tests around capturing it in the payload, as these were missing previously.

activemq:
@danwilliams Seems like there is a issue when building the pupil app 

> [0] ERROR in src/app/services/check-complete/check-complete.service.ts(4,10): error TS2300: Duplicate identifier 'AppUsageService'.
[0] src/app/services/check-complete/check-complete.service.ts(16,10): error TS2300: Duplicate identifier 'AppUsageService'.

PR #924, opened 2018-12-03
Feature/27090 Upload check form UI v2 

GeorgeChatzigiannis:
- [x] Create new views for submitting check form v2 when feature toggle `newCheckForm` is enabled
- [x] Remove disabled attribute on upload button when file is not selected
- [x] Support multiple file uploading in azure blob storage

**_Functionality around submission and validation will be added in a follow-up_**


admin/controllers/check-form-v2.js
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Upload check form view.
+ * @param req
+ * @param res
+ * @param next
+ * @returns {Promise.<void>}
+ */
+controller.uploadCheckFormPage = async (req, res, next) => {
+  req.breadcrumbs('Upload and view forms', '/check-form-v2/upload-and-view-forms-v2')
```
jon-shipley: I'm not sure about exposing 'v2' in the url.

admin/controllers/check-form-v2.js
```diff
@@ -0,0 +1,64 @@
...
+ * Submit check form.
+ * @param req
+ * @param res
+ * @param next
+ * @returns {Promise.<*>}
+ */
+controller.submitCheckForm = async (req, res, next) => {
+  const uploadedFile = req.files && req.files.csvFile
+  try {
+    await checkFormV2Service.submit(uploadedFile)
```
jon-shipley: Surprising that user information is not also required in the service.

admin/routes/check-form-v2.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const express = require('express')
+const router = express.Router()
+const isAuthenticated = require('../authentication/middleware')
+const rolesConfig = require('../roles-config')
+const checkFormV2Controller = require('../controllers/check-form-v2')
+const featureToggles = require('feature-toggles')
+
+if (featureToggles.isFeatureEnabled('newCheckForm')) {
```
jon-shipley: I don't think we should change to these new urls at all.  Surely there is a better way?

admin/controllers/check-form-v2.js
GeorgeChatzigiannis: @jon-shipley this is just a placeholder and will change over the following commits.
The implementation is currently blank.

admin/controllers/check-form-v2.js
GuyHarwood: @GeorgeChatzigiannis we don't need both old and new URL's available at runtime, so just add the toggle at the router level

admin/routes/check-form.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const express = require('express')
+const router = express.Router()
+const isAuthenticated = require('../authentication/middleware')
+const rolesConfig = require('../roles-config')
+const checkFormV2Controller = require('../controllers/check-form-v2')
+const featureToggles = require('feature-toggles')
+
```
GuyHarwood: as jon mentioned - why do we need new URLs.  the toggle can exist at the routing level to determine which implementation should be used, with no change on the app surface.

admin/controllers/check-form-v2.js
GeorgeChatzigiannis: @GuyHarwood the toggle is already added at route level similar to other implementations.

admin/routes/check-form-v2.js
GeorgeChatzigiannis: @jon-shipley the new urls do not include `v2` anymore as you pointed out. If there's any suggestion for improvements or a better workaround here please let me know.

admin/app.js
```diff
@@ -124,7 +125,8 @@ busboy.extend(app, {
 
 const allowedPath = (url) => (/^\/pupil-register\/pupil\/add-batch-pupils$/).test(url) ||
   (/^\/test-developer\/upload-new-form$/).test(url) ||
-  (/^\/service-manager\/upload-pupil-census\/upload$/).test(url)
+  (/^\/service-manager\/upload-pupil-census\/upload$/).test(url) ||
+  (/^\/check-form\/submit-check-form$/).test(url)
```
jon-shipley: What do you think of making the URL ''/check-form/upload'?  I dislike the generic 'submit', as it says little.

admin/controllers/check-form-v2.js
```diff
@@ -0,0 +1,65 @@
...
+controller.uploadCheckFormPage = async (req, res, next) => {
+  req.breadcrumbs('Upload and view forms', '/check-form/upload-and-view-forms')
+  res.locals.pageTitle = 'Upload new form'
+  let error
+  let formData
+  try {
+    req.breadcrumbs(res.locals.pageTitle)
+    res.render('check-form/upload-new-form', {
+      breadcrumbs: req.breadcrumbs(),
+      error: error || new ValidationError(),
```
jon-shipley: It looks to me like `error` is always undefined?

admin/controllers/check-form-v2.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param res
+ * @param next
+ * @returns {Promise.<*>}
+ */
+controller.submitCheckForm = async (req, res, next) => {
+  // This is placeholder method
+  const uploadedFile = req.files && req.files.csvFile
+  const requestData = req.body
+  try {
+    await checkFormV2Service.process(uploadedFile, requestData)
```
jon-shipley: `process` seems generic.  What does this service do?

admin/routes/check-form.js
```diff
@@ -0,0 +1,17 @@
...
+if (featureToggles.isFeatureEnabled('newCheckForm')) {
+  /* Check Form v2 routing */
+  router.get('/upload-and-view-forms', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormV2Controller.uploadAndViewFormsPage(req, res, next))
+  router.get('/upload-new-form', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormV2Controller.uploadCheckFormPage(req, res, next))
+  router.post('/submit-check-form', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormV2Controller.submitCheckForm(req, res, next))
```
jon-shipley: I am not sure about the 'submit' part of the URL.

admin/routes/check-form.js
```diff
@@ -0,0 +1,17 @@
...
+const router = express.Router()
+const isAuthenticated = require('../authentication/middleware')
+const rolesConfig = require('../roles-config')
+const checkFormV2Controller = require('../controllers/check-form-v2')
+const featureToggles = require('feature-toggles')
+
+if (featureToggles.isFeatureEnabled('newCheckForm')) {
+  /* Check Form v2 routing */
+  router.get('/upload-and-view-forms', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormV2Controller.uploadAndViewFormsPage(req, res, next))
+  router.get('/upload-new-form', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormV2Controller.uploadCheckFormPage(req, res, next))
```
jon-shipley: This controller method would make more sense to me if it was called getUploadPage().  In this context "upload" seems to be a verb which sounds like it ought to be doing something.

@GuyHarwood  - do we have guidance on controller names?

admin/controllers/check-form-v2.js
GeorgeChatzigiannis: @jon-shipley placeholder method that could potentially be renamed on the follow up

admin/app.js
GeorgeChatzigiannis: @jon-shipley url changed to the suggested one.

admin/controllers/check-form-v2.js
GeorgeChatzigiannis: @jon-shipley error is now deriving from the arguments and formData from the req object

admin/routes/check-form.js
GeorgeChatzigiannis: @jon-shipley changed to upload as discussed

admin/controllers/check-form-v2.js
GeorgeChatzigiannis: @jon-shipley this method is now removed and will be introduced at a later stage.

admin/routes/check-form.js
GeorgeChatzigiannis: @jon-shipley routes have been changed based on discussions.

admin/controllers/check-form.js
```diff
@@ -22,10 +24,12 @@ const winston = require('winston')
  */
 const getTestDeveloperHomePage = async (req, res, next) => {
   res.locals.pageTitle = 'MTC for test development'
+  const isNewCheckForm = featureToggles.isFeatureEnabled('newCheckForm')
   try {
     req.breadcrumbs(res.locals.pageTitle)
     res.render('test-developer/test-developer-home', {
-      breadcrumbs: ''
+      breadcrumbs: '',
+      isNewCheckForm
```
jon-shipley: It's not clear from the name that this `isNewForm` is a feature toggle.  What do you think of `isNewCheckFormFeatureToggleEnabled`?  I am open to other suggestions.

admin/views/check-form/upload-new-forms.ejs
```diff
@@ -0,0 +1,122 @@
...
+        <h2 class="heading-medium top-padding-30"><a name="csv-guidance">Download example</a></h2>
+
+        <ol class="numeric-list">
+            <li>
+                Download the example
+                <div class="panel panel-border-wide short">
+                    <a href="<%= assetPath %>csv/check-form-sample.csv" class="heading-small icon-download">Check form example (CSV, 4KB)</a>
+                </div>
+            </li>
+            <li>Open file in a programme like Excel, OpenOffice or Google Sheets</li>
```
jon-shipley: This may be the actual business wording, but how many of our users would call Excel a 'programme'? 👀

admin/controllers/check-form.js
```diff
@@ -22,10 +24,12 @@ const winston = require('winston')
  */
 const getTestDeveloperHomePage = async (req, res, next) => {
   res.locals.pageTitle = 'MTC for test development'
+  const isNewCheckFormFeatureToggleEnabled = featureToggles.isFeatureEnabled('newCheckForm')
```
GuyHarwood: as jon mentioned - these need to be well named.

admin/lib/azure-upload.js
```diff
@@ -1,42 +1,31 @@
...
 
 const path = require('path')
-const azure = require('azure-storage')
 const moment = require('moment')
-const blobService = azure.createBlobService()
-const winston = require('winston')
+const blobService = require('../services/data-access/azure-blob.data.service')
 
 // Files get uploaded to this container.  dns naming conventions.
 const container = 'check-development-app-upload-files'
```
GuyHarwood: can you rename this to something more appropriate while you are working 'in this area' please.   always bugged me.  
maybe `admin-app-uploaded-files` ?

admin/services/data-access/azure-blob.data.service.js
```diff
@@ -9,18 +10,21 @@ if (config.AZURE_STORAGE_CONNECTION_STRING) {
   blobService = azureStorage.createBlobService()
 }
 
-const service = {
-  getBlobText: async (containerName, blobName) => {
+bluebird.promisifyAll(blobService, {
```
GuyHarwood: please thoroughly test this against your own storage account to ensure it works ok.  as it affects all functions on the library

admin/lib/azure-upload.js
GeorgeChatzigiannis: @GuyHarwood done

admin/services/data-access/azure-blob.data.service.js
GeorgeChatzigiannis: @GuyHarwood done against my own storage account and verified it is working as expected.

admin/controllers/check-form.js
GeorgeChatzigiannis: @jon-shipley updated as suggested.

admin/views/check-form/upload-new-forms.ejs
GeorgeChatzigiannis: @jon-shipley 😆

GuyHarwood: jokes aside - probably best to report this to the BAs

GeorgeChatzigiannis: @GuyHarwood reported already and I was told there is a content editor coming in to update the content.

PR #926, opened 2018-12-04
load test prep for usage with flood

GuyHarwood:
- [x] update readme
- [x] remove hardcoded paths
- [x] give hostname variables meaningful names


load-test/notes/pupil id extraction.txt
```diff
@@ -0,0 +1,35 @@
+Form POST
```
jon-shipley: We shouldn't have space in a filename.

PR #929, opened 2018-12-07
Feature/27352 update node modules packages

jon-shipley:
admin app: major and minor version updates


GuyHarwood:
signing off with known vulnerability, as this will be addressed in a future PR.
container validation actually passed, but never reported back from AzureDevOps 🤷‍♂️

PR #938, opened 2018-12-12
update app insights to 1.0.8

GuyHarwood:
[apparently prevents wrapping](https://github.com/Microsoft/ApplicationInsights-node.js/pull/464#issuecomment-446373231) of stack traces with the 'zone error' app insights stack


GuyHarwood:
@GeorgeChatzigiannis no.  read the package.json docs on what the caret signifies.

PR #941, opened 2018-12-12
Bug/27484 Migrations exit code

danwilliams:
Removed unnecessary try/catch and ensured a non 0 exit code is returned on error


GuyHarwood:
@danwilliams the broken view migration that you fixed in another PR caused this PR validation to fail during migrating down, which proves that this PR fixes the issue 😄 

This can be merged as soon as the validation container & migration check is green 👍

PR #944, opened 2018-12-13
Feature/27445 service manager check settings

danwilliams:


admin/config.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
   RESTART_MAX_ATTEMPTS: 2,
   SESSION_SECRET: process.env.NODE_ENV === 'production' ? process.env.SESSION_SECRET : 'anti tamper for dev',
   TIME_BETWEEN_QUESTIONS: 3,
+  LENGTH_OF_CHECK: 30,
```
GuyHarwood: please make the unit explicit, ie minutes/seconds

admin/lib/validator/settings-validator.js
```diff
@@ -18,6 +18,13 @@ const settingsValidationSchema = {
       options: [{ min: 1, max: 5 }] // Max number to be confirmed
     },
     errorMessage: settingsErrorMessages.loadingTimeLimit
+  },
```
GeorgeChatzigiannis: This is probably the only place along with the old HDF validator where we are still following this validation schema pattern which seems very outdated.
Why not rewrite it in the same approach that the other validators for consistency?
I'd expect it to be simple and quite cleaner.
@GuyHarwood @jon-shipley thoughts on that?

jon-shipley: @GeorgeChatzigiannis @GuyHarwood : It needs to be re-written and express-validator removed.   There are tickets for the remaining legacy validation to do just this.

admin/data/sql/migrations/20181212170211.do.add-check-time-limit-settings.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE [mtc_admin].[settings] ADD checkTimeLimit TINYINT NOT NULL CONSTRAINT temp_default_check_limit DEFAULT 30;
```
jon-shipley: @GuyHarwood - do we have a naming convention for constraints?

admin/lib/validator/settings-validator.js
GuyHarwood: @danwilliams as per discussion - please switch to new validator

PR #950, opened 2018-12-19
Feature/27594 send all pupils as one preparecheck message

jon-shipley:
Performance improvement

* Add batch message capabilities to `pupil-status` and `prepare-check`
* Reduce network round-trip latency when creating several checks in one go by sending up to 20 prepare-check messages and pupil-status change messages in a single batch message format.


functions/pupil-status/v1.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const pupilStatusService = require('./pupil-status.service')
+
+async function process (context, pupilStatusMessage) {
+  if (!pupilStatusMessage.hasOwnProperty('pupilId')) {
```
GeorgeChatzigiannis: @jon-shipley isn't that essentially a validator?
I did extract validation separately into a file on some other functions.
Perhaps the same approach should be followed for consistency?

functions/pupil-status/v2.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+const pupilStatusBatchService = require('./pupil-status-batch.service')
+
+function validate (message) {
+  if (!Array.isArray(message.messages)) {
+    throw new Error('pupil-status: invalid version:2 message.  `messages` should be an Array.')
+  }
+  message.messages.forEach(msg => {
+    if (!msg.pupilId) {
```
GeorgeChatzigiannis: Same comment around validation here as well.

functions/pupil-status/v1.js
GuyHarwood: @GeorgeChatzigiannis This is a lift & shift of existing code to prove a performance driven change.  I would rather not change too much in this PR

functions/check-started/index.js
```diff
@@ -129,5 +129,5 @@ function canCheckBeStarted (code) {
 async function requestPupilStatusChange (pupilId, checkCode) {
   const pupilStatusQueueName = 'pupil-status'
   const message = { version: 1, pupilId, checkCode }
-  return await azureStorageHelper.addMessageToQueue(pupilStatusQueueName, message)
+  return azureStorageHelper.addMessageToQueue(pupilStatusQueueName, message)
```
GuyHarwood: Why remove the await?

functions/prepare-check/index.js
```diff
@@ -1,58 +1,64 @@
...
-const azure = require('azure-storage')
-const azureStorageHelper = require('../lib/azure-storage-helper')
-const entGen = azure.TableUtilities.entityGenerator
-const moment = require('moment')
+'use strict'
+
+const { performance } = require('perf_hooks')
 const uuid = require('uuid/v4')
 
+const v1 = require('./v1.js')
```
GuyHarwood: 👍🏻

functions/prepare-check/prepare-entity.js
```diff
@@ -0,0 +1,34 @@
...
+ */
```
GuyHarwood: Urgh!

functions/check-started/index.js
jon-shipley: It is an async function so any callers would already `await` on the Promise.  There is no need to await internally and for the caller to also await - it just takes extra time.  A single await is enough.

functions/pupil-status/v1.js
GeorgeChatzigiannis: @GuyHarwood it would make the validation isolated and easily testable which is the main reason why I did it on other functions.
It can certainly happen after this PR is merged.

functions/prepare-check/prepare-entity.js
jon-shipley: I'm guessing that this is auto-generated.

It's azure... if you want to eat Rhino you have to be prepared to pay freight as someone said

PR #951, opened 2018-12-19
Bug/27545 pupil not being highlighted when a restart has been applied

jon-shipley:
* add a missing highlight wrapper class
* send pupil urlSlugs through to the controller rather than restart ids as the page is using pupil `urlSlugs` for item identification

screen shot 2018-12-19 at 19 42 05

### DIFF
--- a/admin/services/restart.service.js
+++ b/admin/services/restart.service.js
@@ -99,7 +99,7 @@ restartService.restart = async (
     return pupilRestartDataService.sqlCreate(pupilRestartData)
   }))
   const pupilInfo = await pupilDataService.sqlFindByIds(pupilsList, schoolId)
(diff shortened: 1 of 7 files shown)


admin/services/restart.service.js
```diff
@@ -86,7 +86,7 @@ restartService.restart = async (
     throw new Error(`One of the pupils is not eligible for a restart`)
   }
   const restartReasonId = await pupilRestartDataService.sqlFindRestartReasonByCode(restartReasonCode)
-  return Promise.all(pupilsList.map(async pupilId => {
+  await Promise.all(pupilsList.map(async pupilId => {
     const pupilRestartData = {
```
GeorgeChatzigiannis: Outside of the scope of this, just noticed this doesn't looks quite good.
We'll need to refactor this into a multiple insertion statement.

PR #955, opened 2018-12-21
Redis as an express session provider

GuyHarwood:
SQL Server is becoming a bottleneck at high load.

This moves the session store dependency away to redis.

Merging to load test support branch, *not* master, for load testing.

### DIFF
--- a/admin/app.js
+++ b/admin/app.js
@@ -32,7 +32,6 @@ const R = require('ramda')
 const session = require('express-session')
 const setupBrowserSecurity = require('./helpers/browserSecurity')
 const setupLogging = require('./helpers/logger')
-const TediousSessionStore = require('connect-tedious')(session)
 const uuidV4 = require('uuid/v4')
 
 const logger = require('./services/log.service').getLogger()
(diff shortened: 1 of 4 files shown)


admin/app.js
```diff
@@ -149,6 +155,20 @@ const sessionOptions = {
...
+const sessionOptions = {
+  name: 'mtc-admin-session-id',
+  secret: config.SESSION_SECRET,
+  resave: false,
+  rolling: true,
+  saveUninitialized: false,
+  cookie: {
+    maxAge: config.ADMIN_SESSION_EXPIRATION_TIME_IN_SECONDS * 1000,
+    httpOnly: true,
+    secure: secureCookie
```
jon-shipley: We can only have the secure flag in the cookie on https, can't we?  Otherwise won't local dev break?

admin/app.js
```diff
@@ -122,18 +123,23 @@ if (process.env.NODE_ENV === 'production') {
...
+    auth_pass: config.Redis.Key,
+    tls: {
+      servername: config.Redis.Host
+    }
+  })
+  sessionStore = new RedisStore({
+    client: client
+  })
+} else {
+  const TediousSessionStore = require('connect-tedious')(session)
```
jon-shipley: Nice to see the fallback in there - can be used in local dev

PR #958, opened 2018-12-21
Bring redis session store support into master

GuyHarwood:
This has been load tested and works as expected.


GuyHarwood:
@jon-shipley resultsTree.csv has already been deleted.  its jmeter output that was include accidentaly.  see files list - its deleted in source branch

PR #970, opened 2019-01-14
Bug/27121 Throw bulk import error instead of displaying it in table

GeorgeChatzigiannis:
Throw the error instead of displaying in the result column


admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -22,7 +22,7 @@ pupilCensusImportDataService.sqlBulkImport = async (pupilData, schools, jobId) =
   try {
     result.output = await bulkLoadData(con, pupilData, schools, jobId)
   } catch (error) {
-    result.errorOutput = error
+    result.errorOutput = error && error.message
```
GuyHarwood: if this works, then great.  but i think you could make this more robust because if error is undefined you will have no idea what has happened.  i would create a function, add some object checks to it, try and parse it etc and definitely return something useful if the error object is empty or null to say 'object undefined' or 'object has no error information'

GeorgeChatzigiannis: @GuyHarwood this has been updated based on your feedback.

GuyHarwood: @GeorgeChatzigiannis could we not be more explicit and call toString on the object if its defined?  This seems a little obtuse to me

GeorgeChatzigiannis: @GuyHarwood certainly, it's updated.

admin/services/data-access/pupil-census-import.data.service.js
```diff
@@ -69,4 +77,25 @@ const bulkLoadData = (connection, pupilData, schools, jobId) => {
...
+  if (typeof error !== 'object') {
+    return `Error type is ${typeof error} with value ${error.toString()}`
+  }
+  if (error || Object.keys(error).length === 0) {
+    return `Error with no properties: ${error.toString()}`
+  }
+  if (error || !error.message) {
+    return `Error with no message property: ${error.toString()} `
+  }
+  return error && error.message.toString()
```
GuyHarwood: this is much better.  i would just modify the last line, and call toString on the error object, rather than message.  as we know message is always a string anyway.

GeorgeChatzigiannis: @GuyHarwood done

PR #974, opened 2019-01-17
Bug/28351 pupil restart ids not being found

jon-shipley:
Fix bug where selected pupils were being issued a restart and the message 'Pupil Ids not found' was appearing


GeorgeChatzigiannis:
@jon-shipley approved and thanks for updating the readme!

PR #976, opened 2019-01-18
Feature/27838 Assign check forms v2

GeorgeChatzigiannis:
- [x] Display assigned check forms
- [x] Submit newly assigned check forms
- [x] Message highlighting
- [x] Read only form page when relevant check window period is in play


admin/services/data-access/check-form-v2.data.service.js
```diff
@@ -197,6 +212,92 @@ const checkFormV2DataService = {
...
+      })
+    })
+    const insertSql = `INSERT INTO ${sqlService.adminSchema}.[checkFormWindow] (
+    checkForm_id,
+    checkWindow_id
+    ) VALUES`
+
+    queries.push([insertSql, inserts.join(', \n')].join(' '))
+    const sql = queries.join('\n')
+    return sqlService.modify(sql, params)
```
jon-shipley: It would be handy to wrap this DELETE+INSERT in a transaction to prevent errors leading to an unknown state.

admin/views/check-form/view-select-forms.ejs
```diff
@@ -40,34 +41,38 @@
...
                      aria-checked="false"
+                     <% if (checkForm.checked) { %>checked="true"<% } %>
                      role="checkbox">
               <div></div>
             </div>
           </td>
         </tr>
+        <% } %>
         <% }) %>
+        <input type="hidden" name="hasAssignedForms" value="<%= hasAssignedForms %>" />
```
jon-shipley: I wonder why these two hidden variables are required?

GeorgeChatzigiannis: Both are utilised by the post method.
- `hasAssignedForms` is a convenient way to avoid performing deletion if no check forms have been assigned to the check window at the given state.
- `checkWindowName` is being used to display the name of the check window on the green notification bar. I've done this to avoid an additional data service call in the post method to fetch just that information.

admin/helpers/check-form-presenter.js
```diff
@@ -86,15 +86,51 @@ checkFormPresenter.getPresentationCheckWindowListData = (checkWindows) => {
  * @returns {Object}
  */
 checkFormPresenter.getPresentationCheckWindowData = (checkWindow, checkFormType) => {
+  const checkTypeStartDate = checkFormType === 'live' ? checkWindow.checkStartDate : checkWindow.familiarisationCheckStartDate
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley Problem with `checkStartDate` is that the check window table already has that name allocated for the live check start date.
Check type is either live or familiarisation. Thought this was clear enough?
How about `startDateByType` ?

admin/helpers/check-form-presenter.js
```diff
@@ -86,15 +86,51 @@ checkFormPresenter.getPresentationCheckWindowListData = (checkWindows) => {
...
     name: checkWindow.name,
     urlSlug: checkWindow.urlSlug,
     familiarisationCheckStartDate: dateService.formatFullGdsDate(checkWindow.familiarisationCheckStartDate),
     familiarisationCheckEndDate: dateService.formatFullGdsDate(checkWindow.familiarisationCheckEndDate),
     checkStartDate: dateService.formatFullGdsDate(checkWindow.checkStartDate),
     checkEndDate: dateService.formatFullGdsDate(checkWindow.checkEndDate),
-    checkFormTypeTitle: checkFormType === 'live' ? 'Multiplication tables check' : 'Try it out'
+    checkFormTypeTitle: checkFormType === 'live' ? 'Multiplication tables check' : 'Try it out',
+    checkPeriod: checkFormType === 'live' ? 'MTC' : 'Try it out',
+    isWithinCheckType: dateService.utcNowAsMoment().isAfter(checkTypeStartDate) && dateService.utcNowAsMoment().isBefore(checkTypeEndDate)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: How did I miss that.
Great point, already updated.

admin/helpers/check-form-presenter.js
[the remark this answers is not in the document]
jon-shipley: `const checkStartDate = ...`  - I mean the variable, not the DB field name.

admin/services/data-access/check-form-v2.data.service.js
GeorgeChatzigiannis: @jon-shipley updated based on discussion.

admin/helpers/check-form-presenter.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley yea I was aware. I updated this as well as the returned property names.

admin/views/check-form/view-select-forms.ejs
jon-shipley: If you are using a value from the GUI to determine whether or not you need to run a delete operation, then that's bad as the GUI is entirely untrusted.  Instead, either always run the delete or determine at run time.

In my view, the `checkWindowName` should also be looked up, rather than passed in through the web request.  I don't see any issue with that.

GeorgeChatzigiannis:
@jon-shipley updated based on our chat

admin/data/sql/seeds/20190123182010.custom.familiarisation-check-form-dev-data.sql
```diff
@@ -0,0 +1,30 @@
...
+        {
+            "f1" : 5,
+            "f2" : 6
+        },
+        {
+            "f1" : 10,
+            "f2" : 4
+        },
+        {
+            "f1" : 7,
```
jon-shipley: This is a minor non-blocking comment, but as the 8 and 9 times tables are generally considered the most difficult AIUI perhaps we should just use up something very easy for the developer version of the familiarisation check.  e.g. 2 times table or 5 or 10 times table?

GuyHarwood: i assumed this familiarisation form had come from STA, or did you define it @GeorgeChatzigiannis ?

GeorgeChatzigiannis: @GuyHarwood missed that message for some reason. I defined it myself.

GeorgeChatzigiannis: I'll request one that is STA approved.

GeorgeChatzigiannis: @jon-shipley @GuyHarwood Apparently there is no way at this stage to acquire a familiarisation form from STA. I did update the one that is used with data from table 1 to 6.

PR #980, opened 2019-01-22
feature/28706 update HSTS headers

GuyHarwood:
add `preload` and `includeSubdomains` attributes to the [HSTS security header](https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet)

instead of 
`Strict-Transport-Security: max-age=31536000;` 
it is now...
`Strict-Transport-Security: max-age=31536000; includeSubDomains; preload`


GuyHarwood:
@activemq please can we update our security tests to reflect this change

PR #984, opened 2019-01-23
Feature/28217 optimise config calls in check preparation

jon-shipley:
# Performance and code cleanup

Retrieve pupil check config more efficiently from the database by using a single SQL call for all pupils, rather one call per pupil.


admin/services/data-access/config.data.service.js
```diff
@@ -0,0 +1,53 @@
...
+          s.id as schoolId,
+      
+          -- global question time settings
+          st.loadingTimeLimit as loadingTime,
+          st.questionTimeLimit as questionTime,
+          st.checkTimeLimit as checkTime,
+      
+          -- pupil configs
+          p.speechSynthesis,
+          STRING_AGG (aa.code, ',') as accessArrangementCodes,
```
GuyHarwood: nice 💯  wasn't aware of this function

admin/spec/back-end/service/check-start.service.spec.js
```diff
@@ -206,9 +207,15 @@ describe('check-start.service', () => {
       )
       spyOn(pupilDataService, 'sqlUpdateTokensBatch').and.returnValue(Promise.resolve())
       spyOn(checkStartService, 'prepareCheckQueueMessages').and.returnValue(mockPreparedCheckQueueMessages)
-      spyOn(azureQueueService, 'addMessageAsync')
+      spyOn(azureQueueService, 'addMessageAsync').and.returnValue(Promise.resolve())
```
GeorgeChatzigiannis: Minor comment: even if you don't return a value here it will still work for the spy.

jon-shipley: Reverted

admin/services/config.service.js
```diff
@@ -87,7 +87,135 @@ const configService = {
...
+          break
+      }
+    })
+
+    return config
+  },
+
+  /**
+   * Generate a config object for a batch of pupils
+   * @param configDataByPupil
```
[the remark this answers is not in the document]
jon-shipley: updated

admin/app.js
```diff
@@ -66,10 +66,9 @@ const index = require('./routes/index')
 const testDeveloper = require('./routes/test-developer')
 const serviceManager = require('./routes/service-manager')
 const school = require('./routes/school')
-const questions = require('./routes/questions')
-const pupilFeedback = require('./routes/pupil-feedback')
-const checkStarted = require('./routes/check-started')
-const completedCheck = require('./routes/completed-check')
+// const pupilFeedback = require('./routes/pupil-feedback')
```
GeorgeChatzigiannis: Can we remove the comments? Any particular reason for keeping them?

admin/app.js
```diff
@@ -236,10 +237,9 @@ app.use(function (req, res, next) {
   next()
 })
 
-app.use('/api/questions', questions)
-app.use('/api/pupil-feedback', pupilFeedback)
-app.use('/api/completed-check', completedCheck)
-app.use('/api/check-started', checkStarted)
+// app.use('/api/pupil-feedback', pupilFeedback)
+// app.use('/api/completed-check', completedCheck)
```
GeorgeChatzigiannis: @jon-shipley extra commented lines here as well.

admin/services/config.service.js
```diff
@@ -4,90 +4,147 @@ const R = require('ramda')
...
 /** @namespace */
 
 const configService = {
+  getBaseConfig: function getBaseConfig () {
+    return {
+      audibleSounds: false,
+      checkTime: undefined,
+      colourContrast: false,
+      colourContrastCode: undefined,
+      fontSize: false,
```
GeorgeChatzigiannis: @jon-shipley missing property fontSizeCode with value undefined

admin/app.js
jon-shipley: I've commented them out as these `api` calls should no longer be used, therefore better security not to expose them.

We should raise a PR soon to delete this, an any other, dead code.

admin/app.js
jon-shipley: Same as above.

admin/services/config.service.js
jon-shipley: Fixed, thanks

PR #996, opened 2019-01-29
Feature/23280 staff submit hdf

danwilliams:
Third HDF ticket out of four, adds in confirmation and submission.


admin/assets/stylesheets/elements/_hdf.scss
```diff
@@ -0,0 +1,21 @@
+// Smaller inset checkboxes
+.multiple-choice.inset {
+  margin-left: 50px !important;
+
+  input[type=checkbox] + label {
```
jon-shipley: I am not sure why we would need custom css, why doesn't the standard markup cover the design?
https://govuk-elements.herokuapp.com/form-elements/

admin/views/hdf/attendance-edit-reason.ejs
```diff
@@ -7,64 +7,58 @@
...
+        </div>
 
-            <br />
+        <br />
 
-            <% if (attendanceCodes && attendanceCodes.length > 1) { %>
-            <h3>Select reason</h3>
+        <form method="post" action="/attendance/submit-edit-reason">
+            <input type="hidden" name="_csrf" value="<%= csrftoken %>" />
+            <input type="hidden" name="pupilId" value="<%= pupil.id %>" />
```
jon-shipley: We should use `pupil.urlSlug` and not the `id` to prevent pupil id enumeration.

admin/views/hdf/attendance-edit-reason.ejs
```diff
@@ -7,64 +7,58 @@
...
-            <a href="#top">Back to top</a>
-        </div>
     </div>
 </div>
-    
\ No newline at end of file
+
+<div class="back-to-top-container">
+    <a href="#top">Back to top</a>
+</div>
```
jon-shipley: Missing newline

admin/views/hdf/confirm-and-submit.ejs
```diff
@@ -0,0 +1,82 @@
...
+            <h1 class="heading-xlarge">Confirm and submit</h1>
+
+            <%- partial('partials/_session_expiry') %>
+
+            <div class="form-group <%= error.isError('confirmBoxes') ? 'form-group-error' : '' %>" id="confirm">
+                <div class="multiple-choice">
+                    <input id="confirmYes" type="radio" role="radio" name="confirm" value="Y" checked="checked">
+                    <label for="confirmYes">
+                        I <strong>confirm</strong> that my school administered the multiplication tables check according to the arrangements set out in the guidance.
+                        <br /><br />
```
jon-shipley: Probably should be done in css rather than `br` tags

admin/lib/validator/hdf-confirm-validator.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports.validate = function (hdfConfirmData) {
+  const validationError = new ValidationError()
+  const {
+    confirm,
+    pupilDetails,
+    uniquePins,
+    staffConfirm
+  } = hdfConfirmData
+
+  if (confirm === 'Y') {
```
GeorgeChatzigiannis: @danwilliams Can you avoid the nested if by returning the validation error if confirm !== 'Y' ?
This way the code readability is better.

admin/assets/stylesheets/elements/_hdf.scss
danwilliams: Needed to add styles for a inset list of checkboxes because the standard gov uk elements don’t have any styles for it. Also, the checkboxes on the prototype are smaller than the standard gov uk size.

jon-shipley: @danwilliams - I question both of these statements.  Why are we using non-standard sizes?   We should be using standard sizes, surely?  Perhaps we should discuss?

jon-shipley: Okay - I've seen the prototype and now understand that GDS does not have a pattern for inset radio boxes.

PR #1002, opened 2019-01-30
Feature/28374 refactor ps report to support hpa mode for feb trial

jon-shipley:
Update the PS & anomaly reports to work with the new data location (checkResult.payload)


admin/services/data-access/completed-check.data.service.js
```diff
@@ -84,20 +91,28 @@ completedCheckDataService.sqlFindByIds = async (batchIds) => {
...
+    if (!x.payload) {
       return R.clone(x)
     }
-    const d = JSON.parse(x.data)
-    return R.assoc('data', d.data, x)
+    const transformations = {
+      formData: JSON.parse,
+      payload: JSON.parse
+    }
+    const d1 = R.evolve(transformations, x)
```
GeorgeChatzigiannis: nice 💯

admin/bin/psychometricians-report-v2.js
```diff
@@ -23,41 +21,39 @@ const options = commandLineArgs(optionDefinitions)
 
 async function main (options) {
   try {
+    logger.info('force detected: re-processing all checks')
     if (options.forceReprocess) {
       // force the report to re-calculate the cached ps-report
-      winston.info('force detected: re-processing all checks')
+      logger.info('force detected: re-processing all checks')
```
GuyHarwood: why log this twice?  surely just once after the detection of `options.forceReprocess`

jon-shipley: I suspect this was a bad merge or conflict resolution.  I've removed the line above that didn't make sense.

PR #1003, opened 2019-01-31
Add migration to exclude pupils who have a familiarisation pin

jon-shipley:


admin/data/sql/migrations/20190131044243.do.fix-fam-pin-generation-bug.sql
```diff
@@ -0,0 +1,49 @@
...
+    p.urlSlug
+  FROM
+    [mtc_admin].[pupil] p
+      LEFT JOIN [mtc_admin].[pupilAttendance] pa ON (p.id = pa.pupil_id and pa.isDeleted = 0)
+      LEFT JOIN [mtc_admin].[attendanceCode] ac ON (pa.attendanceCode_id = ac.id )
+      LEFT JOIN  [mtc_admin].[pupilGroup] g ON (g.pupil_id = p.id)
+      INNER JOIN [mtc_admin].[pupilStatus] ps ON (p.pupilStatus_id = ps.id)
+  WHERE
+    -- include all pupils except those who are marked as not taking check because they left school
+    (ac.id IS NULL OR ac.code <> 'LEFTT')
```
GeorgeChatzigiannis: @jon-shipley what about incorrect registration code? shall these be taken into account as well in a similar manner with the ones who left?

jon-shipley: @GeorgeChatzigiannis - this is the spec, so no, they shouldn't.

PR #1009, opened 2019-02-01
Bug/type checking issue on check form lines length comparison

GeorgeChatzigiannis:


admin/config.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
   CHECK_FORM_NAME_MAX_CHARACTERS: 128,
   CHECK_FORM_MIN_INTEGER: 1,
   CHECK_FORM_MAX_INTEGER: 12,
-  LINES_PER_CHECK_FORM: process.env.LINES_PER_CHECK_FORM || 25,
+  LINES_PER_CHECK_FORM: process.env.LINES_PER_CHECK_FORM || '25',
```
GuyHarwood: why are you making this into a string?

admin/lib/validator/check-form/single-check-form-validator.js
```diff
@@ -40,7 +40,7 @@ singleCheckFormValidator.validate = async (uploadedFile) => {
   const fileContent = fileBuffer && fileBuffer.toString().trim()
   const fileLines = fileContent && fileContent.split('\n').length
   // Invalid total file lines
-  if (fileLines !== config.LINES_PER_CHECK_FORM) {
+  if (fileLines !== parseInt(config.LINES_PER_CHECK_FORM)) {
```
GuyHarwood: can you please parse the value on a separate line, and deal with the fact that it could return `NaN`

admin/services/check-form.service.js
```diff
@@ -297,7 +297,7 @@ const checkFormService = {
     result = csvData.toString().split('\n').map(function (line) {
       return line.trim()
     }).filter(Boolean)
-    return result.length === config.LINES_PER_CHECK_FORM
+    return result.length === parseInt(config.LINES_PER_CHECK_FORM)
```
GuyHarwood: i would consider moving the parseInt into the config file instead.  and return default value if parsing fails

admin/lib/validator/check-form/single-check-form-validator.js
jon-shipley: parseInt() should still be used with a  radix, because `parseInt("010")` can either be octal or decimal depending on the implementation, and it's always possible implementation can change.

```suggestion
  if (fileLines !== parseInt(config.LINES_PER_CHECK_FORM, 10)) {
```

Same for all the other uses.

admin/config.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
   CHECK_FORM_NAME_MAX_CHARACTERS: 128,
   CHECK_FORM_MIN_INTEGER: 1,
   CHECK_FORM_MAX_INTEGER: 12,
-  LINES_PER_CHECK_FORM: process.env.LINES_PER_CHECK_FORM || '25',
+  LINES_PER_CHECK_FORM: parseInt(process.env.LINES_PER_CHECK_FORM) || 25,
```
GuyHarwood: this will never return the default, because parseInt returns the parsed integer or `NaN`

GuyHarwood: `function getFormLines() {
    const defaultValue = 25
    const val = process.env.LINES_PER_CHECK_FORM
    const parsed = parseInt(val)
    if (parsed === 'NaN') return defaultValue
    if (parsed < 1) return defaultValue
    return parsed
}`

GeorgeChatzigiannis:
@GuyHarwood @jon-shipley updated

PR #1013, opened 2019-02-01
Feature/27642 view hdf submission

danwilliams:
Final HDF ticket, bringing everything together and integrating with other recently-merged features.


admin/controllers/attendance.js
```diff
@@ -198,13 +198,46 @@ const postConfirmSubmit = async (req, res, next) => {
...
+  }
+
+  // Re-check hdf eligibility
+  let hdfEligibility
+  try {
+    hdfEligibility = await headteacherDeclarationService.getEligibilityForSchool(req.user.School)
+  } catch (error) {
+    return next(error)
+  }
+  if (!hdfEligibility) {
```
GeorgeChatzigiannis: @danwilliams why not include that within the headteacherDeclarationService getEligibilityForSchool and throw an error if ! hdfEligibility within that method? This seems to me a behavior that should be part of the service rather than the controller.

admin/controllers/attendance.js
```diff
@@ -243,16 +276,45 @@ const postDeclarationForm = async (req, res, next) => {
...
-  res.locals.pageTitle = 'Headteacher\'s declaration form submitted'
+  res.locals.pageTitle = "Headteacher's declaration form"
+  req.breadcrumbs(res.locals.pageTitle)
+  try {
+    const hdf = await headteacherDeclarationService.findLatestHdfForSchool(req.user.School)
+    if (!hdf) {
+      return res.redirect('/attendance/declaration-form')
+    }
+    const resultsDate = hdf.checkEndDate.add(1, 'weeks').isoWeekday('Monday') // next monday after the check ends
+    const canViewResults = moment().isAfter(resultsDate)
```
GeorgeChatzigiannis: @danwilliams I am aware this is a two-liner but it concerns the presentation level rather than the controller so I would suggest adding a presenter. Thoughts?

admin/controllers/attendance.js
GuyHarwood: is it perfectly OK to ask for eligibility and receive nothing though?  throwing error just because no data exists seems a bit extreme to me? 🤔

GeorgeChatzigiannis: Sounds like this should have been a validation error that the view would present something relevant then?

admin/helpers/hdf-presenter.js
```diff
@@ -0,0 +1,14 @@
...
+const moment = require('moment')
+
+const hdfPresenter = {}
+
+hdfPresenter.getResultsDate = (hdf) => {
+  return moment(hdf.checkEndDate).add(1, 'weeks').isoWeekday('Monday') // next monday after the check ends
+}
+
+hdfPresenter.getCanViewResults = (resultsDate) => {
+  return moment().isAfter(resultsDate)
```
GeorgeChatzigiannis: @danwilliams should be moment.utc()

admin/helpers/hdf-presenter.js
```diff
@@ -4,7 +4,7 @@ const moment = require('moment')
 const hdfPresenter = {}
 
 hdfPresenter.getResultsDate = (hdf) => {
-  return moment(hdf.checkEndDate).add(1, 'weeks').isoWeekday('Monday') // next monday after the check ends
+  return moment(hdf.checkEndDate).add(1, 'weeks').isoWeekday('Monday').set({ hour: 8, minutes: 0, seconds: 0 }) // next monday after the check ends
```
GeorgeChatzigiannis: @danwilliams does the checkEndDate need a moment wrapper?
I was under the impression it is already 'momentised' at that level.

PR #1015, opened 2019-02-06
Feature/29191 Score calculation

GeorgeChatzigiannis:
- [x] Added new time triggered function to address score calculation and database transmision
- [x] Introduced store procedure to handle score data refresh


functions/calculate-score/index.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const { performance } = require('perf_hooks')
+
+const v1 = require('./v1')
+
+module.exports = async function (context) {
+  const start = performance.now()
+  try {
+    await v1.process(context)
```
jon-shipley: Usually, we return some meta-data from the process.  This could be the number of scores processed, the number of checks processed, the number of pupils processed?  Potentially some interesting stuff.

I wouldn't mind seeing a relatively verbose message for this one in the log.

GeorgeChatzigiannis: @jon-shipley I doubt what you are sugggesting can be achieved since the store procedure will conclude at a later time from the function completion time.
Does this answer your question?

jon-shipley: @GeorgeChatzigiannis.  The SP can easily return some data to the client that would `await` on it.  We do exactly this for the check creation.   I'm happy to help.   Give me a shout when the SP is further along.

functions/calculate-score/function.json
```diff
@@ -0,0 +1,14 @@
+{
+  "disabled": false,
+  "bindings": [
+    {
+      "schedule": "0 0 21 * * *",
```
GuyHarwood: 9pm?  i know these are not valid CRON expressions, hence why i ask...

functions/calculate-score/index.js
GuyHarwood: @GeorgeChatzigiannis you can use explicit return values in stored procedures, or just do a SELECT.  will be easier to consume in tedious if you do a SELECT.

functions/lib/sql-helper.js
```diff
@@ -173,3 +174,33 @@ module.exports.sqlUpdateAnswersWithResults = async (checkId, answers) => {
   const sql = [insertSql, inserts.join(', \n')].join(' ')
   return sqlService.modify(sql, params)
 }
+
```
GuyHarwood: this 'helper' class is starting to turn into a dumping ground / god class.  we should be creating specific services for things like check calculation.  please move this into a dedicated 'checkWindowScore' service

jon-shipley: `sql-helper.js` is ONLY for shared functionality - unless the function is used in at least 2 functions please don't put stuff there.

GeorgeChatzigiannis: @GuyHarwood @jon-shipley was thinking about the same thing, totally agreed.
It is addressed on f8de44c

functions/calculate-score/function.json
GeorgeChatzigiannis: @GuyHarwood this serves as a placeholder since there is no direct requirement with regards to the time of the day where this function will be executed.
It can be decided before merging.

GuyHarwood: ok, but what does the expression mean?

GeorgeChatzigiannis: Means that it will execute `every day at 21:00:00`

admin/data/sql/migrations/20190207104758.do.create-sp-calculate-score.sql
```diff
@@ -0,0 +1,133 @@
...
+    DECLARE @checkWindowId INTEGER = @scoreCalculationCheckWindowId
+
+    DECLARE @checkWindowSchoolId INTEGER
+    DECLARE @checkWindowSchoolIds CURSOR
+
+    -- FETCH ALL SCHOOL IDS RELEVANT TO THE CHECK WINDOW
+    SET @checkWindowSchoolIds = CURSOR FOR
+      SELECT DISTINCT s.id
+      FROM mtc_admin.[check] chk
+        JOIN mtc_admin.pupil p
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley excellent point!
I have updated that query to fetch school ids that do have live checks only.

functions/calculate-score/index.js
GuyHarwood: with regard to sp completion and function completion - that is a rather scary statement and forward view on this piece of work.  needs discussion

admin/data/sql/migrations/20190210134733.do.create-view-schools-average.sql
```diff
@@ -0,0 +1,31 @@
...
+
+SELECT
+  p.school_id,
+  CAST(SUM(
+      CASE
+      WHEN (pa.id IS NOT NULL) THEN 0
+      WHEN (latestPupilCheck.mark IS NOT NULL) THEN latestPupilCheck.mark
+      ELSE 0
+      END
+  ) AS DECIMAL(5,2)) / COUNT(p.id) AS [schoolAverage]
```
GuyHarwood: according to [the documentation ](https://docs.microsoft.com/en-us/sql/t-sql/functions/sum-transact-sql?view=sql-server-2017)SUM will adapt and return the correct datatype based on the inputs.  have you tested to see if it dynamically returns decimal without an explicit cast?

admin/data/sql/migrations/20190210134733.do.create-view-schools-average.sql
```diff
@@ -0,0 +1,31 @@
...
+FROM mtc_admin.pupil p
+-- FETCH COMPLETED PUPIL CHECK WITHIN THE CHECK WINDOW PERIOD
+LEFT OUTER JOIN
+(SELECT chk2.id, chk2.pupil_id, chk2.mark, ROW_NUMBER() OVER ( PARTITION BY chk2.pupil_id ORDER BY chk2.markedAt DESC ) as rank
+  FROM mtc_admin.[check] chk2
+  INNER JOIN mtc_admin.checkStatus cs ON cs.id = chk2.checkStatus_id
+  INNER JOIN mtc_admin.checkWindow cw ON cw.id = chk2.checkWindow_id
+  WHERE cs.code = 'CMP'
+        AND GETUTCDATE() BETWEEN cw.checkStartDate AND cw.adminEndDate
```
GuyHarwood: what if i run this query in november, and i want to see the results from April earlier in the year?

admin/data/sql/migrations/20190210204450.do.grant-delete-school-score.js
```diff
@@ -0,0 +1,5 @@
+const config = require('../../../config')
+
+module.exports.generateSql = () => {
+  return `GRANT DELETE ON [mtc_admin].[schoolScore] TO ${config.Sql.Application.Username}`
```
GuyHarwood: you could just use the default of `UPDATE` and set the column to NULL.  this would be the least permissive option, which is always preferable.

admin/data/sql/migrations/20190210134733.do.create-view-schools-average.sql
jon-shipley: @GuyHarwood - it returns a decimal as you surmise, but the cast will limit the number of decimal places to two.  So the choice is to either keep the cast and limit the decimal places to two, or lose it and have lots of decimal places.

GeorgeChatzigiannis: @GuyHarwood I tested this without the cast and it does not return a decimal.

admin/data/sql/migrations/20190210134733.do.create-view-schools-average.sql
GuyHarwood: OK. but if you need to refresh the figures at any point outside of the check window, this query won't work.  I would modify the query to remove the hardcoded check window and use a `WHERE` constraint when you call it to filter to the check window you are targeting.  That way the query can be used at any time.

admin/data/sql/migrations/20190210204450.do.grant-delete-school-score.js
GeorgeChatzigiannis: @GuyHarwood if I use `UPDATE` instead then I need to add more calls to determine if I need to perform insert (on the very first data submission) or update for the school score dataset.

jon-shipley: As SQL Server does not quite have Upsert; some patterns are listed here:
http://michaeljswart.com/2017/07/sql-server-upsert-patterns-and-antipatterns/

functions/calculate-score/score-calculation.data.service.js
```diff
@@ -0,0 +1,68 @@
...
+  DELETE FROM [mtc_admin].[schoolScore]
+  WHERE checkWindow_id = @checkWindow_id
+  
+  DECLARE @schoolScoreDataCursor CURSOR;
+  DECLARE @checkWindowId INT = @checkWindow_id
+  DECLARE @schoolId INT
+  DECLARE @schoolScore DECIMAL(5,2)
+  BEGIN
+     SET @schoolScoreDataCursor = CURSOR FOR
+      SELECT * from [mtc_admin].vewSchoolsAverage
```
jon-shipley: Should this view be called `vewSchoolsAverageScore`?

GeorgeChatzigiannis: @jon-shipley done

admin/data/sql/migrations/20190210204450.do.grant-delete-school-score.js
GeorgeChatzigiannis: @GuyHarwood @jon-shipley resolving this as the comments concern the older implementation

admin/data/sql/migrations/20190210134733.do.create-view-schools-average.sql
GeorgeChatzigiannis: @GuyHarwood resolving as there is a newer query for review

admin/data/sql/migrations/20190214132032.do.create-sp-refresh-score-data.sql
```diff
@@ -0,0 +1,73 @@
...
+      -- CLEAR THE SCHOOL SCORE TABLE
+      DELETE FROM [mtc_admin].schoolScore
+
+      -- ADD DATA TO SCHOOL SCORE TABLE
+      INSERT INTO mtc_admin.schoolScore (checkWindow_id, school_id, score)
+        (
+          SELECT
+          ISNULL(latestPupilCheck.checkWindow_id, @checkWindowId) as checkWindowId,
+          p.school_id,
+          CAST(SUM(ISNULL(latestPupilCheck.mark, 0)) AS DECIMAL(5,2)) / COUNT(p.id) as schoolScore
```
jon-shipley: The divisor here shouldn't be the number of pupils, rather the number of checks.   My bad, I gave you the wrong SQL!

GeorgeChatzigiannis: @jon-shipley absolutely right!
Tacked on this commit c9c6152 please have a review

functions/calculate-score/index.js
```diff
@@ -0,0 +1,20 @@
...
+  } catch (error) {
+    context.log.error(`calculate-score: ERROR: ${error.message}`)
+    throw error
```
GuyHarwood: do we really want to throw this if there is simply no check window to calculate?

functions/calculate-score/v1.js
```diff
@@ -0,0 +1,40 @@
...
+
+  // Terminate execution if a check window is not within the calculation period
+  if (!liveCheckWindow || !liveCheckWindow.id) {
+    context.log(`calculate-score: Live check window not found`)
+    return
+  }
+
+  // Terminate execution if relevant check window is finalised
+  if (liveCheckWindow.complete) {
+    context.log(`calculate-score: National score has been calculated for check window id ${liveCheckWindow.id}`)
```
GuyHarwood: could you change the wording slightly to 'Check window calculation already complete.  Bypassing calculation run'

functions/calculate-score/index.js
GeorgeChatzigiannis: @GuyHarwood within `v1` I am not throwing any error at all in case a check window is not detected but instead I simply return.
This is a standard try/catch we use in all the functions in case any data method throws an exception.
I would consider it quite important to keep this structure on the index level.

functions/calculate-score/v1.js
GeorgeChatzigiannis: @GuyHarwood done

functions/calculate-score/index.js
GuyHarwood: the concern i have is that we are assuming that this specific situation will never happen because it's coded perfectly to only all the stored procedure when there is a check window active.  I am not suggesting you swallow every error, but to handle the 'no check window' errors more gracefully.

GuyHarwood: you can use raiserror to return a specific custom error code and handle it based upon that

GeorgeChatzigiannis: @GuyHarwood in case a check window is not detected at the beginning of `v1` I am displaying this log message:
`calculate-score: Live check window not found`
There is no error thrown at all as I simply log and return.
Would you anticipate a different approach?

GeorgeChatzigiannis: I believe the store procedure call or any other data layer service call should only throw in case of a connection issue to the database.
This is a case why the catch exists on the top level of the function.

GuyHarwood: i am clear about what is going on and why.  but you just have a generic catch when you could be doing something useful and informative in the 'exceptional' circumstance when you *do* execute the function outside a check window, for whatever reason

GuyHarwood: the irony is that you are handling the 'exception' but not thinking that somehow there could be an exceptional situation where you execute it and it raises an error that you don't need to throw

GeorgeChatzigiannis: @GuyHarwood changed severity on the SP as discussed

PR #1019, opened 2019-02-07
Feature/29404 nightly build - pre & post database reset

GuyHarwood:
- move nightly to yaml based build definition
- add pre & post test database reset
- move all yaml build definitions to `deploy` folder
- create base build definitions for various tasks


deploy/build-defs/var-test.yml
```diff
@@ -0,0 +1,22 @@
+resources:
```
jon-shipley: Do we need this file?

GuyHarwood: not anymore, no 👍

GuyHarwood:
yaml build definitions are still buggy.   the bash commands in particular just self complete without even running the tasks.   the nightly build will have to remain as a UI based defintiion until such time as the yaml feature matures.

PR #1030, opened 2019-02-20
Feature/29889 - Added NTR checks to psychometrician report

danwilliams:
This adds in the "incomplete" checks to the psychometrician report.


admin/services/psychometrician-util.service.js
```diff
@@ -486,7 +486,19 @@ psUtilService.getInputMethod = function (inputs) {
   }
 }
 
+psUtilService.getPupilStatus = function (check) {
```
GeorgeChatzigiannis: Is there test coverage anywhere for this?

PR #1031, opened 2019-02-20
Feature/29466 aa readonly check closed

danwilliams:
Allows school staff to  view read-only access arrangements after the check end date and before the admin end date.


admin/views/partials/_readonly.ejs
```diff
@@ -1,5 +1,12 @@
-<% if (hdfSubmitted) { %>
+<%
+var availabilityData = availabilityData || {};
```
jon-shipley: This `var` surprised me.  Is there a reason to use `var` over `let`?

admin/services/business-availability.service.js
```diff
@@ -45,4 +48,23 @@ businessAvailabilityService.determinePinGenerationEligibility = (isLiveCheck, ch
...
 
+/**
+ * Retuns data for the avalibilty partial
+ * @param {Object} checkWindowData
+ * @returns {Promise<any>}
+ */
+businessAvailabilityService.getAvailabilityData = async (dfeNumber, checkWindowData) => {
+  const currentDate = moment.utc()
+  const hdfSubmitted = await headteacherDeclarationService.isHdfSubmittedForCheck(dfeNumber, checkWindowData.id)
+  const checkWindowClosed = currentDate.isAfter(checkWindowData.checkEndDate)
```
GeorgeChatzigiannis: I believe the final day should be admin end date and NOT check end date.
HDF can be submitted after check end date which should be an edge case.

admin/services/headteacher-declaration.service.js
```diff
@@ -115,7 +115,21 @@ headteacherDeclarationService.findLatestHdfForSchool = async (dfeNumber) => {
  * @return {Promise<boolean>}
  */
 headteacherDeclarationService.isHdfSubmittedForCurrentCheck = async (dfeNumber) => {
-  const hdf = await headteacherDeclarationDataService.findCurrentHdfForSchool(dfeNumber)
+  let checkWindow = await checkWindowV2Service.getActiveCheckWindow()
+  if (!checkWindow) {
+    return false
+  }
+  return headteacherDeclarationService.isHdfSubmittedForCheck(dfeNumber, checkWindow.id)
```
GeorgeChatzigiannis: what if dfeNumber is undefined or invalid?

admin/views/partials/_readonly.ejs
GeorgeChatzigiannis: does the value even change directly?
I think const should be fine?

admin/services/business-availability.service.js
danwilliams: Hmmmm, the requirements specifically state: “After the check end date and before the admin end date, school staff can view read-only access arrangements page, as per prototype”

admin/views/partials/_readonly.ejs
danwilliams: This uses `var` in because `availabilityData` may be undefined.

admin/services/business-availability.service.js
```diff
@@ -45,4 +48,23 @@ businessAvailabilityService.determinePinGenerationEligibility = (isLiveCheck, ch
   }
 }
 
+/**
+ * Retuns data for the avalibilty partial
+ * @param {Object} checkWindowData
+ * @returns {Promise<any>}
```
GeorgeChatzigiannis: @danwilliams return type should be an object right?

PR #1034, opened 2019-02-22
Feature/29950 new pupil validation and enhancements

GeorgeChatzigiannis:
- [x] Single pupil validation: New rules( Additional field to provide reason for pupils aged 7 or 10) and text changes
- [x] Multiple pupil validation: New error message that differentiates dob validation from single pupil submission


admin/data/sql/migrations/20190221174823.do.add-age-reason-pupil-table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE [mtc_admin].[pupil] ADD ageReason NVARCHAR(1000) NULL
```
jon-shipley: That is a lot of length on the reason field.  Does it need to be that long?  Do we mind have such long records inside the pupil table?   How often would we access the reason?  

Another pattern we could consider would be a pupilNote table (pupil_id, note_id) that offloaded the longer notes to a separate 'notes' table.

@GuyHarwood - thoughts?

GeorgeChatzigiannis: @jon-shipley same length is applied on access arrangements reason field and yes the requirement would be to be max 1000 characters long.

GeorgeChatzigiannis: We could offload it to `pupilAgeReason` table (pupil_id, reason)

GuyHarwood: @GeorgeChatzigiannis please use `nvarchar(max)` https://docs.microsoft.com/en-us/sql/t-sql/data-types/nchar-and-nvarchar-transact-sql?view=sql-server-2017

GuyHarwood: @GeorgeChatzigiannis comment updated as i understand this is against pupil table, which is fairly central.

GeorgeChatzigiannis: @GuyHarwood @jon-shipley updated as discussed

admin/controllers/pupil.js
```diff
@@ -229,7 +229,8 @@ const postEditPupil = async (req, res, next) => {
   }
 
   const trimAndUppercase = R.compose(R.toUpper, R.trim)
-
+  await pupilAgeReasonService.refreshPupilAgeReason(pupil['id'], req.body.ageReason, pupil['ageReason'])
```
GuyHarwood: why not `pupil.id` as per other calls?  threw me for a second...

admin/services/data-access/pupil.data.service.js
```diff
@@ -85,10 +85,13 @@ pupilDataService.sqlFindOneBySlug = async function (urlSlug, schoolId) {
...
   const sql = `
       SELECT TOP 1 
-      *  
-      FROM ${sqlService.adminSchema}.${table}
-      WHERE urlSlug = @urlSlug
-      AND school_id = @schoolId  
+      p.*,
+      pag.reason AS ageReason
+      FROM ${sqlService.adminSchema}.${table} p
+      LEFT OUTER JOIN ${sqlService.adminSchema}.[pupilAgeReason] pag
```
jon-shipley: Why would we return the pupilAgeReason in a generic method used in many places?

admin/controllers/pupil.js
GeorgeChatzigiannis: @GuyHarwood it's the same thing.
Reason why I did this is because my linter complains for an unresolved variable when described as `pupil.id` which is a minor warning.

admin/services/data-access/pupil-age-reason.data.service.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+pupilAgeReasonDataService.sqlInsertPupilAgeReason = async function (pupilId, reason) {
+
+  const sql = `
+    DECLARE @pupilAgeReasonId INT
+    DECLARE @pupilAgeReasonTable TABLE (id INT)
+
+    INSERT INTO ${sqlService.adminSchema}.pupilAgeReason
+    (pupil_id, reason)
+    OUTPUT inserted.id INTO @pupilAgeReasonTable
```
jon-shipley: You don't need to store a single id in a table, which is usually used for sets.  Why not do something like:
```
INSERT INTO [mtc_admin].[pupilAgeReason] (pupil_id, reason) VALUES (@pupilId, @reason);
UPDATE [mtc_admin].[pupil] SET pupilAgeReason_id = SCOPE_IDENTITY()
WHERE pupil_id = @pupilId;
```

And the SQL should be wrapped in a transaction, please.

admin/controllers/pupil.js
GuyHarwood: yes its just an object indexer, i get that.  In other places you use `pupil.id`, just looking for consistency 👍

admin/services/data-access/pupil.data.service.js
GuyHarwood: good catch.  this should be changed, as the main driver for a separate table was the overhead

admin/controllers/pupil.js
GeorgeChatzigiannis: @GuyHarwood updated for consistency

admin/services/data-access/pupil.data.service.js
GeorgeChatzigiannis: @GuyHarwood @jon-shipley it's changed now.

admin/services/data-access/pupil-age-reason.data.service.js
GeorgeChatzigiannis: @jon-shipley yea I should have used `SCOPE_IDENTITY` to simplify this, good point 👍 
It's updated.

admin/controllers/pupil.js
```diff
@@ -166,7 +166,7 @@ const getEditPupilById = async (req, res, next) => {
   res.locals.pageTitle = 'Edit pupil data'
   let pupilExampleYear
   try {
-    const pupil = await pupilDataService.sqlFindOneBySlug(req.params.id, req.user.schoolId)
+    const pupil = await pupilDataService.sqlFindOneBySlugWithAgeReason(req.params.id, req.user.schoolId)
```
jon-shipley: Much better, Thank you!

admin/services/data-access/pupil-age-reason.data.service.js
```diff
@@ -0,0 +1,96 @@
...
+      type: TYPES.Int
+    },
+    {
+      name: 'reason',
+      value: reason,
+      type: TYPES.NVarChar
+    }
+  ]
+
+  return sqlService.modify(sql, params)
```
jon-shipley: As there are actually two SQL statements here it would be preferable to wrap this in a transaction to ensure atomicity.  Would you consider changing this to use `sqlService.modifyWithTransaction()`?

admin/services/data-access/pupil-age-reason.data.service.js
```diff
@@ -0,0 +1,96 @@
...
+
+  const params = [
+    {
+      name: 'pupilId',
+      value: pupilId,
+      type: TYPES.Int
+    }
+  ]
+
+  return sqlService.modify(sql, params)
```
jon-shipley: Again, there are two SQL statements here, so please consider using `sqlService.modifyWithTransaction()`

admin/services/data-access/pupil-age-reason.data.service.js
GeorgeChatzigiannis: ✅

admin/services/data-access/pupil-age-reason.data.service.js
GeorgeChatzigiannis: ✅

admin/data/sql/migrations/20190221174823.do.add-age-reason-pupil-table.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE [mtc_admin].[pupilAgeReason] (
+  id int IDENTITY (1,1) NOT NULL,
+  pupil_id int NOT NULL,
```
jon-shipley: Sorry, I missed this earlier.  Shouldn't `pupil_id` be UNIQUE?

GeorgeChatzigiannis: @jon-shipley yes and pushed.

PR #1042, opened 2019-03-06
Spike/azure batch service

jon-shipley:
Spike into using Azure Batch Service with docker running node connecting to SQL Database


GeorgeChatzigiannis:
Codewise this looks good to me

PR #1043, opened 2019-03-06
Update pupils eligible for a restart view

jon-shipley:
Fix: Prevents pupils with an unconsumed restart from appearing in the list of pupils who can have a restart

Diff:
```
 > diff -w data/sql/migrations/20190306160234.undo.fix-pupils-eligible-for-restart-view.sql data/sql/migrations/20190306160234.do.fix-pupils-eligible-for-restart-view.sql
22a23,32
>     AND   p.id NOT IN (
>       -- remove pupils with unconsumed restarts
>       SELECT
>         p2.id
>       FROM
>         [mtc_admin].[pupil] p2
>           LEFT JOIN [mtc_admin].[pupilRestart] pr2 ON (p2.id = pr2.pupil_id and pr2.isDeleted = 0)
>       WHERE
>         (pr2.id IS NOT NULL AND pr2.check_id IS NULL)
>     )
33a44
> 
```


admin/data/sql/migrations/20190306160234.do.fix-pupils-eligible-for-restart-view.sql
```diff
@@ -0,0 +1,44 @@
...
+  WHERE
+    -- don’t select pupils who are not attending
+    pa.id IS NULL
+    -- pupils must have already attempted 1 or more checks that are started, complete, not received
+    AND   chkStatus.code IN ('STD', 'CMP', 'NTR')
+    AND   chk.isLiveCheck = 1
+    AND   p.id NOT IN (
+      -- remove pupils with unconsumed restarts
+      SELECT
+        p.id
```
GeorgeChatzigiannis: @jon-shipley shouldn't that be p2.id?
I was also wondering if checking the pupil status instead would be helpful and potentially provide a simpler solution?

jon-shipley: Updated to use p2.id - good catch 👍 

I looked at using pupil status as well, but this seems impossible as the pupil status does not directly encode restart status.

PR #1057, opened 2019-03-14
Feature/30131 service manager sce add school

danwilliams:
First of two tickets related to SCE schools... this one is mostly UI, plus adding the schools, and the timezone selection logic - the next one is for listing / removing them from list and applying timezone logic everywhere.


admin/helpers/sce.js
```diff
@@ -0,0 +1,27 @@
...
+      val.zones.forEach(z => {
+        const zoneName = z.split('/').pop().replace('_', ' ')
+        countryZones.push({ name: `${val.name}, ${zoneName} (GMT ${getOffset(z)})`, zone: z })
+      })
+    } else {
+      const z = val.zones[0]
+      countryZones.push({ name: `${val.name} (GMT ${getOffset(z)})`, zone: z })
+    }
+  })
+  return countryZones
```
jon-shipley: This looks like it might be worth caching.  The `json` file will only be updated by a release, and so the parsed `countryZones` data should be good for the entire life of the app.

On the other hand, if this is only used by the service-manager, it won't make much difference.

danwilliams: That is true indeed, it can be cached - I didn't think of it because the json data is not that huge in size. Caching added! 👍

admin/controllers/service-manager.js
```diff
@@ -298,6 +301,90 @@ const controller = {
...
+   * @param res
+   * @param next
+   * @returns {Promise.<void>}
+   */
+  postSceAddSchool: async (req, res, next) => {
+    const schools = await sceService.getSchools()
+    const schoolNames = schools.map(s => s.name)
+    const schoolUrns = schools.map(s => s.urn)
+
+    let validationError = await sceSchoolValidator.validate(req.body, schoolNames, schoolUrns)
```
GeorgeChatzigiannis: Minor comment: would expect the validator to be called from a service for consistency with the other controllers.

admin/helpers/sce.js
```diff
@@ -0,0 +1,30 @@
+'use stict'
```
GeorgeChatzigiannis: Shouldn't the file be called `sce-presenter` to align with the naming convention used on other presenters?

danwilliams:
No longer needed, as part of 30130 now.

PR #1059, opened 2019-03-15
Feature/30640 admin app cannot recover from failed connection issues

jon-shipley:
Add additional robustness to the sql handling


admin/config/sql.config.js
```diff
@@ -14,8 +14,8 @@ module.exports = {
   user: process.env.SQL_APP_USER || 'mtcAdminUser', // docker default
   password: process.env.SQL_APP_USER_PASSWORD || 'your-chosen*P4ssw0rd_for_dev_env!', // docker default
```
GuyHarwood: How come we have 2 files with same configuration settings?

admin/services/data-access/sql.service.js
```diff
@@ -230,32 +241,38 @@ sqlService.transformResult = function (data) {
...
-  if (params) {
-    for (let index = 0; index < params.length; index++) {
-      const param = params[index]
-      // TODO support other options
-      request.input(param.name, param.type, param.value)
+  addParamsToRequestSimple(params, request)
+
+  let result
+  try {
+    result = await request.query(sql)
```
GuyHarwood: Would be good to have a configurable amount of retries and wait times

admin/config/sql.config.js
jon-shipley: The `config/sql.config.js` went in with the move to use `mssql`.  The only thing holding up the removal of the previous config is:

* the `connect-tedious` session handler
* the Pupil Census upload 

Both of which are scheduled to be removed from the admin app in the near future.  Then the refactor can be completed.

admin/services/data-access/sql.service.js
jon-shipley: I would quite like to put this minimal version in and see a single recovered query from the logs before committing to a full-featured version.

The load-test environment might be suitable?

GuyHarwood: yep, sounds good

_spikes-poc/mssql-test/app.js
```diff
@@ -0,0 +1,74 @@
...
+    startTime = new Date()
+    try {
+      logger.warn('Attempting to initialise the connection pool')
+      await sqlService.initPool()
+      connected = true
+    } catch (error) {
+      // await sleep(2000)
+      // process.exit(1)
+      await sqlService.drainPool()
+      sqlService.pool = null
```
GeorgeChatzigiannis: Is it going to assume there is still a connection pool if you don't set it to null I suppose?

_spikes-poc/mssql-test/routes/index.js
```diff
@@ -0,0 +1,27 @@
+const express = require('express');
+const router = express.Router();
+const sqlService = require('../sql.service')
+const R = require('ramda')
+const data = 'sample-data '.repeat(1000)
+
+/* GET home page. */
+router.get('/', async function(req, res, next) {
+  try {
+    const result3 = await sqlService.query('SELECT COUNT(*) FROM [mtc_admin].[sessions]')
```
GeorgeChatzigiannis: whats the point of storing the first and last db transactions to a const if you are not using them?
I suppose you just replace these variables on your code below?
What is the goal here?

admin/app.js
```diff
@@ -32,7 +32,34 @@ const logger = require('./services/log.service').getLogger()
...
+ * @param ms - milliseconds
+ * @return {Promise}
+ */
+function sleep (ms) {
+  return new Promise(resolve => {
+    setTimeout(resolve, ms)
+  })
+}
+
+// Initialise the Database Connection pool
```
GeorgeChatzigiannis: App.js is growing in size, may I suggest to isolate the init db conn pool in a separate file that is called from here?

admin/app.js
```diff
@@ -32,7 +32,34 @@ const logger = require('./services/log.service').getLogger()
...
+    logger.debug('Attempting to initialise the connection pool')
+    await sqlService.initPool()
+  } catch (error) {
+    logger.alert('Failed to connect to the database', error)
+    // The initial probe connection was not able to connection: the DB is not available.  This will cause all
+    // connections in the connection pool to be initialised to closed connections. By pausing, we allow time for the
+    // db to become available.  When run in a docker container the PM2 process manager will restart the process, and
+    // hopefully the DB will be up by then.
+    await sqlService.drainPool()
+    await sleep(30000)
```
GeorgeChatzigiannis: why 30secs sleep?

admin/app.js
```diff
@@ -32,7 +32,34 @@ const logger = require('./services/log.service').getLogger()
...
+    process.exit(1)
```
GeorgeChatzigiannis: Isn't the exit part covered by drainPool?

admin/availability/middleware.js
```diff
@@ -9,20 +9,24 @@ async function isAdminWindowAvailable (req, res, next) {
   if (config.OVERRIDE_AVAILABILITY_MIDDLEWARE) {
     return next()
   }
-  const currentDate = moment.utc()
-  let checkWindow = await checkWindowV2Service.getActiveCheckWindow()
-  // if there is no current active check window, default to the latest found
-  if (!checkWindow) checkWindow = await checkWindowV2Service.getLatestCheckWindow()
-  if (checkWindow && currentDate.isBetween(checkWindow.adminStartDate, checkWindow.adminEndDate)) {
-    return next()
+  try {
```
GeorgeChatzigiannis: 👍

admin/app.js
GuyHarwood: @jon-shipley can we make this configurable please?  i would rather we have the ability to act quickly and set this to a lower value if ever necessary

admin/app.js
GuyHarwood: @GeorgeChatzigiannis not sure where you get this idea from?  drain pool is for draining a pool, not for exiting a process.  these are 2 completely separate tasks.  i raised a similar concern on your recent PR

GuyHarwood: @GeorgeChatzigiannis here is the code...
```
sqlService.drainPool = async () => {
  if (!pool) {
    logger.warn('The connection pool is not initialised')
    return
  }
  return pool.close()
}
```

admin/app.js
GuyHarwood: good point George, however i think this particular piece needs to be here to make a statement about its particular behaviour.  if anything, the route setup and other bigger chunks should be extracted

_spikes-poc/mssql-test/app.js
jon-shipley: This block of code basically represents my expectation of how I thought the connection handling should work.    Draining and setting the pool to null were my increasingly desperate attempts to make the connection pool re-connect to the database.   However, If we bring the app up without the database, the connection pool is filled with closed connections, which are unusable, and they never re-connect.

In the admin app implementation I called process.exit() to terminate the app and allow the pm2 node manager to restart it after a delay of 30 seconds to allow the DB to come up.

_spikes-poc/mssql-test/routes/index.js
jon-shipley: This is just a spike.  Code was being moved around, whilst I made many attempts to get a closed database handle open again.  It never happened.

admin/app.js
jon-shipley: 30 seconds is a wait time to allow the database to come back online.  It also prevents resource over-utilisation should the database not be available.

I will make it configurable.

admin/app.js
jon-shipley: @GeorgeChatzigiannis - what @GuyHarwood said.

GeorgeChatzigiannis: 👍

PR #1061, opened 2019-03-19
Task/30809 create 600k pupil census csv

GeorgeChatzigiannis:
- Isolated project that contains script to generate 600k unique pupil records


design/census-import/generate-census-records.js
```diff
@@ -26,53 +26,78 @@ function writeCsv (data) {
...
-  let baseUpn = '702500001'
-  let schoolsLength
+  let dfeNumbers
+  let dfeNumberCount
+  let upns = []
   try {
-    schoolsLength = await getSchoolsLength()
-    schoolsLength = schoolsLength - 1
+    dfeNumbers = await getSchoolDfeNumbers()
+    dfeNumberCount = await getSchoolDfeNumberCount()
```
GuyHarwood: seems pointless, why not just `dfeNumbers.length` ?

design/census-import/generate-census-records.js
```diff
@@ -26,53 +26,78 @@ function writeCsv (data) {
...
+    dfeNumberCount = dfeNumberCount - 1
```
GuyHarwood: why are you doing this?  add a comment or give the resulting variable a meaningful name

design/census-import/generate-census-records.js
```diff
@@ -26,53 +26,78 @@ function writeCsv (data) {
...
   } catch (error) {
-    winston.info(error)
+    console.log(error)
```
GuyHarwood: should you exit at this point or go into the loop?

design/census-import/config.js
```diff
@@ -0,0 +1,61 @@
...
+const oneMinuteInMilliseconds = 60000
+
+const getEnvironment = () => {
+  return process.env.ENVIRONMENT_NAME || 'Local-Dev'
+}
+
+module.exports = {
+  Logging: {
+    LogLevel: process.env.LOG_LEVEL || 'info',
+    LogDna: {
```
jon-shipley: LogDNA is no longer used

design/census-import/config.js
```diff
@@ -0,0 +1,61 @@
...
+      mac: undefined,
+      app: `MTCAdmin:${getEnvironment()}`,
+      env: `${getEnvironment()}`
+    },
+    Express: {
+      UseWinston: process.env.EXPRESS_LOGGING_WINSTON || false
+    },
+    SendToAppInsights: process.env.APPINSIGHTS_WINSTON_LOGGER || false
+  },
+  Monitoring: {
```
jon-shipley: This config is surprising - it should only have values in it that are required by the import.  A lot of this stuff seems unnecessary.

design/census-import/config/sql.config.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+require('dotenv').config()
+const toBool = require('to-bool')
+
+const oneMinuteInMilliseconds = 60000
+
+module.exports = {
+  database: process.env.SQL_DATABASE || 'mtc',
```
jon-shipley: Why do you need another config file?

design/census-import/generate-census-records.js
```diff
@@ -26,53 +26,76 @@ function writeCsv (data) {
...
-      upn,
-      faker.name.lastName(), // Surname
-      faker.name.firstName(), // Forename
-      i % 100 === 0 ? faker.name.firstName() : '', // Middlenames
-      i % 10 === 0 ? 'M' : 'F', // Gender
-      moment(faker.date.between('2005', '2006')).format('MM/DD/YY') // DOB
-    ]
-    csvData.push(record)
   }
   writeCsv(csvData)
```
jon-shipley: Is `writeCSV()` being called only once? i.e. the whole lot is stored in memory.  If so, what is the memory required for this process?

design/census-import/generate-census-records.js
GeorgeChatzigiannis: @GuyHarwood indeed, thanks that was totally redundant. updated

design/census-import/generate-census-records.js
GeorgeChatzigiannis: @GuyHarwood this count variable is redundant, I have updated the external loop on 9f34ce8

design/census-import/generate-census-records.js
GeorgeChatzigiannis: @GuyHarwood updated this to call drainPool() on catch and therefore exit.

GuyHarwood: does drainPool call `process.exit` ? surely you need to explicitly exit the routine?

GeorgeChatzigiannis: when drainPool is called the node process will exit

design/census-import/config.js
GeorgeChatzigiannis: @jon-shipley removed

design/census-import/config.js
GeorgeChatzigiannis: @jon-shipley I have extracted the sql config parts in `sql.config.js` and kept only the logging and monitoring parts in `config.js`

design/census-import/config/sql.config.js
GeorgeChatzigiannis: @jon-shipley I personally like having an isolated config for the sql. config js contains everything else as it stands for this small project.

design/census-import/generate-census-records.js
jon-shipley: Okay, could be worse 😀 

That's fine with me for a one-off dev tool

design/census-import/generate-census-records.js
GeorgeChatzigiannis: @GuyHarwood I have updated this block by assigning the necessary failure exit code since this is all it really needs.
`process.exit(1)` is not required and it is considered a bad practice for that case as the script will terminate gracefully when there is no additional work pending in the event loop.

GuyHarwood: but we do need to exit the routine.  does setting `process.exitCode` make the current procedure return early?  i would have thought not 🤔  there is a loop directly below the catch block, which as far as i can tell will execute after you have caught the error.  surely you do not want that loop to run?

GeorgeChatzigiannis: @GuyHarwood yea I overlooked it. We do need to exit immediately actually. You are correct.
I updated to use `process.exit(1)` instead

GeorgeChatzigiannis:
@GuyHarwood last comment addressed

design/census-import/generate-census-records.js
GuyHarwood: no worries.  many thanks for the changes. will approve now

PR #1064, opened 2019-03-20
Feature/30130 service manager

danwilliams:
This is the second SCE (MoD service schools) ticket, which completes the implementation of the logic etc.

It builds upon the first ticket.


admin/data/sql/migrations/20190308122704.do.add-sce-table.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE [mtc_admin].[sce] (
```
jon-shipley: Wondering if we should change this name? Apparently, SCE hasn't been used since 2013.  Perhaps something like schoolTimezone?

admin/data/sql/migrations/20190315104300.do.add-procedure-upsertSceSchools.sql
```diff
@@ -0,0 +1,48 @@
+CREATE PROCEDURE [mtc_admin].[spUpsertSceSchools]
+    @sceSchools [mtc_admin].[SceTableType] READONLY
+AS
+    IF @@TRANCOUNT <> 0
+        ROLLBACK TRANSACTION
+
+    BEGIN TRY
+    BEGIN TRANSACTION
+
+    MERGE INTO
```
jon-shipley: Nice one 🚀

pupil-api/src/services/pupil-authentication.service.ts
```diff
@@ -53,10 +53,11 @@ export const pupilAuthenticationService = {
...
     }
 
-    const pinExpires = moment(result.pinExpiresAt._)
-    const now = moment()
+    const school = JSON.parse(result.school._)
+    const pinExpires = moment.tz(moment(result.pinExpiresAt._).format('YYYY-MM-DDTHH:mm:ss.SSS'), school.timezone)
+    const now = moment.tz(school.timezone)
 
-    if (pinExpires.isBefore(now)) {
+    if (now.isAfter(pinExpires)) {
```
GeorgeChatzigiannis: isn't the previous if statement the exact same thing?

admin/services/data-access/sce.data.service.js
```diff
@@ -0,0 +1,128 @@
...
+  })
+  const params = []
+  schools.map((school, index) => {
+    params.push({ name: `schoolId${index}`, value: school.id, type: TYPES.Int })
+    params.push({ name: `timezone${index}`, value: school.timezone, type: TYPES.NVarChar })
+  })
+  const exec = 'EXEC [mtc_admin].[spUpsertSceSchools] @tvp'
+  const insertSql = insertHeader + inserts.join(',\n')
+  const sql = [declareTable, schools.length ? insertSql : '', exec].join(';\n')
+  const res = await sqlService.query(sql, params)
```
GeorgeChatzigiannis: shouldn't it be `modify` when Insert occurs?

PR #1066, opened 2019-03-20
Feature/30981 Census import function

GeorgeChatzigiannis:
New pupil census function - takes a csv upoaded to BLOB storage and creates a new table in SQL Server with the contents.

- [x] Introduced census import blob triggered function

The next ticket will UPSERT the contents into the pupil table and drop the staging table.


admin/data/sql/migrations/20190320145618.do.grant-census-import-user-permissions.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
+const config = require('../../../config')
+
+module.exports.generateSql = function () {
+  if (config.Sql.Azure.Scale) {
+    return `GRANT CREATE TABLE ON schema::[mtc_admin] to [${config.Sql.PupilCensus.Username}];`
```
GuyHarwood: is this enough and do they qualify for insert and update on tables by default?

functions/census-import/census-import.data.service.js
```diff
@@ -0,0 +1,59 @@
+const moment = require('moment')
+const mssql = require('mssql')
+const uuidv4 = require('uuid/v4')
+
+const config = require('../config')
+
+const schema = '[mtc_admin]'
```
GuyHarwood: unnecessary IMO.

ive seen this in sql service and its not proven worthwhile.  we arent going to change the schema name.

GeorgeChatzigiannis: @GuyHarwood removed const since its included in the name now on a1a933d

admin/data/sql/migrations/20190320145618.do.grant-census-import-user-permissions.js
GeorgeChatzigiannis: @GuyHarwood data insertion confirmed working. Update not tested since we haven't reached the stage where we update the pupil table.

GuyHarwood: needs testing on azure.  you are defaulting to sysadmin (sa) in local so everything works regardless

GeorgeChatzigiannis: @GuyHarwood updated that migration with insert and update permissions aside from create table.

PR #1068, opened 2019-03-21
Feature/31008 Census function data insertion

GeorgeChatzigiannis:
- Insert census table data insertion to pupil table


GuyHarwood:
looks good so far

PR #1079, opened 2019-03-25
Feature/31170 Include job id the pupil loading procedure

GeorgeChatzigiannis:
Screen Shot 2019-03-25 at 11 40 28 AM


GuyHarwood:
@GeorgeChatzigiannis given that this needs to be tested in dev, do we hold off on QA until it hits dev env?

GeorgeChatzigiannis:
@GuyHarwood I would expect so.

PR #1082, opened 2019-03-25
add new role SchoolNomAAMTC and mapping to teacher

GuyHarwood:


GuyHarwood:
@activemq @sinhaamit1980 this can only be tested in PP, so signing off and will QA in there.

PR #1085, opened 2019-03-25
Bug/30637 Pages availability issues

GeorgeChatzigiannis:
- [x] Include opening hours for generate pins availability


admin/services/business-availability.service.js
```diff
@@ -113,6 +113,7 @@ businessAvailabilityService.determineAccessArrangementsEligibility = (checkWindo
  */
 businessAvailabilityService.getAvailabilityData = async (dfeNumber, checkWindowData) => {
   const currentDate = moment.utc()
+  const isWithinOpeningHours = currentDate.hour() >= 8 && currentDate.hour() < 16
```
jon-shipley: @danwilliams - how will this play with your mod change?

GuyHarwood: not very well given that we are not taking into account the timezone of the school.  the offset is available as part of the session object

GeorgeChatzigiannis:
@GuyHarwood this is now using `moment-timezone` to determine the correct date time

activemq:
@GeorgeChatzigiannis - Additional pages found that need to be covered - 
http://localhost:3001/pupil-register/pupil/add
http://localhost:3001/pupil-register/pupil/add-batch-pupils
http://localhost:3001/access-arrangements/select-access-arrangements?
http://localhost:3001/attendance/confirm-and-submit

GeorgeChatzigiannis:
@activemq I don't think there is a limitation for pupil register. The page should always be visible as far as I know

PR #1096, opened 2019-03-27
Bug/31318 UPN validation changes

GeorgeChatzigiannis:
- UPN errors are now being stored in an array that are projected appropriately in the UI instead of displaying only 1 (usually the last occurred error)
- Validation for characters 5 to 12 has changed to validate **only** characters 5-12 instead of the whole upn value


admin/views/pupil-register/edit-pupil.ejs
```diff
@@ -20,11 +20,17 @@ const headingClass = 'heading-xlarge';
     <div class="column-two-thirds">
         <div class="error-summary" role="group" aria-labelledby="error-summary-heading-1">
             <h2 class="heading-medium error-summary-heading" id="error-summary-heading-1">Check the pupil’s details</h2>
-            <p>You need to fix the errors on this page before continuing. See highlighted errors below.</p>
+            <p>You need to fix the errors on this <p></p>age before continuing. See highlighted errors below.</p>
```
jon-shipley: typo?

GeorgeChatzigiannis: 🎖 🏅 🥇

PR #1097, opened 2019-03-27
Task/28505 accessibility statement accessible from footer

danwilliams:


admin/views/accessibility-statement.ejs
```diff
@@ -0,0 +1,176 @@
...
+        <h3 class="heading-medium">Pupil site</h3>
+
+        <p>The MTC does not support third-party screen readers due to compatibility issues with the timed element of the check.</p>
+        
+        <p>
+            To address this, MTC provides a number of access arrangements that can be applied to the pupil check. The following 
+            table shows which access arrangements are compatible with each device.
+        </p>
+
+        <table class="table-font-xsmall">
```
jon-shipley: NonBlocker: It seems likely that putting text in a table element makes it harder to hear when a screen reader is used.

PR #1104, opened 2019-04-02
Add a script to generate sql to move pupils

jon-shipley:


admin/bin/gen-move-pupil-sql.js
```diff
@@ -0,0 +1,42 @@
...
+  arr.forEach(async (row) => {
+    if (row[0].match(/Date/)) {
+      return
+    }
+    const targetDfeNumber = row[4]
+    const pupilUpn = row[7].replace(/[^a-zA-Z0-9]/g, '')
+    const s = generateSql(targetDfeNumber, pupilUpn)
+    sql.push(s)
+  })
+  console.log(sql.join('\n'))
```
GeorgeChatzigiannis: @jon-shipley Is the goal of this script to just log generated sql on the console?

jon-shipley: It outputs sql, that's all.

PR #1110, opened 2019-04-03
Feature/27845 pupilapi winston ai upgrades

danwilliams:


pupil-api/src/services/log.service.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Return the underlying `winston` logger
+   * @return {winston.Logger | *}
+   */
+  getLogger () {
+    return this.logger
+  }
+}
+
+export default new Logger()
```
jon-shipley: My reading of this line is that it will generate a new logger every time it is  `require`ed.   In my opinion, the logger should be a singleton.

What do you think?

GeorgeChatzigiannis: Seems wrong to me as well, good spot @jon-shipley

GeorgeChatzigiannis: Seems wrong to me as well, good spot @jon-shipley

danwilliams: Heh, it _looks_ like it does it every time it's required, but really that's the singleton pattern with TS - it instantiates the class and then exports it.

It might depend on the require cache, but really it's the same as the admin app approach.

PR #1114, opened 2019-04-03
Bug/31218 Display pins that correspond to pupils that are not flagged as NTTC

GeorgeChatzigiannis:
Migrations diffs:
Screen Shot 2019-04-05 at 12 33 12 PM

Screen Shot 2019-04-05 at 12 33 36 PM


GuyHarwood:
@activemq this will need a 'today & tomorrow' test as we did earlier.

PR #1123, opened 2019-04-05
Fixed time check on results page to check the opening hour with timezone

danwilliams:


admin/helpers/school-home-feature-eligibility-presenter.js
```diff
@@ -177,8 +177,10 @@ schoolHomeFeatureEligibilityPresenter.isHdfPageAccessible = (currentDate, checkW
  * @returns {Boolean}
  */
 schoolHomeFeatureEligibilityPresenter.isResultsPageAccessible = (currentDate, checkWindowData) => {
-  const resultsPageEligibilityDateTime = checkWindowData.checkEndDate.clone().add(3, 'days').set({ hour: 8, minutes: 0, seconds: 0 })
-  return currentDate.isSameOrAfter(resultsPageEligibilityDateTime)
+  const resultsPageEligibilityDate = checkWindowData.checkEndDate.clone().add(3, 'days')
```
GeorgeChatzigiannis: @danwilliams What happened to the part around setting the hour at 8am on this change?

Another thing I wanted to bring up is that we are using two different approaches to determine result page eligibility date.
You introduced one that seems a better fit overall in the same file (line 18):
`const resultsPageEligibilityDateTime = checkWindowData.checkEndDate.clone().add(1, 'weeks').isoWeekday('Monday').set({ hour: 8, minutes: 0, seconds: 0 })`

Should we use this one?

danwilliams: "Another thing I wanted to bring up is that we are using two different approaches to determine result page eligibility date."

I'm not exactly sure why there were two different ones, but I agree we should be using only one, I've updated it. Good catch! 😄

"What happened to the part around setting the hour at 8am on this change?"

It was changed below to check `currentDate.hour() >= 8`, but it's updated to use set now again after the approach change.

PR #1124, opened 2019-04-08
Feature/27768 retire prepare check messaging

danwilliams:


pupil-spa/gen_config.sh
```diff
@@ -1,13 +1,5 @@
 #!/bin/bash
-featureUseHpa=${FEATURE_USE_HPA:-"true"}
```
GeorgeChatzigiannis: @GuyHarwood we will need to remove the env variable `FEATURE_USE_HPA` from all environments once this gets merged.

GuyHarwood: yes

PR #1125, opened 2019-04-08
Feature/31570 Check Expiry Improvements (part 1)

GeorgeChatzigiannis:
This iteration includes:
- `check-expiry` & `check-not-received`: Exposed `isLiveCheck` property
- Filtering only live checks prior to updating pupil statuses


functions/lib/azure-storage-helper.js
```diff
@@ -166,7 +166,7 @@ const azureStorageHelper = {
   updatePupilStatus: async function (logger, logPrefix, checkData) {
     logger.info(`${logPrefix}: updatePupilStatus(): got ${checkData.length} pupils`)
     // Batch the async messages up, to limit max concurrency
-    const batches = R.splitEvery(100, checkData)
+    const batches = R.splitEvery(100, R.filter(c => c.isLiveCheck, checkData))
```
GuyHarwood: can we add some method documentation please that stipulates the requirements for the checkData param to indicate that `isLiveCheck` (and other existing required properties) is a requirement, and how it behaves.  I also wonder if it might be best to call it `updatePupilStatusForLiveChecks` to be more indicative of what it is doing.

GeorgeChatzigiannis: @GuyHarwood since we are using v2 for the callers in v2 of both check-expiry and check-not-received I extracted a new method as per your name suggestion which includes updated jsdoc and documentation.

PR #1116, opened 2019-04-04
Bug/31427 Incorrect offset for check window dates upon submission

GeorgeChatzigiannis:
Changed presentation to display start and end dates in utc format

PR #1130, opened 2019-04-10
Task/31769 Check window datetime2 type spike

GeorgeChatzigiannis:
- [x] Add migrations to convert check window start and end date columns to DATETIME2 type
- [x] Amend check window submission data to align with expected type change
- [x] Attempt to submit isostring dates

### DIFF
--- a/admin/data/sql/migrations/20190409201024.do.alter-check-window-date-types.sql
+++ b/admin/data/sql/migrations/20190409201024.do.alter-check-window-date-types.sql
@@ -1,30 +1,71 @@
-DECLARE @columnName VARCHAR(max)
-DECLARE @sql NVARCHAR(max)
+DECLARE
+    @ObjectName1 NVARCHAR(100)
+SELECT @ObjectName1 = OBJECT_NAME([default_object_id])
+FROM SYS.COLUMNS
+WHERE [object_id] = OBJECT_ID('[mtc_admin].[checkWindow]')
+  AND [name] = 'adminStartDate'
+EXEC ('ALTER TABLE [mtc_admin].[checkWindow] DROP CONSTRAINT ' + @ObjectName1)
+ALTER TABLE [mtc_admin].checkWindow
+    ALTER COLUMN adminStartDate DATETIME2 NOT NULL
+ALTER TABLE [mtc_admin].checkWindow
+    ADD DEFAULT GETUTCDATE() FOR adminStartDate
 
-DECLARE cur CURSOR FOR
- SELECT COLUMN_NAME
-   FROM INFORMATION_SCHEMA.COLUMNS
-  WHERE COLUMN_NAME LIKE '%Date'
-    AND TABLE_NAME = 'checkWindow'
+DECLARE
+    @ObjectName2 NVARCHAR(100)
+SELECT @ObjectName2 = OBJECT_NAME([default_object_id])
+FROM SYS.COLUMNS
+WHERE [object_id] = OBJECT_ID('[mtc_admin].[checkWindow]')
+  AND [name] = 'checkStartDate'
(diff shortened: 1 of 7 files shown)


admin/data/sql/migrations/20190409201024.do.alter-check-window-date-types.sql
```diff
@@ -0,0 +1,30 @@
...
+WHILE @@fetch_status = 0
+BEGIN
+    SET @sql = '
+    DECLARE @ObjectName NVARCHAR(100)
+    SELECT @ObjectName = OBJECT_NAME([default_object_id]) FROM SYS.COLUMNS
+    WHERE [object_id] = OBJECT_ID(''[mtc_admin].[checkWindow]'') AND [name] = ' + QUOTENAME(@columnName,'''') + '
+    EXEC(''ALTER TABLE [mtc_admin].[checkWindow] DROP CONSTRAINT '' + @ObjectName)
+    ALTER TABLE [mtc_admin].checkWindow ALTER COLUMN ' + @columnName + ' DATETIME2 NOT NULL
+    ALTER TABLE [mtc_admin].checkWindow ADD DEFAULT GETUTCDATE() FOR ' + @columnName + '
+    '
```
GuyHarwood: This is over engineered and makes it very difficult to understand your intentions.

Please just add explicit statements for each column change

GuyHarwood:
Please ensure this is NOT merged to master as it represents exploratory work at this point

admin/data/sql/migrations/20190409201024.do.alter-check-window-date-types.sql
GeorgeChatzigiannis: @GuyHarwood it was quite convenient to have this instead of 6 groups of queries. The intention was to get the start and end date column names from the check window table and for each one:
- drop the unnamed constraint
- modify the type to datetime2
- add constraint for default value

I can agree that it seems a bit over-engineered so I'll make these explicit.

GeorgeChatzigiannis:
@GuyHarwood indeed, changed target branch

admin/data/sql/migrations/20190409201024.do.alter-check-window-date-types.sql
GeorgeChatzigiannis: @GuyHarwood updated

GeorgeChatzigiannis:
@GuyHarwood submitting ISO date string proved to be successful with datetime2!
However I am looking into an issue where the presentation layer is using moment to display the data in the UI and therefore converting the date based on the offset from UTC.

admin/data/sql/migrations/20190409201024.do.alter-check-window-date-types.sql
GuyHarwood: There is a misunderstanding here - there is absolutely no point in querying out the column metadata.  Just perform the column changes explicitly, one by one with statically typed names

GuyHarwood:
@GeorgeChatzigiannis looking for [similar behaviours may help](https://github.com/tediousjs/tedious/issues/339) too

GeorgeChatzigiannis:
@GuyHarwood I checked that Github issue while investigating and the only suggestion was to construct a date object out of the string.
I did confirm though that `DateTime2` works fine with submitted ISO date strings so that does solve the data store issue we were facing.

GuyHarwood:
@GeorgeChatzigiannis just to reiterate, the purpose of this spike is to establish...

1. if we can send an exact date with time to the database, and have it stored as-is in dateTimeOffset.
2. if item 1 fails, can it be done with dateTime2 date type.  its not necessarily sending as string that is important, but saving unmodified, as UTC, without the offset which 'seems' to be picked up from the web server time

GeorgeChatzigiannis:
I am closing this as the presentation layer should be responsible for displaying strictly UTC data.
It is addressed on #1116

PR #1131, opened 2019-04-11
28591 - Added migration to add group_id to vewPupilsEligibleForRestart

danwilliams:


GeorgeChatzigiannis:
Diff:
Diff as seen below:
``>     pg.group_id,
16a18
>       LEFT JOIN [mtc_admin].[pupilGroup] pg ON p.id = pg.pupil_id
40c42,43
<     p.school_id
---
>     p.school_id,
>     pg.group_id``

PR #1146, opened 2019-04-17
Feature/31975 Incomplete status display

GeorgeChatzigiannis:
Incomplete status display on both school home and pupil register page

Screenshot 2019-04-17 at 11 48 36

Screenshot 2019-04-17 at 11 49 18


admin/controllers/school.js
```diff
@@ -19,9 +20,11 @@ controller.getSchoolLandingPage = async (req, res, next) => {
     const checkWindowData = await checkWindowV2Service.getActiveCheckWindow()
     const schoolName = await schoolService.findSchoolByDfeNumber(req.user.School)
     const featureEligibilityData = schoolHomeFeatureEligibilityPresenter.getPresentationData(checkWindowData, req.user.timezone)
+    const hasIncompleteChecks = await pupilRegisterService.hasIncompleteChecks(req.user.schoolId)
```
GeorgeChatzigiannis: This could have a performance impact on school home page. 
It will need to be measured with load testing.

admin/controllers/pupil-register.js
```diff
@@ -18,6 +19,8 @@ const listPupils = async (req, res, next) => {
     next(error)
   }
 
+  const incompletePupils = incompleteChecksPresenter.getPupilWithIncompleteChecks(pupilsFormatted)
```
GuyHarwood: how come we select the pupils and put it in 'pupilsFormatted' then go back for another query just to establish whether any of these pupils have an incomplete status?  could we not do all of that in one query?

GuyHarwood:
@GeorgeChatzigiannis think i have accidentally deleted your comment instead of mine 🤦‍♂️

admin/controllers/school.js
jon-shipley: This is fine but we might want to consider running those await calls in parallel in the future.

GuyHarwood: good spot! that would make a lot of sense.  Is that not an easy win by awaiting a promise.all ?

PR #1149, opened 2019-04-17
Bug/31811 check form blank rows

GeorgeChatzigiannis:
Enhanced check form file validation to return an error when blank rows are detected


admin/spec/back-end/lib/validator/check-form/single-check-form-validator.spec.js
```diff
@@ -68,6 +68,16 @@ describe('singleCheckFormValidator', function () {
         const singleCheckFormErrors = await singleCheckFormValidator.validate(uploadedFile)
         expect(singleCheckFormErrors).toEqual([`${checkFormErrorMessages.invalidFileCharacters} ${fileName}`])
       })
+      it('and returns a validation error object when two or more additional empty rows are found', async () => {
```
GuyHarwood: why aren't we making this more tolerant and just throwing away the blank lines?

jon-shipley: indeed

GeorgeChatzigiannis: @GuyHarwood @jon-shipley The requirement is to validate as false any file that contains extra blank lines.
The validator so far was tolerant by trimming blank lines but that caused issues during submission as blank lines were taken into account.

jon-shipley: My view is that validating white-space makes it harder for users.

GuyHarwood: i think we should be applying common sense here.  trim the blank lines and don't add them as empty questions. we are potentially making this more complicated for the user otherwise.  Talk to the BAs if you need to

GuyHarwood: the end goal should be to 'keep adding rows with 2 factors until i encounter a row without 2 factors separated by a comma'

GeorgeChatzigiannis: Business expects a flawless file with no extra blank lines from the test dev submission.
Meaning that additional blank lines shouldn't be allowed anywhere in the file.

GuyHarwood: 🤷‍♂️ fair enough

admin/services/check-form-v2.service.js
```diff
@@ -43,6 +43,9 @@ checkFormV2Service.prepareSubmissionData = async (uploadedFiles, checkFormType)
     return new Promise((resolve, reject) => {
       csv.fromPath(uploadedFile.file, { headers: false, trim: true })
         .on('data', function (row) {
+          if (!row[0] && !row[1]) {
```
GeorgeChatzigiannis: @jon-shipley @GuyHarwood It was agreed that we can go for a more tolerant approach on the check form submission rule about blank lines.
I reverted the trim removal change from the validation layer and added this change to discard lines with empty content(blank)

GuyHarwood: lovely 🏆

GeorgeChatzigiannis: @GuyHarwood this is going to trigger validation errors.
Mid-file blank lines are not allowed.
trim will only remove blank lines from beginning and end of the file.

PR #1150, opened 2019-04-18
Added group-pupil load test scenario

swapnaravula:
This branch contains below load test file, which uses 01-teacherLogins.csv file for teacher logins

teacher-groups-pupils.jmx


load-test/scenarios/group-pupil/teacher-groups-pupils.jmx
```diff
@@ -0,0 +1,269 @@
...
+          <hashTree/>
+        </hashTree>
+        <ConstantTimer guiclass="ConstantTimerGui" testclass="ConstantTimer" testname="Wait-Select-Pupils" enabled="true">
+          <stringProp name="ConstantTimer.delay">10000</stringProp>
+        </ConstantTimer>
+        <hashTree/>
+        <HTTPSamplerProxy guiclass="HttpTestSampleGui" testclass="HTTPSamplerProxy" testname="POST-Create-Pupil-Group" enabled="true">
+          <elementProp name="HTTPsampler.Arguments" elementType="Arguments" guiclass="HTTPArgumentsPanel" testclass="Arguments" testname="User Defined Variables" enabled="true">
+            <collectionProp name="Arguments.arguments">
+              <elementProp name="pupil[${PUPIL_ID_1}]" elementType="HTTPArgument">
```
GuyHarwood: does this only add 2 pupils to a group? if so, can we make this configurable?

swapnaravula: Yes, every teacher adds first two pupil from the list...we can add more pupil but, we need to know the number of students that each teacher has and remove the group or prepare them to be appeared for selection again if we are re running the test.

GuyHarwood: i think this should be configurable.  2 doesn't really simulate an actual real situation.  You should speak to BA and QA about realistic group sizes

swapnaravula: I have asked Ben and Amit and came to know that the group size would be 10-15 so, I have made this configurable now with default to 10 and changes are in the branch for the review.

PR #1160, opened 2019-04-25
Feature.32250 - updated wait time

activemq:


test/admin-hpa/features/step_definitions/pupil_register_steps.rb
```diff
@@ -47,7 +47,7 @@
 
 Then(/^I can see the status for the pupil is '(.*)'$/) do |status|
   pupil_register_page.load
-  Timeout.timeout(10){visit current_url until pupil_register_page.find_pupil_row(@details_hash[:first_name]).result.text == status}
+  Timeout.timeout(120){visit current_url until pupil_register_page.find_pupil_row(@details_hash[:first_name]).result.text == status}
```
GuyHarwood: does this keep checking on an interval and then timeout after 120 secs?

activemq: @GuyHarwood - This doesn't have a interval, it just keeps refreshing the page for a max of 120 secs  till the required status is displayed, if it doesn't then it times out

PR #1162, opened 2019-04-26
Feature/32141 force user interaction on refresh to enable speech

danwilliams:


pupil-spa/src/app/check/check.component.ts
```diff
@@ -392,10 +413,21 @@ export class CheckComponent implements OnInit {
...
+    let stateDesc = this.getStateDescription();
     console.log(`Refresh detected during state ${this.state} ${stateDesc}`);
     this.auditService.addEntry(new RefreshDetected());
 
+    if (this.storageService.getItem(CheckComponent.forceUserInteractionKey)) {
+      /*
+        there has already been a refresh during the questions, decriment the state,
+        as the force-user-interaction view will no longer be in allowedStates
+      */
+      this.state -= 1;
```
GeorgeChatzigiannis: @danwilliams what would be the state when you decrement?

danwilliams: The process of the change is:
1. Detect that the user has refreshed during a question with speech enabled
2. Insert a `force-user-interaction` slide at the current `state` index, which makes the user click a button (enabling speech in the browser) before then showing them the question they refreshed on
3. Insert a `forceuserinteraction` key into localstorage
4. If they refresh for a second time, detect the `forceuserinteraction` key in localstorage. Because the `force-user-interaction` slide was inserted but is now not present post second refresh, the `state` is off by one, so decrement to make it correct.

That line is #4 above. So it’s that the user has refreshed a second time with speech enabled, causing the `state` to be off by one, which needs fixing.

activemq:
Blocked till further notice, this is now with the BA's as it needs some analysis

GuyHarwood:
@danwilliams @activemq @sinhaamit1980 whats happening with this PR?  its been stale for 18 days

GuyHarwood:
Closing until feedback received from BAs

PR #1165, opened 2019-04-29
Bug/32144 hide absent pupils when selecting more

danwilliams:


admin/services/data-access/group.data.service.js
```diff
@@ -19,6 +21,14 @@ groupDataService.sqlFindGroups = async (schoolId) => {
   AND g.school_id=@schoolId
   GROUP BY g.id, g.name
   ORDER BY name ASC`
+  if (atleastOnePresentPupil) {
```
GeorgeChatzigiannis: I would prefer to keep the data service layer intact from any kind of logic.
Why not provide different data service methods based on what's required from the service layer?
@jon-shipley @GuyHarwood opinions?

admin/services/data-access/group.data.service.js
```diff
@@ -29,6 +30,34 @@ groupDataService.sqlFindGroups = async (schoolId) => {
...
+  FROM ${sqlService.adminSchema}.[group] g
+  LEFT OUTER JOIN ${sqlService.adminSchema}.pupilGroup pg 
+  ON g.id = pg.group_id
+  LEFT JOIN ${sqlService.adminSchema}.pupilAttendance pa
+  ON pa.pupil_id=pg.pupil_id
+  WHERE pa.id IS NULL
+  AND g.isDeleted=0
+  AND g.school_id=@schoolId
+  GROUP BY g.id, g.name
+  ORDER BY name ASC`
```
jon-shipley: I wonder whether it wouldn't be better to order the results in the application servers rather than the DB.  There is only 1 DB server, but we can horizontally scale the app servers.

This comment is not specific to this code but rather would apply across the app.

admin/services/data-access/pupils-not-taking-check.data.service.js
```diff
@@ -29,6 +29,29 @@ const pupilsNotTakingCheckDataService = {
     return sqlService.query(sql, params)
   },
 
+  /**
+   * @param {number} dfeNumber
+   * @description returns all pupils with specified school that don't have a record of attendance
```
jon-shipley:
```suggestion
   * @description returns all pupils with the specified school that are taking the check
```

admin/services/data-access/pupils-not-taking-check.data.service.js
```diff
@@ -29,6 +29,29 @@ const pupilsNotTakingCheckDataService = {
...
+   * @returns {Promise.<*>}
+   */
+  sqlFindPupilsWithoutReasons: async (dfeNumber) => {
+    const sql = `
+      SELECT p.*, NULL as reason
+      FROM ${sqlService.adminSchema}.[pupil] p 
+      INNER JOIN ${sqlService.adminSchema}.[school] s ON p.school_id = s.id
+      LEFT JOIN ${sqlService.adminSchema}.[pupilAttendance] pa ON p.id = pa.pupil_id
+      WHERE s.dfeNumber = @dfeNumber AND pa.id IS NULL
+      ORDER BY p.lastName ASC, p.foreName ASC, p.middleNames ASC, p.dateOfBirth ASC
```
jon-shipley: Same as previous order clause comment

admin/services/data-access/group.data.service.js
danwilliams: Having looked at this, I am of the opinion that moving the ordering to the application will have a negligible effect. I think also that the work being done elsewhere to improve the performance of the data layer will have a larger impact.

PR #1167, opened 2019-04-30
Feature/26005 check complete throw error with client check started at

danwilliams:


GuyHarwood:
@danwilliams having spoke to QA about the origins of the bug card behind this work, i've decided that this is unnecessary.  if anything, the functions should ensure they have the data they _need_ in the message, rather than an arbitrary property that should not be included.

PR #1174, opened 2019-05-01
Bug/32245 Duplicate records on pin generated list

GeorgeChatzigiannis:
Avoid duplicate active pupil pin records appearing by joining only non deleted pupil attendance records


GeorgeChatzigiannis:
@GuyHarwood that is correct here's the diff as well for both views:

Screenshot 2019-05-02 at 19 34 23

PR #1175, opened 2019-05-02
Added new seed to override passwordHash of user records inserted during seeding

danwilliams:


admin/data/sql/migrations/054.do.update-admin-passwords.js
```diff
@@ -1,7 +0,0 @@
-'use strict'
```
GeorgeChatzigiannis: @danwilliams what about the undo script?

GuyHarwood: good catch @GeorgeChatzigiannis

GuyHarwood:
@danwilliams thanks for the changes

PR #1177, opened 2019-05-02
Feature/32402 - optimise try it out pins view

GuyHarwood:
'quick win' optimisations that were identified in Wednesdays session.

note: this _could_ have been done as a stored procedure, but the difference in filtering requirements between the 2 usages would have made this a lot more effort.  

More formal changes to this area of the system are planned, so this will do as an interim perf improvement

### DIFF
--- a/admin/data/sql/migrations/20190502111953.do.create-new-tryitout-pin-view.sql
+++ b/admin/data/sql/migrations/20190502111953.do.create-new-tryitout-pin-view.sql
@@ -1,0 +1,33 @@
+CREATE VIEW [mtc_admin].[vewPupilsEligibleForTryItOutPin]
+AS
+ SELECT
+    p.id,
+    p.foreName,
+    p.middleNames,
+    p.lastName,
+    p.dateOfBirth,
+    p.school_id,
+    g.group_id,
+    p.urlSlug
+FROM
+  [mtc_admin].[pupil] p
+    LEFT JOIN [mtc_admin].[pupilAttendance] pa ON (p.id = pa.pupil_id and pa.isDeleted = 0)
+    LEFT JOIN [mtc_admin].[attendanceCode] ac ON (pa.attendanceCode_id = ac.id )
+    LEFT JOIN  [mtc_admin].[pupilGroup] g ON (g.pupil_id = p.id)
+    INNER JOIN [mtc_admin].[pupilStatus] ps ON (p.pupilStatus_id = ps.id)
+WHERE
+  -- include all pupils except those who are marked as not taking check because they left school
+  (ac.id IS NULL OR ac.code <> 'LEFTT')
+AND    ps.code IN ('UNALLOC',
+                  'ALLOC',
+                  'LOGGED_IN')
+    -- Exclude pupils who already have an active familiarisation check
+AND p.id NOT IN (SELECT
+    p2.id
+FROM
+    [mtc_admin].[pupil] p2
+        LEFT JOIN [mtc_admin].[check] AS chk ON (p.id = chk.pupil_id)
+        LEFT JOIN [mtc_admin].[checkStatus] AS chkStatus ON (chk.checkStatus_id = chkStatus.id)
+WHERE  chk.isLiveCheck = 0
+  AND    chkStatus.code IN ('NEW', 'STD', 'COL')
+  AND p2.school_id = p.school_id)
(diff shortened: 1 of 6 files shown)


admin/services/data-access/pin-generation.data.service.js
```diff
@@ -11,12 +11,48 @@ const serviceToExport = {
...
+                  INNER JOIN [mtc_admin].[pupilStatus] ps ON (p.pupilStatus_id = ps.id)
+              WHERE
+                -- include all pupils except those who are marked as not taking check because they left school
+                (ac.id IS NULL OR ac.code <> 'LEFTT')
+              AND    ps.code IN ('UNALLOC',
+                                'ALLOC',
+                                'LOGGED_IN')
+              AND school_id = ${schoolId}
+                  -- Exclude pupils who already have an active familiarisation check
+              AND p.id NOT IN (SELECT
```
jon-shipley: Actually, this could easily be a view by re-writing the sub-select:

```
  AND p.id NOT IN (SELECT
                       p2.id
                   FROM
                       [mtc_admin].[pupil] p2
                           LEFT JOIN [mtc_admin].[check] AS chk ON (p.id = chk.pupil_id)
                           LEFT JOIN [mtc_admin].[checkStatus] AS chkStatus ON (chk.checkStatus_id = chkStatus.id)
                   WHERE  chk.isLiveCheck = 0
                     AND    chkStatus.code IN ('NEW', 'STD', 'COL')
                     AND p2.school_id = p.school_id)`
```

GuyHarwood: nice 👌  will try this out...

GeorgeChatzigiannis: It's only a part of the query though, what would be the advantage to do so?

admin/services/data-access/pin-generation.data.service.js
```diff
@@ -11,12 +11,48 @@ const serviceToExport = {
    * @return {Promise<*>}
    */
   sqlFindEligiblePupilsBySchool: async (schoolId, isLiveCheck) => {
-    const view = isLiveCheck === true ? 'vewPupilsEligibleForLivePinGeneration' : 'vewPupilsEligibleForFamiliarisationPinGeneration'
-    const sql = `SELECT 
-                  * 
-                FROM ${sqlService.adminSchema}.${view}
-                WHERE school_id=@schoolId                 
+    let sql
+    if (!isLiveCheck) {
```
GeorgeChatzigiannis: Even though it's probably late at this stage, I still think data service layer should be logic free and encapsulate any logic related code (like if statements) in the caller service method that could potentially call different data service methods.

admin/data/sql/migrations/20190502111953.do.create-new-tryitout-pin-view.sql
```diff
@@ -0,0 +1,33 @@
...
+WHERE
+  -- include all pupils except those who are marked as not taking check because they left school
+  (ac.id IS NULL OR ac.code <> 'LEFTT')
+AND    ps.code IN ('UNALLOC',
+                  'ALLOC',
+                  'LOGGED_IN')
+    -- Exclude pupils who already have an active familiarisation check
+AND p.id NOT IN (SELECT
+    p2.id
+FROM
```
jon-shipley: non-blocking - sub-query needs indenting

admin/services/data-access/pin-generation.data.service.js
GuyHarwood: @GeorgeChatzigiannis its joining/binding via school id on the outer query, therefore less greedy on the `NOT IN` query due to the restriction.  When we call the view with `school_id=xxx` it should account for this in its plan and work over a reduced set, specific to the school in question.

admin/services/data-access/pin-generation.data.service.js
GuyHarwood: noted for future work

admin/data/sql/migrations/20190502111953.do.create-new-tryitout-pin-view.sql
GuyHarwood: beautified 💄

admin/services/data-access/pin-generation.data.service.js
GeorgeChatzigiannis: @GuyHarwood thanks for clarifying, that does make sense.

PR #1179, opened 2019-05-02
Feature/32442 Pupil status explanation panel

GeorgeChatzigiannis:


admin/views/pupil-register/pupils-list.ejs
```diff
@@ -22,6 +22,57 @@
...
+                    <div>
+                      This status will show when:
+                      <ul class="list list-bullet">
+                        <li>A PIN has yet to be generated for pupil to start the check</li>
+                        <li>A PIN has expired before use</li>
+                      </ul>
+                    </div>
+                  </li>
+                  <li>
+                    <div><strong>PIN generated</strong></div>
```
jon-shipley: It would have better accessibility if this were written as a definition list.

PR #1180, opened 2019-05-03
WIP exploratory work on caching check window calls

GuyHarwood:
basic teacher journeys hit the check window data service over 30 times.
this data rarely changes, and should be cached


GuyHarwood:
closing until parallel work on query response caching is complete

PR #1184, opened 2019-05-03
Bug/32454 Incorrect alignment on pupil register 

GeorgeChatzigiannis:
Limit name cell width and move content to next line if exceeds

Screenshot 2019-05-03 at 18 36 25


admin/views/pupil-register/pupils-list.ejs
```diff
@@ -59,18 +59,12 @@
                     >
                         <div class="highlight-wrapper">
                             <% if (isFeatureEnabled('pupilEdit') && !availabilityData.hdfSubmitted && !availabilityData.checkWindowClosed) { %>
-                                <a class="bold-small truncate" href="/pupil-register/pupil/edit/<%= p.urlSlug %>" >
-                                    <%= `${p.fullName}` %>
-                                </a>
+                                <a class="bold-small name-text-wrap" href="/pupil-register/pupil/edit/<%= p.urlSlug %>" ><%= `${p.fullName}` %></a>
```
jon-shipley: You don't need to convert the variable to a string before outputting it, just the variable will do.  Am I missing something?

GeorgeChatzigiannis: This has been there forever.
I only removed the spaces but I didn't notice it was stringified, it's updated now, good catch!

jon-shipley: What about this stringified value?

GeorgeChatzigiannis: Covered, thanks @jon-shipley

PR #1186, opened 2019-05-08
Feature/32530 Sas token expiry handling

GeorgeChatzigiannis:
- [x] Increased default sas token timeout to 24 hours
- [x] New route and component when expired sas token is detected on an attempt to submit check data
- [x] Added integration test


pupil-spa/src/app/services/check-complete/check-complete.service.ts
```diff
@@ -76,8 +76,14 @@ export class CheckCompleteService {
       this.auditService.addEntry(new CheckSubmissionAPICallSucceeded());
       this.onSuccess(startTime);
     } catch (error) {
+      console.log(error);
       this.auditService.addEntry(new CheckSubmissionAPIFailed(error));
-      this.router.navigate(['/submission-failed']);
+      if (error.statusCode === 403
+        && error.authenticationerrordetail.split(':')[0] === 'Signature not valid in the specified time frame') {
```
GuyHarwood: this is very explicit, which is good.  i wonder if we could get caught out by them changing the error message though.  🤔 should we use `contains` or even omit the message altogether?  as we are running this from within the app code, could we assume that they are following the correct process, so would only ever receive a 403 in this situation anyway?
thoughts @jon-shipley @GeorgeChatzigiannis

GeorgeChatzigiannis: @GuyHarwood 
I had a chat with Jon earlier about this and I raised a similar concerns.

We could rely entirely on the error code 403 and I wouldn't expect another case aside from sas token expiry where we would receive that.

However as a safeguard i added the extra part with `authenticationerrordetail`.
I am also concerned the received error structure can potentially change in the near future thus making this block obsolete.
For that reason me and Jon agreed to add an integration test that verifies the structure at least for these two keys and values.

I have also updated this to use `includes()` instead.

PR #1190, opened 2019-05-10
Feature/32512 modify ps reports

jon-shipley:
# Modify Psychometrician Report Generator

Move the psychometrician report to a long-running function 

- [x] Create a new long-running function app (to run on Function App Service)
- [x]  Stream data from the DB to disk when generating PS Report
- [x]  Stream data from the DB to disk when generating Anomaly Report
- [x] Generate Zip file using streams rather than memory


functions-app/.nvmrc
```diff
@@ -0,0 +1 @@
+v10.14.1
```
GuyHarwood: 🏆

functions-app/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "functions-app",
+  "version": "0.0.1",
+  "description": "MTC Azure Functions App Service Plan",
```
GuyHarwood: Please can we add the purpose in here instead - that they are long runnng functions

functions-app/report-psychometrician-create-download/service/data-access/psychometrician-report.data.service.js
```diff
@@ -52,4 +97,4 @@ const psychometricianReportDataService = {
   }
 }
 
-module.exports = psychometricianReportDataService
+module.exports = Object.assign(base, psychometricianReportDataService)
```
GuyHarwood: Nice 👌🏼

functions-app/lib/sql/sql.service.js
```diff
@@ -0,0 +1,607 @@
...
+ * Return a mssql.Request object
+ * @return {Promise<mssql.Request>}
+ */
+sqlService.getRequest = async function getRequest () {
+  await this.initPool()
+  return new mssql.Request(pool)
+}
+
+sqlService.streamQuery = async function streamQuery (recordSetFunc, rowFunc, errorFunc, doneFunc, sql, request, params = []) {
+  await this.initPool()
```
GeorgeChatzigiannis: sounds like this is not needed?

functions-app/package.json
jon-shipley: Done

functions-app/lib/sql/sql.service.js
jon-shipley: Good catch.  Fixed.

functions-app/report-psychometrician/service/data-service/psychometrician-report-cache.data.service.js
```diff
@@ -49,19 +60,18 @@ const psychometricianReportCacheDataService = {
...
-
+    const sql = getUnprocessedChecksSql(batchSize)
     const results = await sqlService.query(sql)
     return results.map(r => r.id)
+  },
+
+  hasUnprocessedChecks: async function () {
+    const sql = getUnprocessedChecksSql(1)
+    const results = await sqlService.query(sql)
+    if (results && results.length) {
```
GeorgeChatzigiannis: @jon-shipley can length ever be 0 here? shouldn't you also be checking if length is > 0?

jon-shipley: Yes, it can be 0, and, no, I don't think it needs to be `length > 0`.  Shorter seems clearer to me.

GeorgeChatzigiannis: wouldn't 0 make your expected result wrong?

PR #1192, opened 2019-05-13
Feature/32415 redis caching

danwilliams:
This is the first stage of the performance improvements work. It introduces caching in Redis, but without making changes that are too far-reaching. It introduces the general approach, which will be expanded upon in further work.


functions/sql-update/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+const process = require('process')
+const sqlService = require('../lib/sql/sql.service')
+const { DONT_DROP_REDIS, TYPES } = sqlService
+
+if (process.env.NODE_ENV !== 'production') {
+  require('dotenv').config()
+}
+
+const schema = `[mtc_admin]`
```
GeorgeChatzigiannis: minor one: To the best of my knowledge it is always going to be `mtc_admin` so that makes this variable declaration kinda redundant

GuyHarwood: agreed.  i removed these from the sql service a while back

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+              if (changes.delete && changes.delete.indexOf(r.id.toString()) > -1) {
+                return false
+              }
+              return r
+            }).filter(r => r !== false)
+            redis.set(foundKey, JSON.stringify(result), async () => {
+              console.log(`REDIS (update): Updated \`${foundKey}\``)
+              const sqlUpdateQueueName = queueNameService.getName(queueNameService.NAMES.SQL_UPDATE)
+              await azureQueueService.addMessageAsync(sqlUpdateQueueName, { version: 2, messages: [changes] })
+              console.log(`REDIS (update): Sent \`${foundKey}\` update to \`sql-update\` message queue`)
```
GuyHarwood: introduce `models/logger` and use logger.debug` in every instance where `console.log` is used

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+                }
+              }
+              if (changes.delete && changes.delete.indexOf(r.id.toString()) > -1) {
+                return false
+              }
+              return r
+            }).filter(r => r !== false)
+            redis.set(foundKey, JSON.stringify(result), async () => {
+              console.log(`REDIS (update): Updated \`${foundKey}\``)
+              const sqlUpdateQueueName = queueNameService.getName(queueNameService.NAMES.SQL_UPDATE)
```
GuyHarwood: the queue update functionality needs to be enabled separately from caching in general

functions/sql-update/function.json
```diff
@@ -0,0 +1,12 @@
+{
+  "disabled": false,
```
GuyHarwood: please set this to disabled for now, as it won't be used in phase 1.

functions/sql-update/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+module.exports = async function (context, sqlUpdateMessage) {
```
GuyHarwood: please can you follow the existing convention for functions, which encapsulates the function definition into a `v1.js` file with a single `process` method.  See some of the existing functions for an example

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * Returns the data from a Redis key entry
+ * @param redisKey - the full Redis key to get
+ * @returns {String}
+ */
+redisCacheService.get = redisKey => {
+  return new Promise((resolve, reject) => {
+    redis.get(redisKey, (err, result) => {
```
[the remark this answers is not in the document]
danwilliams: It would return a false result, allowing the query to proceed, because Redis errored. But to be thorough, I’ve added `reject` and a try/catch.

1. Before executing a SQL Query which is set to be cached in Redis, it looks for an existing cache.
2. If it can’t find it or there is an error, it returns false, which executes the SQL query (the result set is then cached)
3. If it can find it, it returns the cache

I’ve tweaked #2 to cause a `reject` error on error. So “no result” and error are treated slightly differently.

functions/sql-update/index.js
danwilliams: Hmmmm... I’ve removed it, but it’s still declared separately in `admin/services/data-access/sql.service.js`:
```
sqlService.adminSchema = '[mtc_admin]'
```

...which is why I declared it separately. Checking the latest master, just in case this was a recent change, it is still there - and then used 255 times across all the queries. Perhaps it's in a branch that hasn't been merged yet...?

If `mtc_admin` is the only schema we ever intend to use, then it’s quite right to remove that declaration. However, if the expectation is that the removal has already been done, that does not appear to be the case - at least, not everywhere.

admin/services/redis-cache.service.js
danwilliams: Changed all to use the logger.

admin/services/redis-cache.service.js
danwilliams: Added a second env var `REDIS_CACHE_UPDATING`. Documented in the README.

functions/sql-update/index.js
danwilliams: I've done this - however, when I added mine, I happened to look at `check-started` which doesn’t use a separate `v1.js` file. Neither do `diag` or `feedback`. Looking at the latest master for any updates, they all still just use `index.js` directly (as I had done) without splitting out into a `v1.js` file 🤔 

Am I looking in the right place?

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+ * which will be performed in SQL Server by `/functions`
+ * @param serviceKey - the prefix of the key in the format `serviceName.methodName`
+ * @param changes - object with changes in the format { tableName: 'foo-bar', update: { 1: { foo: 'bar' } } }
+ * @returns {Boolean}
+ */
+redisCacheService.update = (serviceKey, changes) => {
+  return new Promise(async (resolve, reject) => {
+    if (!REDIS_CACHING) {
+      resolve(false)
+    } else {
```
jon-shipley: Please turn the first part of the `if` into a short-circuit and ditch the `else` syntax.

functions/sql-update/index.js
GuyHarwood: i'm wondering if my changes were scrapped / overwritten then, because this was a while back.  I've always been of the opinion that it's overkill, as it won't be changed.

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+      const keys = await findKeys(new RegExp(`^${serviceKey}_`))
+      const foundKey = keys.length ? keys[0] : false
+      if (!foundKey) {
+        resolve(false)
+      } else {
+        redis.get(foundKey, (err, result) => {
+          if (err || !result) {
+            resolve(false)
+          } else {
+            result = JSON.parse(result)
```
jon-shipley: Cyclomatic complexity seems high for this block.  Simplify?

functions/sql-update/index.js
GuyHarwood: prepare-check is a good example that does use the preferred arrangement.

those 3 need bringing up to date, i will create a work item for this.

functions/sql-update/index.js
```diff
@@ -0,0 +1,74 @@
...
+        return param
+      })
+      if (deleteParams.length) {
+        queries.push(`DELETE FROM ${schema}.[${table}] WHERE id IN (@${deleteParams.join(',@')})`)
+      }
+    }
+  })
+
+  if (queries.length) {
+    console.log(queries)
```
jon-shipley: debug?

GuyHarwood:
Looking good, thanks for the changes 👍

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Returns the data from a Redis key entry
+ * @param redisKey - the full Redis key to get
+ * @returns {String}
+ */
+redisCacheService.get = redisKey => {
+  return new Promise((resolve, reject) => {
+    redis.get(redisKey, (err, result) => {
+      if (err || !result) {
+        console.log(`REDIS (get): Failed to retrieve \`${redisKey}\``)
```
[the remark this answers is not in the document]
danwilliams: Strange, this was added an hour ago but appears to refer to code that was changed before the comment was added

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+        const thisTables = redisCacheService.storedProceduresAffectedTables[p]
+        if (thisTables) {
+          tables = tables.concat(thisTables)
+        }
+      })
+    }
+    if (tables.length) {
+      const stream = redis.scanStream()
+      stream.on('data', keys => {
+        const keyRegex = new RegExp(`(_|-)(${tables.join('|')})`, 'i')
```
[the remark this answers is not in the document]
danwilliams: Good catch!

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+ * @param redisKey - the full Redis key to get
+ * @returns {String}
+ */
+redisCacheService.get = redisKey => {
+  return new Promise((resolve, reject) => {
+    redis.get(redisKey, (err, result) => {
+      if (err || !result) {
+        console.log(`REDIS (get): Failed to retrieve \`${redisKey}\``)
+        if (err) {
+          console.log(`REDIS (get): Error: ${err.message}`)
```
GuyHarwood: use logger.info

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+        }
+      } else {
+        console.log(`REDIS (get): Retrieved \`${redisKey}\``)
```
GuyHarwood: use logger.info

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+ * @param data - string data to insert
+ * @returns {Boolean}
+ */
+redisCacheService.set = (redisKey, data) => {
+  return new Promise((resolve, reject) => {
+    if (typeof data === 'object') {
+      data = JSON.stringify(data)
+    }
+    redis.set(redisKey, data, () => {
+      console.log(`REDIS (set): Stored \`${redisKey}\``)
```
GuyHarwood: use logger.info

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+    }
+    if (tables.length) {
+      const stream = redis.scanStream()
+      stream.on('data', keys => {
+        const keyRegex = new RegExp(`(_|-)(${tables.join('|')})`, 'i')
+        const matchedKeys = keys.filter(k => keyRegex.test(k))
+        if (matchedKeys.length) {
+          const pipeline = redis.pipeline()
+          matchedKeys.forEach(key => {
+            console.log(`REDIS (dropAffectedCaches): Dropped \`${key}\``)
```
GuyHarwood: use logger.info

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+                  r[prop] = changes.update[r.id][prop]
+                }
+              }
+              if (changes.delete && changes.delete.indexOf(r.id.toString()) > -1) {
+                return false
+              }
+              return r
+            }).filter(r => r !== false)
+            redis.set(foundKey, JSON.stringify(result), async () => {
+              console.log(`REDIS (update): Updated \`${foundKey}\``)
```
GuyHarwood: use logger.info

admin/services/redis-cache.service.js
GuyHarwood: use logger.info

functions/sql-update/index.js
GuyHarwood: use logger.debug

functions/sql-update/index.js
```diff
@@ -0,0 +1,74 @@
...
+    console.log(params)
```
GuyHarwood: use logger.debug

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,215 @@
...
+      stream.on('data', keys => {
+        const keyRegex = new RegExp(`(_|-)(${tables.join('|')})`, 'i')
+        const matchedKeys = keys.filter(k => keyRegex.test(k))
+        if (matchedKeys.length) {
+          const pipeline = redis.pipeline()
+          matchedKeys.forEach(key => {
+            logger.info(`REDIS (dropAffectedCaches): Dropped \`${key}\``)
+            pipeline.del(key)
+          })
+          pipeline.exec()
```
GuyHarwood: should we be awaiting this?

GuyHarwood:
@danwilliams please see Slack for wider discussion on next steps

danwilliams:
> console.log appearing again?

Sorry, what do you mean appearing again?

GuyHarwood:
@danwilliams they were swapped out with logger calls, but then console.log was appearing again.  Maybe an outdated commit / merge or rebase issue.  Not sure ,  but i was seeing them again 😄

admin/services/data-access/group.data.service.js
```diff
@@ -13,7 +14,7 @@ groupDataService.sqlFindGroups = async (schoolId) => {
   const sql = `
   SELECT g.id, g.name, COUNT(pg.pupil_id) as pupilCount 
   FROM ${sqlService.adminSchema}.[group] g
-  LEFT OUTER JOIN ${sqlService.adminSchema}.pupilGroup pg 
+  LEFT OUTER JOIN ${sqlService.adminSchema}.[pupilGroup] pg
```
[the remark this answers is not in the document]
danwilliams: This is now different as I’ve removed the dropping caches for every update query, which is what matched table names on the square braces.

admin/services/data-access/sql.service.js
```diff
@@ -260,15 +264,30 @@ function addParamsToRequestSimple (params, request) {
...
+    let result = false
+    if (redisKey) {
+      let redisResult = await redisCacheService.get(redisKey)
+      result = JSON.parse(redisResult)
+    }
+    if (!result) {
+      const request = new mssql.Request(pool)
+      addParamsToRequestSimple(params, request)
+      result = await request.query(sql)
+      await redisCacheService.dropAffectedCaches(sql)
```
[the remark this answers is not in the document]
danwilliams: 👍 Added.

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,213 @@
...
+      }
+      procedures.forEach(p => {
+        const thisTables = redisCacheService.storedProceduresAffectedTables[p]
+        if (thisTables) {
+          tables = tables.concat(thisTables)
+        }
+      })
+    }
+    if (tables.length) {
+      const stream = redis.scanStream()
```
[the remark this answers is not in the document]
danwilliams: This should be drastically reduced, now that dropping caches is done selectively, given that we are going to merge this method and not wait for the queuing.

admin/services/redis-cache.service.js
danwilliams: The structure has changed, but it is now an await.

admin/services/data-access/sql.service.js
```diff
@@ -476,17 +523,15 @@ sqlService.generateUpdateStatement = async (table, data) => {
...
   const preparedData = convertMomentToJsDate(data)
-  const {
-    sql,
-    params,
-    outputParams
-  } = await sqlService.generateInsertStatement(tableName, preparedData)
+  const { sql, params } = await sqlService.generateInsertStatement(tableName, preparedData)
   try {
-    const res = await sqlService.modify(sql, params, outputParams)
+    const res = await sqlService.modify(sql, params, dropRedisCaches ? tableName : false)
```
GuyHarwood: what's happened to the outputParams?

GuyHarwood:
To reduce risk / complexity / focus around this work i recommend the following...

1.  Close this PR
2. Create a fresh branch from master
3. Include the redis cache service and modifications to the data services.
4. Address the cache from the data services, rather than the sql service (as stipulated in point 2 of the acceptance criteria list)
3. Come back to this strategy when we have the next release candidate out

admin/services/data-access/sql.service.js
```diff
@@ -258,17 +262,49 @@ function addParamsToRequestSimple (params, request) {
...
  */
-sqlService.query = async (sql, params = []) => {
+sqlService.query = async (sql, params = [], redisKey, affectedTables) => {
   logger.debug(`sql.service.query(): ${sql}`)
   logger.debug('sql.service.query(): Params ', R.map(R.pick(['name', 'value']), params))
   await pool
 
+  if (affectedTables) {
+    /*
+      there shouldn't be any reason to do an update/insert/deete AND cache a result set,
```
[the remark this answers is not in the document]
danwilliams: whoops.

admin/services/data-access/sql.service.js
danwilliams: `generateInsertStatement` did used to return `outputParams`, but there is only one place in the codebase using `generateInsertStatement` and it doesn’t use the `outputParams` value. So I removed it as it’s not being used.

GuyHarwood: ok, thanks for clarification

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,206 @@
...
+redisCacheService.dropAffectedCaches = (tables = []) => {
+  return new Promise((resolve, reject) => {
+    if (!REDIS_CACHING || (Array.isArray(tables) && tables.length === 0)) {
+      return resolve(true)
+    }
+    if (typeof tables === 'string') {
+      tables = [tables]
+    }
+    tables = tables.map(t => t.replace(/[^a-z]+/gi, ''))
+    const stream = redis.scanStream()
```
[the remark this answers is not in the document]
danwilliams: A/C 2 doesn't describe getting rid of this. However, it has now gone as I have updated to directly drop rather than passing flags or table names.

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,206 @@
...
+}
+
+/**
+ * Stores data for a Redis key
+ * @param redisKey - the full Redis key to set
+ * @param data - string data to insert
+ * @returns {Boolean}
+ */
+redisCacheService.set = (redisKey, data) => {
+  return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
danwilliams: The previous feedback was specifically on a `pipeline.exec()` call. I wasn’t aware the other ioredis methods could use promises - good catch. Changed.

admin/services/data-access/check-window.data.service.js
```diff
@@ -203,10 +203,10 @@ const checkWindowDataService = {
    * @return {Promise.<*>}
    */
   sqlCreate: async (data) => {
-    return sqlService.create(table, data)
+    return sqlService.create(table, data, true)
```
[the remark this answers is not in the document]
danwilliams: I would have made this change, as I agree, but it has now been altered in approach so this bit has gone now.

functions/lib/sql/sql.service.js
```diff
@@ -187,7 +187,9 @@ const sqlService = {
     TinyInt: mssql.TinyInt,
     UniqueIdentifier: mssql.UniqueIdentifier,
     VarChar: mssql.VarChar
-  }
+  },
+  // string to identify SQL updates from message queues (where Redis has been updated manually first)
+  DONT_DROP_REDIS: '/*DONT_DROP_REDIS*/'
```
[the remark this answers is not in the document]
danwilliams: Yes, the SQL database is the _ultimate_ source of truth, but we do want to get things from redis when we can.

At the time, every update query was being scanned to find any caches that needed to be dropped. This was for the delayed SQL updates, where redis had already been updated directly, so they needed to be excluded from the cache dropping.

Now that caches are being dropped directly by methods which affect them, this is no longer required anyway.

admin/services/redis-cache.service.js
```diff
@@ -0,0 +1,132 @@
...
+    if (changes.delete && changes.delete.indexOf(r.id.toString()) > -1) {
+      return false
+    }
+    return r
+  }).filter(r => r !== false)
+  try {
+    await redis.set(serviceKey, JSON.stringify(result))
+    logger.info(`REDIS (update): Updated \`${serviceKey}\``)
+    const sqlUpdateQueueName = queueNameService.getName(queueNameService.NAMES.SQL_UPDATE)
+    await azureQueueService.addMessageAsync(sqlUpdateQueueName, { version: 2, messages: [changes] })
```
jon-shipley: I don't advise using an SQL update queue mechanism at this time.  This is a large new part to add in just before the June trials. It would be better parked until July.

I would suggest that updates simply write to the DB first, then update the cache.

danwilliams: As covered in previous discussions, this was added as a foundation for the next phase. The SQL queuing is not used at this stage. To make this clearer, all the unused areas have been removed, which does help to clarify the shape of the PR and likely reduce any confusion, so thank you for commenting on this.

PR #1195, opened 2019-05-15
Feature/32791 Override hdf submission on results view

GeorgeChatzigiannis:
Amended results controller and added presenter method for indicating date to override hdf submission as the second Monday after check end date


admin/controllers/results.js
```diff
@@ -36,10 +36,12 @@ controller.getViewResultsPage = async (req, res, next) => {
     return next(error)
   }
   const currentDate = moment.tz(req.user.timezone || config.DEFAULT_TIMEZONE)
-  const isResultsPageAccessible = schoolHomeFeatureEligibilityPresenter.isResultsPageAccessible(currentDate, checkWindow)
-  const nationalScore = resultPresenter.getScoreWithOneDecimalPlace(checkWindow.score)
-  schoolScore = resultPresenter.getScoreWithOneDecimalPlace(schoolScoreRecord && schoolScoreRecord.score)
-  if (!isHdfSubmitted || !isResultsPageAccessible) {
+
+  const isResultsPageAccessibleForSubmittedHdfs =
```
GuyHarwood: maybe i don't understand the context well enough, but it feels like you are appending some new logic to this method, outside the actual method. 🤔   or, is this just a single situation where the logic is different to other uses?

GeorgeChatzigiannis: @GuyHarwood all the logic lies within the presenter now.

admin/controllers/results.js
```diff
@@ -50,6 +52,8 @@ controller.getViewResultsPage = async (req, res, next) => {
     })
   }
   const pupilData = resultPresenter.getResultsViewData(pupils)
+  const nationalScore = resultPresenter.getScoreWithOneDecimalPlace(checkWindow.score)
```
jon-shipley: I don't much like the implementation-specific method name.  What if the business decides to change the business rules?  There are really two methods here: `getScore` and `formatScore`.

admin/controllers/results.js
```diff
@@ -36,10 +36,12 @@ controller.getViewResultsPage = async (req, res, next) => {
     return next(error)
   }
   const currentDate = moment.tz(req.user.timezone || config.DEFAULT_TIMEZONE)
-  const isResultsPageAccessible = schoolHomeFeatureEligibilityPresenter.isResultsPageAccessible(currentDate, checkWindow)
-  const nationalScore = resultPresenter.getScoreWithOneDecimalPlace(checkWindow.score)
-  schoolScore = resultPresenter.getScoreWithOneDecimalPlace(schoolScoreRecord && schoolScoreRecord.score)
-  if (!isHdfSubmitted || !isResultsPageAccessible) {
+
+  const isResultsPageAccessibleForSubmittedHdfs =
+    schoolHomeFeatureEligibilityPresenter.isResultsPageAccessibleForSubmittedHdfs(currentDate, checkWindow, isHdfSubmitted)
```
jon-shipley: Given that optimisation is now in full swing, how would you cache this to remove the need to query the db in the first place?

admin/helpers/school-home-feature-eligibility-presenter.js
```diff
@@ -8,19 +8,32 @@ const dateService = require('../services/date.service')
 const schoolHomeFeatureEligibilityPresenter = {}
 
 /**
- * Determine the results eligibility date
+ * Determine the results earliest eligibility date when hdf has been submitted
```
jon-shipley: I think a Presenter should only be formatting data for display.  This is actual business logic that might be better in a service.

Also the description is unclear to me.  Perhaps 'Determine the earliest date the results page can be shown for schools that have submitted the HDF' is better.  It

admin/helpers/school-home-feature-eligibility-presenter.js
```diff
@@ -185,14 +198,37 @@ schoolHomeFeatureEligibilityPresenter.isHdfPageAccessible = (currentDate, checkW
 }
 
 /**
- * Determine if results page is accessible
+ * Determine if results feature is accessible
+ * @param currentDate
+ * @param checkWindowData
+ * @returns {Boolean}
+ */
+schoolHomeFeatureEligibilityPresenter.isResultsFeatureAccessible = (currentDate, checkWindowData) => {
```
jon-shipley: Not sure about the word feature in the method name.  Is this the same as Results Page or something different?

admin/spec/back-end/controllers/results.spec.js
```diff
@@ -138,10 +142,41 @@ describe('results controller:', () => {
...
-      expect(schoolHomeFeatureEligibilityPresenter.isResultsPageAccessible).toHaveBeenCalled()
+      expect(schoolHomeFeatureEligibilityPresenter.isResultsPageAccessibleForSubmittedHdfs).toHaveBeenCalled()
       expect(resultPresenter.getResultsViewData).not.toHaveBeenCalled()
       expect(res.render).toHaveBeenCalledWith('results/view-unavailable-results', { breadcrumbs: undefined })
     })
+    it('renders results view page when hdf record for school is not found but datetime for unsubmitted hdfs has passed', async () => {
+      const res = getRes()
+      const req = getReq(reqParams)
+      spyOn(res, 'render')
+      spyOn(checkWindowV2Service, 'getActiveCheckWindow').and.returnValue({ id: 1 })
```
jon-shipley: The amount of setup here - the number of spy calls - is a bit high.  It feels like the controller is doing too much work.

admin/helpers/school-home-feature-eligibility-presenter.js
GeorgeChatzigiannis: @jon-shipley I was thinking about this yesterday myself and for that reason I have extracted the result availability related logic pieces to a dedicated service.

admin/spec/back-end/controllers/results.spec.js
GeorgeChatzigiannis: @jon-shipley what are you proposing for this?
This has been merged work into results controller for a while now, perhaps we could extract some work to simplify this taken into account that some of this calls will be removed once redis cache kicks in

jon-shipley: @GeorgeChatzigiannis - I have only read the diff, but this is clearly a complicated controller.  I wonder if it could be simplified?

admin/controllers/results.js
GeorgeChatzigiannis: @jon-shipley I do not see any reason to introduce `getScore` as the services already fetch the scores but I do agree and have introduced `formatScore` to reflect a more generic purpose for the presenter method

admin/controllers/results.js
GeorgeChatzigiannis: @jon-shipley sounds like this along with the rest of the features can be revisited once redis cache pt.1 is in play?

admin/helpers/school-home-feature-eligibility-presenter.js
GeorgeChatzigiannis: @jon-shipley this is moved to a dedicated service along with the other 2 methods.
Essentially this determines access to the feature from school home level where the other two methods decide whether unavailable scores page should be displayed instead of the actual results list when the user is browsing from school home page.

admin/spec/back-end/controllers/results.spec.js
GeorgeChatzigiannis: Not by much in my opinion.
These are the data feeds that are required for the results page:
- checkWindow
- pupils
- groups
- school score
- hdf
- results list page availability

Imagine each one has a dedicated service call.
Feel free to provide suggestions.

admin/helpers/school-home-feature-eligibility-presenter.js
```diff
@@ -7,36 +7,30 @@ const dateService = require('../services/date.service')
...
  * Fetch data for familiarisation and live pin generation eligibility
  * @param checkWindowData
+ * @param timezone
  * @returns {Object} Eligibility data including flags and relevant datetimes
  */
 schoolHomeFeatureEligibilityPresenter.getPresentationData = (checkWindowData, timezone) => {
   const currentDate = moment.tz(timezone || config.DEFAULT_TIMEZONE)
   const featureEligibilityData = {}
-  const resultsPageEligibilityDateTime = schoolHomeFeatureEligibilityPresenter.resultsPageEligibilityDateTime(currentDate, checkWindowData.checkEndDate)
+  const resultsPublishedDate = checkWindowData.checkEndDate.clone()
```
jon-shipley: Isn't this also business logic?

GeorgeChatzigiannis: @jon-shipley moved to service level

admin/helpers/result-presenter.js
```diff
@@ -28,6 +29,19 @@ resultPresenter.getResultsViewData = (pupils) => {
...
  */
-resultPresenter.getScoreWithOneDecimalPlace = (score) => score && (Math.round(score * 10) / 10)
+resultPresenter.formatScore = (score) => score && (Math.round(score * 10) / 10)
+
+/**
+ * Get results opening date in full GDS format
+ * @param {Object} checkWindow
+ * @returns {String}
+ */
+resultPresenter.getResultsOpeningDate = (checkWindow) => {
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley `resultsOpeningDate ` is being passed now as an arg which has been determined from service level

PR #1201, opened 2019-05-23
feature/32966-persist-check-payload-to-check-when-generating-pins

jon-shipley:
## Compress completed check payload

* Persist JSON payload when pin generated to mtc_admin.check.config
* Compress the message where possible
* Decompress when receiving at other end
* Make compression a feature toggle

### Client side library introduced: lz-string

This runs on the browser when sending the completed check back.  This needs a full browser test-suite for the pupil app.


functions/completed-checks/message-schema/message.v2.json
```diff
@@ -0,0 +1,5 @@
+{
+  "version": "2",
+  "checkCode": "5AFDD0DD-AF49-4D24-B72D-B711F9EFCFAC",
+  "archive": "<compressed v:1 message>"
+}
```
GuyHarwood: new line missing

functions/completed-checks/v1.js
```diff
@@ -93,7 +93,7 @@ async function handleCompletedCheck (context, completedCheckMessage) {
     PartitionKey: completedCheckMessage.checkCode,
     RowKey: uuid(),
     eventType: 'completed-check',
-    payload: JSON.stringify(completedCheckMessage),
+    payload: '<removed>',
```
GuyHarwood: maybe this could take the compressed message 🤔

functions/lib/compression.service.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
```
GeorgeChatzigiannis: @jon-shipley I do not personally find much value in generating a service that only call library methods.
Are you planning on extending this any further?

jon-shipley: No.  The real value is that it is an adapter.  We can swap out the implementation at a later date.  We definitely do not want core business logic depending directly on external code, so we have this little module.

functions/completed-checks/message-schema/message.v2.json
jon-shipley: Added

functions/completed-checks/v1.js
jon-shipley: The payload is already stored in the DB, which is surely the source of truth?

GuyHarwood: good point.  this is merely a trace of activity, so should be fine as is 👍

GuyHarwood:
@jon-shipley is this dev complete now, or still draft?

jon-shipley:
> @jon-shipley is this dev complete now, or still draft?

Draft...but not for long I hope

GuyHarwood:
VSO Build requeued

functions/completed-checks/v1.js
GuyHarwood: the conversation in the support meeting makes me consider doing this, as in a replay situation it could be useful.

jon-shipley: Now added

functions/lib/compression.service.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    const completedCheckMock = require('../completed-checks/mocks/completed-check-with-results.js')
+    const stringInput = JSON.stringify(completedCheckMock)
+    const compressed = compressionService.compress(stringInput)
+    const stringOutput = compressionService.decompress(compressed)
+    const outputObject = JSON.parse(stringOutput)
+    const compressedSizePercent = Math.round((compressed.length / stringInput.length) * 100)
+    expect(R.equals(stringInput, stringOutput)).toBeTruthy()
+    expect(stringInput.length > compressed.length)
+    // This get compressed to 18%
+    expect(compressedSizePercent < 20).toBeTruthy()
```
danwilliams: ...hmmm, not sure if checking the compression percentage in a unit test is necessarily best? Just a thought! 🙂

PR #1213, opened 2019-05-23
Feature/32935 pupil spa offline

danwilliams:


pupil-spa/src/app/services/route/route.service.ts
```diff
@@ -9,6 +10,10 @@ export class RouteService {
   }
 
   public setup(): void {
+    if (APP_CONFIG.websiteOffline) {
+      this.router.navigate(['/service-unavailable'])
```
GeorgeChatzigiannis: @danwilliams wouldn't it be better if the router navigation occurred within `app.component.js`?
Potentially after the route service setup call?
I think this shouldn't belong in the setup since the behavior is related to the availability of the app so the master component should be handling it.

danwilliams: I agree - changed 👍

pupil-spa/src/app/website-offline/website-offline.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  templateUrl: './website-offline.component.html',
+  styleUrls: ['./website-offline.component.scss']
+})
+
+export class WebsiteOfflineComponent implements OnInit {
+
+  constructor(private route: ActivatedRoute, private router: Router) { }
+  
+  ngOnInit() {
+    if (!APP_CONFIG.websiteOffline) {
```
GuyHarwood: 🏆

PR #1216, opened 2019-05-23
Added migrations to fix active/familiarisation pin views

danwilliams:
Marking a pupil as not attending and then reversing twice or more caused the views to show duplicates. Due to a join on the pupilAttendance table. Added grouping to fix it


GeorgeChatzigiannis:
@danwilliams wouldn't it be more efficient and correct if you just join on pupil and isDeleted = 0?

danwilliams:
@GeorgeChatzigiannis you are correct - adding a group by seemed the smallest change for me to make to fix it. Updated 👍

PR #1219, opened 2019-05-28
33152- Added migrations to drop `vewCompletedCheckCount` and `vewIncompleteCheckCount`

danwilliams:
They both relied on now removed `check.data` but weren't being used in the admin app anyway


jon-shipley:
If these views are not being used in the admin app perhaps we should drop them instead?

GuyHarwood:
@GeorgeChatzigiannis they are used by reporting users regularly, so will not be dropped.

GuyHarwood:
views are used by Power BI users

GuyHarwood:
@danwilliams feedback from the service manager indicates that these 2 views are used frequently and currently throw an error due to the inclusion of the `check.data` column.

Please could you modify this PR to `ALTER` the views and remove usage of that column.  Then we can push it through to master 😎

GeorgeChatzigiannis:
@danwilliams seems like `vewPupilActivity` and `vewNotStartedCheckCount` need to be tackled as part of this PR?

admin/data/sql/migrations/20190528131913.do.remove-checkdata-from-vewcompletedcheckcount.sql
```diff
@@ -0,0 +1,6 @@
+ALTER VIEW [mtc_admin].[vewCompletedCheckCount] AS
+  SELECT c.pupil_id, COUNT(c.id) AS [CompletedCheckCount]
+  FROM [mtc_admin].[check] c
+    LEFT JOIN [mtc_admin].checkResult ck ON (c.id = ck.check_id)
```
jon-shipley: NB: I am surprised that `ck` is short for the `c`heck `r`esult table.  Why not use `cr`?

admin/data/sql/migrations/20190528131913.do.remove-checkdata-from-vewcompletedcheckcount.sql
```diff
@@ -0,0 +1,6 @@
+ALTER VIEW [mtc_admin].[vewCompletedCheckCount] AS
+  SELECT c.pupil_id, COUNT(c.id) AS [CompletedCheckCount]
```
jon-shipley: This appears to be the number of pupils with completed checks.  Perhaps rename the alias for clarity?

GuyHarwood:
the rollback migrations are failing because the `data` column was removed a while ago.
i think in this particular situation you will have to omit the undo, as its to fix an already broken view

GuyHarwood:
closing temporarily to lower build agent demand

PR #1220, opened 2019-05-28
Feature/33149 hdf v2

GeorgeChatzigiannis:
- [x] Amended service layer method to allow hdf submission when check end date is in the past and incomplete checks are detected
- [ ] Amend client view for the above case to include Incomplete as a status


GeorgeChatzigiannis:
needs to be recreated

PR #1222, opened 2019-05-29
Feature/32916 test developer views raw payload

jon-shipley:
Allow test developers to view the completed check raw payload


admin/services/payload.service.js
```diff
@@ -0,0 +1,62 @@
+'use strict'
+const moment = require('moment')
+const R = require('ramda')
+
+const payloadDataService = require('./data-access/payload.data.service')
+
+const payloadService = {
+  addRelativeTimingsToSection: function addRelativeTimingsToSection (objects) {
```
GeorgeChatzigiannis: feels like this will need jsdoc?

admin/controllers/payload.js
```diff
@@ -0,0 +1,35 @@
...
+        formData: req.body,
+        error: error || new ValidationError(),
+        breadcrumbs: req.breadcrumbs()
+      })
+    } catch (error) {
+      next(error)
+    }
+  },
+
+  postViewPayload: async function postViewPayload (req, res, next, error) {
```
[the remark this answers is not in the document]
jon-shipley: The error path is not complete

admin/services/payload.service.js
jon-shipley: yes, you're right!

admin/routes/test-developer.js
```diff
@@ -20,5 +21,7 @@ router.get('/unassign-forms/:checkWindowId', isAuthenticated(rolesConfig.ROLE_TE
 router.post('/unassign-form', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.unassignCheckFormFromWindow(req, res, next))
 router.get('/download-pupil-check-data', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.getDownloadPupilCheckData(req, res, next))
 router.get('/file-download-pupil-check-data/:urlSlug', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => checkFormController.getFileDownloadPupilCheckData(req, res, next))
+router.get('/view-pupil-payload', isAuthenticated(rolesConfig.ROLE_TEST_DEVELOPER), (req, res, next) => payloadController.getViewPayloadForm(req, res, next))
```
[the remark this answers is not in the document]
jon-shipley: Fixed

admin/controllers/payload.js
```diff
@@ -1,34 +1,33 @@
...
-    console.log('User ', req.user)
+  postViewPayload: async function postViewPayload (req, res, next) {
     logger.info(`postViewPayload(): called for checkCode ${req.body.checkCode} by user '${req.user.UserName}' (id ${req.user.id})`)
-    const payload = await payloadService.getPayload(req.body.checkCode)
+    let response
     try {
-      res.type('json')
-      res.send(JSON.stringify(payload, null, '    '))
+      const payload = await payloadService.getPayload(req.body.checkCode.trim())
+        response = payload
```
GeorgeChatzigiannis: minor: feels like `response` is pushed by an extra tab?

admin/controllers/payload.js
```diff
@@ -1,34 +1,33 @@
...
     } catch (error) {
-      next(error)
+      logger.error(error)
+      response = { message: 'check not found' }
     }
+    res.type('json')
+    res.send(JSON.stringify(response, null, '    '))
```
GeorgeChatzigiannis: what's this whitespace string at the end? 🤔

admin/controllers/payload.js
jon-shipley: Good spot! Now fixed.

admin/controllers/payload.js
jon-shipley: 4 space indentation.  A poor man's pretty printer.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

admin/app.js
```diff
@@ -248,7 +248,8 @@ app.use(function (req, res, next) {
 const csrf = csurf()
 const csrfExcludedPaths = [
   '/auth', // disable CSRF for NCA tools
-  '/sign-in' // disable CSRF for login
+  '/sign-in', // disable CSRF for login
+  '/test-developer/view-pupil-payload'
```
danwilliams: Why is this excluded from CSRF? Do we expect it to need to be called from a script in any way, or can we just add a csrf token to the form? I've seen that the request is returned as raw json, but it's a bit odd that this is POST and comes from the UI, maybe getting it from the query like `/raw-pupil-payload/:checkCode` would be better? 🤔

jon-shipley: @danwilliams - Good question

The implementation returns the raw json in a new window, leaving the original form.  By omitting the csrf token we can allow the user to open several documents, which seems a reasonable use-case to me when investigating many checks from the reports.

Using GET may work too, will have a look next week.

jon-shipley: @danwilliams - updated to use GET and the csrf exclusion has been removed.

admin/services/data-access/payload.data.service.js
```diff
@@ -0,0 +1,25 @@
...
+  sqlFindOneByCheckCode: async function sqlFindOneByCheckCode (checkCode) {
+    const sql = `select cr.payload 
+                 from [mtc_admin].[check] chk 
+                 join [mtc_admin].[checkResult] cr on chk.id = cr.check_id
+                 where chk.checkCode = @checkCode`
+    const params = [
+      { name: 'checkCode', value: checkCode, type: sqlService.TYPES.UniqueIdentifier }
+    ]
+    const res = await sqlService.query(sql, params)
+    return JSON.parse(R.prop('payload', R.head(res)))
```
GuyHarwood: if the check can't be found would this throw or just return an empty `{}`?

admin/controllers/payload.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
```
GuyHarwood: i would have preferred a more descriptive name such as `pupil-payload-viewer` or similar.  Not hung up on it though, and appreciate it may be a pain to change at this stage, along with the service and tests etc..

admin/services/data-access/payload.data.service.js
jon-shipley: It throws from sqlQuery.  The exception is caught in the controller - try/catch wrapped around the `getPayload` method.  It then returns a check not found JSON object.

admin/controllers/payload.js
jon-shipley: Renamed to be descriptive.  Good call 👍

GuyHarwood:
closing temporarily to lower build agent demand

PR #1228, opened 2019-06-04
Spike/33238 nginx spa proxy to azure

danwilliams:


pupil-spa/src/app/services/azure-queue/azure-queue.service.ts
```diff
@@ -29,7 +29,7 @@ export class AzureQueueService {
    */
   public initQueueService(queueName: string, url: string, token: string, retryConfig): IQueueService {
     const service = this.queueStorage
-      .createQueueServiceWithSas(url.replace(queueName, ''), token)
+      .createQueueServiceWithSas(`${window.location.origin}/queue`, token)
```
GeorgeChatzigiannis: @danwilliams if I understand correctly this will be <app_url>/queue?
I am not sure I am following with this change but how will this address each queue and also how is this going to work with local development?

GeorgeChatzigiannis: As @GuyHarwood pointed out don't we need a fallback?

pupil-spa/nginx/default.conf
```diff
@@ -31,6 +31,20 @@ server {
     expires off;
   }
 
+  location /queue/ {
+    # resolver needed for the proxy_pass code below
+    resolver 8.8.8.8;
```
jon-shipley: There ought to be an Azure DNS server we can use.   E.g  from https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances#azure-provided-name-resolution

| The Azure DNS IP address is 168.63.129.16. This is a static IP address and will not change.

GuyHarwood:
This has now been tested successfully, with the following issues rectified...
- I had issues with the Azure DNS.  switching back to google worked
- the max attempts of 10 for the initial submission is way too long.  this has been reduced to a default of 3.

GuyHarwood:
Will sign off after final travis run

PR #1238, opened 2019-06-06
automation/fix-admin-test

sinhaamit1980:


admin/services/data-access/pupils-not-taking-check.data.service.js
```diff
@@ -47,13 +47,11 @@ const pupilsNotTakingCheckDataService = {
...
       LEFT JOIN ${sqlService.adminSchema}.[pupilAttendance] pa ON p.id = pa.pupil_id AND pa.isDeleted=0
       LEFT JOIN ${sqlService.adminSchema}.[pupilGroup] pg ON p.id = pg.pupil_id
-      LEFT JOIN ${sqlService.adminSchema}.[check] c ON p.id = c.pupil_id AND c.isLiveCheck=1
-      LEFT JOIN ${sqlService.adminSchema}.[checkStatus] cs ON cs.id = c.checkStatus_id
       LEFT JOIN ${sqlService.adminSchema}.[pupilRestart] pr ON pr.pupil_id = p.id AND pr.isDeleted=0 AND pr.check_id IS NULL
       WHERE
         p.school_id = @schoolId
-        AND pa.id IS NULL
-        AND (cs.code IS NULL OR pr.id IS NOT NULL)
+        AND p.pupilStatus_id = 1
```
GeorgeChatzigiannis: @jon-shipley why use ids? wouldn't it more secure if we used the relevant pupil status code?

GuyHarwood: it wouldnt be more secure, but it would be more readable.  however, the Ids dont change and unless the code is uniquely indexed it will not perform as well as PK lookup.   with the extra join you could even question whether it would ever perform as well anyway

GuyHarwood: sql server will optimise the crap out of it anyway.  get a query plan of it and see for yourself 😃

GeorgeChatzigiannis: I don't mind that much however it's just not consistent with other queries we got in place where we are making sure to lookup based on the codes instead of the ids.
The performance part is a fair point but in the end if both you @GuyHarwood and @jon-shipley are happy pointing to the relevant id then I will approve.

GuyHarwood: yes, im happy as-is.  i approved it already :)

PR #1240, opened 2019-06-06
Bug/33572 Retrieve only latest pupils checks on hdf review pupils list

GeorgeChatzigiannis:
When pupil restart is applied after a completed check duplicated appeared in the hdf pupil review list


GuyHarwood:
closing temporarily as discussed

PR #1243, opened 2019-06-07
Bug/33598 edit group with trailing space in name

danwilliams:


admin/services/group.service.js
```diff
@@ -83,6 +83,9 @@ groupService.update = async (id, group, schoolId) => {
   if (!(pupils instanceof Array)) {
     pupils = Object.values(pupils)
   }
+  if (typeof group.name === 'string') {
```
GeorgeChatzigiannis: Is there a case where this could come as a non-string value?

admin/services/group.service.js
```diff
@@ -104,6 +107,9 @@ groupService.create = async (groupName, groupPupils, schoolId) => {
   if (!groupName || !schoolId) {
     throw new Error('groupName and schoolId are required')
   }
+  if (typeof groupName === 'string') {
```
GeorgeChatzigiannis: Same as my other question

admin/services/group.service.js
danwilliams: Not really possible, since in both cases we're using a validator that name exists and is a string.

GeorgeChatzigiannis: @danwilliams wouldn't that make the if condition redundant then?

GeorgeChatzigiannis: Resolving since a newer commit covered that question

PR #1244, opened 2019-06-10
Bug/33659 pin generation queue messages are too large

jon-shipley:
[admin] - FIX check message size too large

Allow the number of prepare check messages to be configurable

### DIFF
--- a/admin/services/check-start.service.js
+++ b/admin/services/check-start.service.js
@@ -150,7 +150,8 @@ checkStartService.prepareCheck2 = async function (
 
   // Send batch messages each containing the up to 20 prepare check messages.   This avoids hitting the max message size
   // for Azure Queues of 64Kb
-  const batches = R.splitEvery(20, prepareCheckQueueMessages)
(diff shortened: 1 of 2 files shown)


admin/services/check-start.service.js
```diff
@@ -150,7 +150,8 @@ checkStartService.prepareCheck2 = async function (
 
   // Send batch messages each containing the up to 20 prepare check messages.   This avoids hitting the max message size
   // for Azure Queues of 64Kb
-  const batches = R.splitEvery(20, prepareCheckQueueMessages)
+  logger.info(`check start service: prepare check batch size is ${config.prepareCheckMessageBatchSize}`)
+  const batches = R.splitEvery(config.prepareCheckMessageBatchSize, prepareCheckQueueMessages)
```
GuyHarwood: could we please add in a log of the number of batches here too?  it could prove useful...

jon-shipley: done

PR #1245, opened 2019-06-10
Feature/33107 results v3

GeorgeChatzigiannis:
- [x] Removed average scores
- [x] Introduced new result list pupil states(incomplete, did not participate, did not attempt restart)
- [x] UI Changes
- [x] Introduced store procedure that returns pupil result view related data
- [x] Removed `ORDER BY` clause in data fetch query in favour of server side sorting


GuyHarwood:
Closing temporarily

admin/views/results/view-results.ejs
```diff
@@ -7,53 +7,17 @@
     <%- partial('partials/_session_expiry') %>
 
     <p class="lede" id="lead-paragraph">
-    Until verified, results are subject to review. All results are out of 25.
+      These results are for those who participated in the 2019 pilot.
+      Until verified, results are subject to review. All results are out <of></of> <%= maxMark %>.
```
jon-shipley: odd markup

admin/controllers/results.js
```diff
@@ -57,20 +56,12 @@ controller.getViewResultsPage = async (req, res, next) => {
...
-    })
-  }
-  const pupilData = resultPresenter.getResultsViewData(pupils)
-  const nationalScore = resultPresenter.formatScore(checkWindow.score)
-  schoolScore = resultPresenter.formatScore(schoolScoreRecord.score)
+  const pupilWithStatuses = resultService.assignResultStatuses(pupilResultData)
+  const pupilData = resultPresenter.getResultsViewData(pupilWithStatuses)
   return res.render('results/view-results', {
     pupilData,
+    maxMark: config.LINES_PER_CHECK_FORM,
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley you can get a scenario where there's no maxMark simply because all people are marked as not taking the check. That could potentially make this fragile.
The only stable way of receiving that information would be from the check form.
Since we have an env variable that defines the lines of check forms and to the best of my knowledge this hasn't changed it only makes sense to point to that.

admin/helpers/result-presenter.js
```diff
@@ -16,8 +16,8 @@ resultPresenter.getResultsViewData = (pupils) => {
       lastName: p.lastName,
       middleNames: p.middleNames,
       dateOfBirth: p.dateOfBirth,
-      score: p.reason ? '-' : p.mark,
-      reason: p.reason,
+      score: p.reason || p.statusInformation.length > 0 ? '-' : p.mark,
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley simply says: 
If there's no reason for not taking the check or no non-resulted status to be reported on the UI it means we do have a score to be displayed

admin/views/results/view-results.ejs
GeorgeChatzigiannis: @jon-shipley good spot! updated!

admin/helpers/result-presenter.js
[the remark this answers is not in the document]
jon-shipley: ok

admin/helpers/table-sorting.js
```diff
@@ -0,0 +1,50 @@
...
+  },
+
+  comparer: function (v1, v2, asc) {
+    return this.isNumericValue(v1) && this.isNumericValue(v2)
+      ? this.getNumberComparisonResult(v1, v2, asc) : this.getStringComparisonResult(v1, v2, asc)
+  },
+
+  isNumericValue: function (v) {
+    const numericOnlyPattern = /^\d+$/
+    const numericOnlyRegExp = new RegExp(numericOnlyPattern)
```
jon-shipley: This Regexp only needs to be generated once, rather than every time.  It's static.

admin/helpers/table-sorting.js
```diff
@@ -0,0 +1,50 @@
...
+    } else if (!asc) {
+      return b.toString().localeCompare(a)
+    }
+  },
+
+  getNumberComparisonResult: function (a, b, asc) {
+    return asc ? a - b : b - a
+  },
+
+  isNullString: function (v) {
```
jon-shipley: Would you consider renaming this function to `isEmpty`?

admin/services/data-access/result.data.service.js
```diff
@@ -14,32 +64,36 @@ const R = require('ramda')
...
+          AND chk.checkWindow_id = @checkWindowId
+          AND chk.isLiveCheck = 1
+        ) latestPupilCheck
+        ON p.id = latestPupilCheck.pupil_id
+    LEFT JOIN [mtc_admin].[checkStatus] cs
+        ON (latestPupilCheck.checkStatus_id = cs.id)
+    LEFT JOIN [mtc_admin].pupilAttendance pa
+        ON (p.id = pa.pupil_id AND pa.isDeleted = 0)
+    LEFT JOIN [mtc_admin].attendanceCode ac
+        ON pa.attendanceCode_id = ac.id
```
jon-shipley: Do you think it would be worth combining these queries into a view?

The initial reason for breaking the queries up was to have two smaller queries hit the database rather than one large one.  This isn't relevant now we have a caching strategy.  However, a view would make support much easier.

admin/helpers/table-sorting.js
GeorgeChatzigiannis: @jon-shipley done

admin/helpers/table-sorting.js
GeorgeChatzigiannis: @jon-shipley definitely, as a matter of fact I'll have to reflect that change on the client side sorting too.

GeorgeChatzigiannis: done

admin/services/data-access/result.data.service.js
GeorgeChatzigiannis: It does make sense. I'll update once this is done.

GeorgeChatzigiannis: Added

admin/data/sql/migrations/20190621143742.do.create-pupil-results-view.sql
```diff
@@ -0,0 +1,59 @@
+CREATE VIEW [mtc_admin].[vewPupilsResults]
```
jon-shipley: Should be create or alter view to be idempotent.

admin/data/sql/migrations/20190621143742.undo.create-pupil-results-view.sql
```diff
@@ -0,0 +1 @@
+DROP VIEW [mtc_admin].[vewPupilsResults]
```
jon-shipley: Should be DROP VIEW IF EXISTS to be idempotent.

admin/data/sql/migrations/20190621143742.do.create-pupil-results-view.sql
```diff
@@ -0,0 +1,59 @@
...
+        ON p.id = latestPupilCheck.pupil_id
+    LEFT JOIN [mtc_admin].[checkStatus] cs
+        ON (latestPupilCheck.checkStatus_id = cs.id)
+    LEFT JOIN [mtc_admin].pupilAttendance pa
+        ON (p.id = pa.pupil_id AND pa.isDeleted = 0)
+    LEFT JOIN [mtc_admin].attendanceCode ac
+        ON pa.attendanceCode_id = ac.id
+    WHERE (ac.code IS NULL OR ac.code NOT IN ('LEFTT', 'INCRG'))
+    AND (lastPupilRestart.rank = 1 or lastPupilRestart.rank IS NULL)
+    AND (latestPupilCheck.rank = 1 OR latestPupilCheck.rank IS NULL)
```
jon-shipley: 💯

admin/helpers/table-sorting.js
```diff
@@ -0,0 +1,51 @@
+const numericOnlyPattern = /^\d+$/
+const numericOnlyRegExp = new RegExp(numericOnlyPattern)
```
jon-shipley: 👍

admin/data/sql/migrations/20190621143742.do.create-pupil-results-view.sql
GeorgeChatzigiannis: @jon-shipley I made the change but I think that should be strictly CREATE

admin/data/sql/migrations/20190621143742.undo.create-pupil-results-view.sql
GeorgeChatzigiannis: @jon-shipley changed

admin/services/data-access/result.data.service.js
```diff
@@ -6,54 +6,16 @@ const TYPES = sqlService.TYPES
 const R = require('ramda')
 
 /**
- * Find pupils with scores
+ * Find pupils results data
  * @param {Number} schoolId
  * @param {Number} checkWindowId
  * @returns {Promise<*>}
  */
 resultDataService.sqlFindResultsBySchool = async (schoolId, checkWindowId) => {
```
jon-shipley: I wonder if this function should ensure that its arguments are present as they are required?

GeorgeChatzigiannis: @jon-shipley `result-service` `getPupilResultData ` method ensures these arguments are valid before calling this method. There should be no logic in the data layer as services should be assigned for these tasks.

GeorgeChatzigiannis:
@jon-shipley replied

PR #1246, opened 2019-06-12
Bug/33730 additional expire check logging

danwilliams:


functions/check-expiry/v2.js
```diff
@@ -9,6 +9,9 @@ const v2 = {
     // change the check status
     const checkData = await expireChecks(logger)
 
+    const checkIDs = checkData.map(e => e.checkId)
+    logger(`check-expiry: Updated ${checkData.length} checks' status to \`EXP\`: ${checkIDs.join(', ')}`)
```
jon-shipley: I wonder if using checkCode instead would make it easier to search application insights?

functions/expire-prepared-check/v1.js
```diff
@@ -51,6 +51,9 @@ async function deleteExpiredChecks (logger) {
     logger.error(`expire-prepared-checks: ERROR: deleteExpiredChecks(): ${error.message}`)
   }
 
+  const checkCodes = entities.map(e => e.checkCode.toUpperCase())
+  logger(`expire-prepared-checks: Removed ${entities.length} checks from \`${preparedCheckTable}\`:\n${checkCodes.join('\n')}`)
```
jon-shipley: This is a list of 500 checkCodes.  That's probably too much to log.

GuyHarwood: @danwilliams per Jons comment, can we snip this down to a max of 10 please?

functions/check-expiry/v2.js
```diff
@@ -9,6 +9,9 @@ const v2 = {
     // change the check status
     const checkData = await expireChecks(logger)
 
+    const checkCodes = checkData.map(e => e.checkCode)
+    logger(`check-expiry: Updated ${checkData.length} checks' status to \`EXP\`:\n${checkCodes.join('\n')}`)
```
jon-shipley: This will output 500 check codes.   Why?  It might be better to log each check expiry in the pupilEvent table then at least we can search for it.

Please note that Azure Application Insights will only sample data from functions - so not every logger output actually gets saved.

GuyHarwood:
Will not be merged as different longer term approach is emerging

PR #1253, opened 2019-06-14
Bug/33838 process v1 completed checks as default case

GuyHarwood:
- if no version specified on complete check message, process as v1
- increate batch size on check expiry to 1000
- double frequency of check expiry run to every 5 minutes


functions-app/poison-replay-completed-checks/index.js
```diff
@@ -0,0 +1,28 @@
...
+  } else {
+    version = 1
+    message = check
+  }
+  const id = uuid()
+  const file = path.join('./', `v${version}`, `${id}.json`)
+  context.log(`saving to ${file}`)
+  try {
+    const jsonString = JSON.stringify(message)
+    fs.writeFileSync(file, jsonString)
```
jon-shipley: This seems out of place to me in a function.  If we find these messages interesting we should save them somewhere properly like a DB.

For now, surely all this function needs to do is to cycle the poison queue by injecting the messages back onto the processing queue?  Any messages that are not processed will end up back on the poison queue.

Also, why check this function in if it not working?  The write-to-file aspect suggests there might be a lot of additional functionality to come.

GuyHarwood: this function is very much just a sketch / spike to explore peeking messages, reading them and putting back on the queue.  

Why a function?  our azure queue service library in the admin app only includes functionality for adding messages onto a queue, and given current demands i wanted the quickest route to obtain a picture of schools affected.

Ideally this should be in a completely separate PR, as to keep the conversation focused.  So i think ill do just that.

functions/completed-checks/v2.js
```diff
@@ -32,12 +32,12 @@ const v2 = {
...
     // Default output is bound to the pupilEvents table (saved in table storage)
     context.bindings.pupilEventsTable = []
-    const base64EncodedPayload = Buffer.from(JSON.stringify(completedCheckMessage)).toString('base64')
+    // const base64EncodedPayload = Buffer.from(JSON.stringify(completedCheckMessage)).toString('base64')
     const entity = {
       PartitionKey: completedCheckMessage.checkCode,
       RowKey: uuid(),
       eventType: 'completed-check',
-      payload: base64EncodedPayload,
+      payload: 'removed',
```
GeorgeChatzigiannis: @GuyHarwood is the payload key-value pair really needed then?

GuyHarwood: not at all.  we could remove that.

GuyHarwood: @GeorgeChatzigiannis however, its been tested so i dont want to delay the release

functions/completed-checks/v1.js
```diff
@@ -26,7 +26,7 @@ const v1 = {
       PartitionKey: completedCheckMessage.checkCode,
       RowKey: uuid(),
       eventType: 'completed-check',
-      payload: completedCheckMessage,
+      payload: 'removed',
```
jon-shipley: 💯

functions/completed-checks/v2.js
```diff
@@ -32,12 +32,12 @@ const v2 = {
     await v1.handleCompletedCheck(context, v1Message)
     // Default output is bound to the pupilEvents table (saved in table storage)
     context.bindings.pupilEventsTable = []
-    const base64EncodedPayload = Buffer.from(JSON.stringify(completedCheckMessage)).toString('base64')
+    // const base64EncodedPayload = Buffer.from(JSON.stringify(completedCheckMessage)).toString('base64')
```
jon-shipley: May as well delete commented out code

functions/expire-prepared-check/function.json
```diff
@@ -2,10 +2,10 @@
   "disabled": false,
   "bindings": [
     {
-      "schedule": "0 */10 * * * *",
+      "schedule": "0 */5 * * * *",
```
jon-shipley: 💯

PR #1254, opened 2019-06-14
Feature/replay poison checks

jon-shipley:
* add new function to `functions-app` that reads from the `check-complete-poison` queue and adds the message to the `check-complete` queue.
* this function does not have any bindings so will not be triggered automatically
* reads 32 messages at a time, adds them to the new queue and then deletes them from the poison queue


functions-app/poison-replay-completed-checks/function.json
```diff
@@ -0,0 +1,5 @@
+{
+  "disabled": false,
+  "bindings": [
+  ]
```
GuyHarwood: please add timer trigger input binding to a schedule of something arbitrary like once a year in 2030.
Also set disabled to false.
Then we can play this manually from the portal when necessary, as you suggested.

GeorgeChatzigiannis:
@jon-shipley test coverage would need to be considered as debt I would assume

PR #1255, opened 2019-06-14
Added migration to restore vewcheckdiagnostic

danwilliams:
- Whilst it isn't used in the codebase, it is used externally


admin/data/sql/migrations/20190614171658.do.restore-vewcheckdiagnostic.sql
```diff
@@ -0,0 +1,30 @@
+CREATE VIEW [mtc_admin].[vewCheckDiagnostic] AS
```
jon-shipley: Please make this migration idempotent.  The view now exists in some environments and not others.  This view should always be the result.

danwilliams: Based on the fact that the original branch was merged prematurely, and the last commit ended up coming later... I have tried to establish the possible outcomes and hopefully addressed those now.

GuyHarwood:
Tested and working with latest changes.

Will sign off once we have a green travis build

danwilliams:
> Why are there so many migrations? I was only expecting to see one new up/down migration pair.

I thought you might have run them already, so I was playing safe. Amended to just the ones actually needed.

admin/data/sql/migrations/20190620172547.do.restore-vewcheckdiagnostic.sql
```diff
@@ -0,0 +1,37 @@
+IF NOT EXISTS (
```
jon-shipley: Please turn this back to the standard 'create or alter view ...' statement without an 'EXEC'.

PR #1257, opened 2019-06-18
Added school timezone to NCA auth session data - and logging of each logged in user's timezone

danwilliams:
Add timezone support for NCA Tools users


admin/authentication/nca-tools-authentication-strategy.js
```diff
@@ -42,10 +43,13 @@ module.exports = async function (req, done) {
     }
 
     const mtcUser = await ncaToolsUserService.mapNcaUserToMtcUser(userData)
+    const school = await schoolDataService.sqlFindOneById(mtcUser.school_id)
```
jon-shipley: mtcUser.school_id will only be present if the user has a school. Some roles don't.

admin/controllers/authentication.js
```diff
@@ -34,7 +34,11 @@ const getSignIn = (req, res) => {
 }
 
 const postSignIn = (req, res) => {
-  switch (req.user.role) {
+  // Only id is available from local and NCA auth
+  const { displayName, id, role, timezone } = req.user
+  logger.info(`User ID ${id} (${displayName}) timezone is "${timezone}"`)
```
jon-shipley: Can we get something to clarify this a bit? e.g. 'postSignIn: USER ID...`

PR #1258, opened 2019-06-19
Bug/34069 implement banned words list

jon-shipley:
Add banned words list
Add new LA Code


admin/services/pin-generation.service.js
```diff
@@ -12,11 +12,16 @@ const pupilAttendanceService = require('../services/attendance.service')
 const pupilIdentificationFlagService = require('../services/pupil-identification-flag.service')
 const restartService = require('../services/restart.service')
 const config = require('../config')
+const logger = require('./log.service').getLogger()
 
 const allowedWords = new Set(
   (config.Data.allowedWords && config.Data.allowedWords.split(',')) || []
 )
 
+const bannedWords = [
```
GeorgeChatzigiannis: @jon-shipley can't we just remove the word from the env variable? 🤔

jon-shipley: It's a safeguard from incorrect config

PR #1260, opened 2019-06-19
Bug/33948 Inconsistent pupil id type after submitting more than 20 restarts

GeorgeChatzigiannis:
- Ensure pupil id list is type of string before further service processing


GeorgeChatzigiannis:
@GuyHarwood the service level changes can be supported with unit tests as we do know at service level exactly how the argument structure looks and we can mock it. It's been added

admin/services/pupil.status.service.js
```diff
@@ -117,11 +117,14 @@ pupilStatusService.recalculateStatusByPupilSlugs = async (pupilSlugs, schoolId)
  * @param {number} schoolId - ID of school in school table
  */
 pupilStatusService.recalculateStatusByPupilIds = async (pupilIds, schoolId) => {
-  if (!(Array.isArray(pupilIds) && pupilIds.length > 0)) {
+  // After exceeding 20 items the request payload received contains object key-value pairs
+  // Detecting and converting them to strings is necessary as part of the processing
+  const processedPupilIds = pupilIds.map(p => typeof p === 'object' ? Object.values(p)[0] : p)
```
jon-shipley: I would prefer it if it was the controller's job to translate from the incoming web request to the format (Array) required by the service layer.

admin/services/restart.service.js
```diff
@@ -78,14 +78,17 @@ restartService.restart = async (
   if (!schoolId) {
     throw new Error('Missing parameter: `schoolId`')
   }
-  await pinService.expireMultiplePins(pupilsList, schoolId)
+  // After exceeding 20 items the request payload received contains object key-value pairs
+  // Detecting and converting them to strings is necessary as part of the processing
+  const processedPupilsIds = pupilsList.map(p => typeof p === 'object' ? Object.values(p)[0] : p)
```
jon-shipley: Same as above re controller's job.

admin/services/pupil.status.service.js
GeorgeChatzigiannis: @jon-shipley this is a data processing activity. I thought the same thing but don't we have services to cover that part?
Shouldn't controllers just be as much as logic free as possible?

jon-shipley: @GeorgeChatzigiannis 

It is explicitly the controller's job to translate from the IO device (the web) in order to call the business tier.  I don't see why we *want* the business tier to handle Arrays and Objects.

GeorgeChatzigiannis: @jon-shipley changes made

PR #1262, opened 2019-06-20
Bug/34039 add unique index

jon-shipley:
Add unique index 
Add new LA code


admin/data/sql/migrations/20190620154630.do.add-unique-index.sql
```diff
@@ -0,0 +1,11 @@
+-- Add an index if it does not exist
+IF NOT EXISTS(
+        SELECT *
+        FROM sys.indexes i
```
jon-shipley: @GuyHarwood - does the user that runs the migrations have permissions to read from sys.indexes?

GuyHarwood: @jon-shipley yes

GeorgeChatzigiannis:
@jon-shipley would it make any sense to have the  duplicate removal set as a store procedure that could be executed on demand?

PR #1263, opened 2019-06-21
Spike/32415 split pupil status

danwilliams:
A new pupilStatusLink table has been added and all queries/views which were using `pupil.pupilStatus_id` have been updated to use it.

A method has also been added to make it possible to directly update a Redis cache and then put a message into a new Azure message queue for updating the table in SQL Server. A listener has been added to `/functions` to parse these messages.


GuyHarwood:
Closing until we review current architecture

PR #1266, opened 2019-06-21
Spike/32415 improve generate check efficiency

danwilliams:
Removed need to join on school and checkForm when generating checks, as this data is already available in process.


admin/controllers/pupil-pin.js
```diff
@@ -154,10 +154,11 @@ const postGeneratePins = async function postGeneratePins (req, res, next) {
     const update = pinGenerationService.generateSchoolPassword(school)
     if (update) {
       await schoolDataService.sqlUpdate(R.assoc('id', school.id, update))
+      school.pin = update.pin
     }
 
     // depends on school pin being ready
-    await checkStartService.prepareCheck2(pupilsList, req.user.School, req.user.schoolId, isLiveCheck, school.timezone)
+    await checkStartService.prepareCheck2(pupilsList, school, isLiveCheck)
```
GeorgeChatzigiannis: @danwilliams does the whole school record required to be passed down?
Would it be best to provide a destructured format with just the props that are needed on `prepareCheck2` method?

GuyHarwood:
closing until architecture review complete

PR #1270, opened 2019-06-27
feature/34348-informative error message

danwilliams:
Provide dedicated error types which contain user and actual error messages


admin/app.js
```diff
@@ -303,10 +303,16 @@ app.use(function (err, req, res, next) {
   // catch CSRF errors and redirect to the previous location
   if (err.code === 'EBADCSRFTOKEN') return res.redirect('back')
 
+  const showError = req.app.get('env') === 'development' || req.user.role === 'HELPDESK'
```
GuyHarwood: this shouldnt be a dev only thing - its for production users.
also, the helpdesk in the spec is just an example.  it could be any type of user

admin/views/error.ejs
```diff
@@ -4,4 +4,8 @@
 
 <h2>Error code: <%= errorCode %></h2>
 <p>ID: <%= errorId %></p>
+<% if (typeof mtcInfo !== 'undefined') { %>
+<h2>MTC Information</h2>
```
GuyHarwood: i dont think we want an 'MTC Information' header for this.  just a paragraph element for the info will be fine.

admin/app.js
```diff
@@ -303,10 +303,16 @@ app.use(function (err, req, res, next) {
...
+
+  if (showError && err.hasOwnProperty('mtcInfo')) {
+    res.locals.mtcInfo = err.mtcInfo
+  }
+
   // render the error page
   // @TODO: provide an error code and phone number? for the user to call support
   res.locals.message = 'An error occurred'
-  res.locals.error = req.app.get('env') === 'development' ? err : {}
+  res.locals.error = showError ? err : {}
```
GuyHarwood: in dev, we still want the stack trace.  as per the criteria - preserve existing behaviour, but allow the info to be presented as an additional to what currently happens, if its provided.

admin/controllers/pupil.js
```diff
@@ -196,7 +196,9 @@ const getEditPupilById = async (req, res, next) => {
     const pupil = await pupilDataService.sqlFindOneBySlugWithAgeReason(req.params.id, req.user.schoolId)
     pupilExampleYear = pupilPresenter.getPupilExampleYear()
     if (!pupil) {
-      return next(new Error(`Pupil ${req.params.id} not found`))
+      let error = new Error(`Pupil ${req.params.id} not found`)
```
GuyHarwood: nice to see how it would be used.  please add this to the admin/readme.md as an example, but remove from the pupil controller, as its not a requirement.

admin/app.js
danwilliams: It's:
```req.app.get('env') === 'development' || req.user.role === 'HELPDESK'```
So it always shows the error (including the stack trace) in dev and *ALSO* shows it for `HELPDESK` users in production

File: admin/views/error.ejs (hunk 2 above)
Comment (danwilliams) [reply admin/views/error.ejs]: I had made it that you add an `mtcInfo` property. Instead, I’ll make that an object, so you could do something like:
```
error.mtcInfo = {
  'HELPDESK': 'Error for helpdesk users to see',
  'TEACHER': 'Error for teacher users to see',
  ... other roles
}
```
It’ll then show the error based on the user’s role.

Comment (GuyHarwood) [reply admin/views/error.ejs]: that's a nice addition, thanks 👍

Comment (GuyHarwood) [reply admin/views/error.ejs]: @danwilliams could you have a 'default/catch all' that shows for any kind of user please

File: admin/README.md [hunk 5]
```diff
@@ -160,3 +160,17 @@ after making changes ensure you do `docker-compose build` to rebuild from source
...
+
+## Error Information
+
+When triggering an error, you can specify a dedicated error type that includes the actual error and a user friendly message as arguments.
+
+These error types will need to be generated and extend from `mtcBaseError`. For example:
+```
+checkWindowV2Service.getCheckWindow = async (urlSlug) => {
+  if (!urlSlug || !validate(urlSlug)) {
+    throw new MtcCheckWindowNotFoundError(mtcCheckWindowNotFoundErrorMessages.errorMessage, mtcCheckWindowNotFoundErrorMessages.userMessage)
```
GuyHarwood: could we replace the mtcCheckWindowNotFoundErrorMessages messages type with an actual string, to make it a more obvious example of what this enables us to do.

admin/lib/errors/mtc-check-window-not-found.js
```diff
@@ -0,0 +1,6 @@
+'use strict'
```
GuyHarwood: the messages being separate from the error type feels fragmented.  Could we not house the message as a const within the type its used by?  seems a bit overkill to have a separate file for the message.  It should have only one purpose.

GuyHarwood:
As discussed with the team, let's progress as follows...

- keep specific error types
- create folder for error types under `/admin/error-types/`
- put user message as static string inside the error type
- change constructor so no user message is accepted
- remove check for user message in app.js
- update example in readme to reflect the above

admin/services/role.service.js
```diff
@@ -28,7 +30,7 @@ const service = {
 
     if (role === 'HELPDESK' && !school) {
       // There is no provision for helpdesk users to log on as themselves
-      throw new Error('Helpdesk users must impersonate a school to sign-in')
+      throw new MtcHelpdeskImpersonationError(mtcHelpdeskImpersonationErrorMessages.errorMessage, mtcHelpdeskImpersonationErrorMessages.userMessage)
```
GuyHarwood: please add new type that stipulates school not found in MTC

GeorgeChatzigiannis:
@GuyHarwood @jon-shipley ready for further reviews

admin/error-types/mtc-school-mismatch.error.js
```diff
@@ -0,0 +1,10 @@
+const mtcBaseError = require('./mtc-base.error')
```
jon-shipley: I would prefer to have all the error classes, including the base class defined in one file.  This decreases the maintenance burden.

```
const { MtcSpecificError } = require('mtc-error')
...
throw new MtcSpecificError(message)
```

admin/error-types/mtc-school-mismatch.error.js
```diff
@@ -0,0 +1,10 @@
...
+
+class MtcSchoolMismatchError extends mtcBaseError {
+  constructor (message) {
+    super(message, 'The school is not found in the MTC database')
```
GuyHarwood:
```suggestion
    super(message, 'The selected school is not registered in MTC')
```

admin/error-types/mtc-school-mismatch.error.js
GeorgeChatzigiannis: I don't mind that change but then we have a single directory which will contain only 1 file.

GeorgeChatzigiannis: @jon-shipley done

GeorgeChatzigiannis:
@GuyHarwood @jon-shipley all comments addressed

admin/services/nca-tools-user.service.js
```diff
@@ -18,9 +19,8 @@ const service = {
     let school
     if (ncaUser.School) {
       school = await schoolDataService.sqlFindOneByDfeNumber(ncaUser.School)
-      if (!school) {
-        throw new Error(`Unknown School:${ncaUser.School}`)
-      }
+    } else {
+      throw new MtcHelpdeskImpersonationError('No Dfe number provided by NCA tools')
```
jon-shipley:
```suggestion
      throw new MtcHelpdeskImpersonationError('No DfE number provided by NCA tools')
```

admin/services/role.service.js
```diff
@@ -28,7 +29,7 @@ const service = {
 
     if (role === 'HELPDESK' && !school) {
       // There is no provision for helpdesk users to log on as themselves
-      throw new Error('Helpdesk users must impersonate a school to sign-in')
+      throw new MtcSchoolMismatchError('No school is found with the given dfe number')
```
jon-shipley:
```suggestion
      throw new MtcSchoolMismatchError('No school found with the given DfE number')
```

PR #1273, opened 2019-06-28
33942 - fix ps report error

jon-shipley:
Allow check.audit to be missing


functions-app/report-psychometrician/service/anomaly-report.service.js
```diff
@@ -351,9 +355,15 @@ anomalyReportService.detectQuestionsThatWereShownForTooLong = (check) => {
   const audits = anomalyReportService.filterAllRealQuestionsAndPauseAudits(check)
   const config = check.data.config
   const head = R.head(audits)
+  if (!head) {
```
GeorgeChatzigiannis: From what I see you return when there's no single audit record.

Is that being handled properly afterwards when you return ?

jon-shipley: Yes.

We bail out of this test and move on to the next.

functions-app/report-psychometrician/service/anomaly-report.service.js
```diff
@@ -351,9 +355,15 @@ anomalyReportService.detectQuestionsThatWereShownForTooLong = (check) => {
...
+    return
+  }
   const tail = R.tail(audits)
+  if (!tail) {
```
GeorgeChatzigiannis: @jon-shipley that can also mean that you return if you have a single item within the array of audits.
Please confirm if that is expected behavior

jon-shipley: Removed - thanks!

PR #1277, opened 2019-06-30
bug/34409-add defaults for sql connections and requests to functions app

GuyHarwood:
adds defaults of 60 seconds and five minutes for connections and requests respectively


functions-app/lib/sql/pool-config.js
```diff
@@ -28,8 +28,8 @@ const mssqlConfig = {
   password: sqlConfig.Application.Password,
   server: sqlConfig.Server,
   database: sqlConfig.Database,
-  connectionTimeout: sqlConfig.connectionTimeout || 30000,
-  requestTimeout: sqlConfig.requestTimeout || 15000,
+  connectionTimeout: parseInt(sqlConfig.connectionTimeout, 10),
```
jon-shipley: Could we have the parseInt on the config.js - so the actual config is correct please?

PR #1278, opened 2019-07-02
Bug/33232 update config on current prepared checks

GeorgeChatzigiannis:
Depending on the live or try out check that is being taken the prepared check config of both live and try out active checks needs to be updated if they exist


functions/prepared-check-sync/check.data.service.js
```diff
@@ -0,0 +1,57 @@
...
+ * Find latest live and single try out checkCode for a particular pupil based on one of the pupil's checkCodes
+ * @param checkCode
+ * @return {Promise<Array>}
+ */
+module.exports.sqlFindActiveCheckCodesByCheckCode = async function (checkCode) {
+  const sql = `          
+  DECLARE
+      @pupilId INT = (SELECT pupil_id
+                      FROM [mtc_admin].[check] chk
+                      WHERE chk.checkCode = @checkCode)
```
jon-shipley: What happens if this returns null?

functions/prepared-check-sync/check.data.service.js
```diff
@@ -0,0 +1,57 @@
...
+  
+  SELECT
+     latestLivePupilCheck.checkCode AS livePupilCheckCode,
```
jon-shipley: Wouldn't this be easier to read as two separate queries?

functions/prepared-check-sync/v1.js
```diff
@@ -5,18 +5,19 @@ const uuid = require('uuid/v4')
...
 v1.process = async function process (context, preparedCheckSyncMessage) {
   const { checkCode } = preparedCheckSyncMessage
   context.log('prepared-check-sync: message received', checkCode)
-  const checks = await sqlUtils.sqlFindChecksByCheckCode(checkCode)
-  checks.map(async (check) => {
-    await v1.updatePreparedChecks(context, check.checkCode)
-  })
+  const checkCodes = await checkDataService.sqlFindActiveCheckCodesByCheckCode(checkCode)
+  for (let idx in checkCodes) {
+    await v1.updatePreparedChecks(context, checkCodes[idx])
```
jon-shipley: This should have a try/catch around it so that a single exception raised does not prevent the entire batch fro bein discarded.

functions/prepared-check-sync/check.data.service.js
GeorgeChatzigiannis: the query will return an object with null values and do nothing since I am filtering these.

functions/prepared-check-sync/v1.js
GeorgeChatzigiannis: added

GuyHarwood:
@sinhaamit1980 @GeorgeChatzigiannis any particular reason why travis was never enabled on this?
will approve once its all green...

GeorgeChatzigiannis:
@GuyHarwood no particular reason, should have been enabled

GuyHarwood:
@GeorgeChatzigiannis no worries, it's on its way now... 🚀

PR #1285, opened 2019-07-05
Bug/34410 Add requestTimeout and connectionTimeout to admin and functions configs

GeorgeChatzigiannis:


GuyHarwood:
@GeorgeChatzigiannis are you going to change the default as per Jons suggestion?   my view is that it makes sense.  a five minute timeout could starve resources in a bad situation.

GeorgeChatzigiannis:
@GuyHarwood done.

PR #1290, opened 2019-07-11
Feature/34771 fix ps download file

jon-shipley:
Fix Azure file uploads

Larger files do not get the blob properties returned.  This fix ensures that blobProperties are always returned, even for large files.

Additional logging added where deemed useful

### DIFF
--- a/functions-app/report-psychometrician-create-download/service/data-access/azure-file.data.service.js
+++ b/functions-app/report-psychometrician-create-download/service/data-access/azure-file.data.service.js
@@ -39,7 +39,7 @@ const azureUploadFromLocalFile = async (container, remoteFilename, localFilename
       resolve()
     })
   })
-  const pr = await new Promise((resolve, reject) => {
+  await new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


functions-app/report-psychometrician-create-download/service/data-access/azure-file.data.service.js
```diff
@@ -39,26 +39,50 @@ const azureUploadFromLocalFile = async (container, remoteFilename, localFilename
...
         else return resolve(result)
       }
     )
   })
-  return pr
+
+  // azure-storage: will return an `uploadBlobResult` for files < 40MB but a different object for files > 40MB
+  // that only contains the numbers of the committedBlocks, mis-spelt with 3 M's.  Wonderful!
+  // I can't see anyone actually needing that info, but the blob properties are definitely useful.
+  return azureGetBlobProperties(container, remoteFilename)
```
GeorgeChatzigiannis: I understand that you wish to return something meaningful from this method but what's the point of making a call to return data that are completely unused?

PR #1308, opened 2019-07-25
Feature/35226 Sign in page govuk toolkit update

GeorgeChatzigiannis:
- Styles and markup update

Screenshot 2019-07-25 at 17 21 33


GeorgeChatzigiannis:
@GuyHarwood yes aside from the STA logo section which had to be supported with custom styling/assets as it is no longer part of the toolkit.

PR #1314, opened 2019-08-05
Feature/35578 new govuk groups view update

GeorgeChatzigiannis:
Pages update:
- [x] Section unavailable
- [x] Groups Overview
- [x] Add group
- [x] Edit group

Components update:
- [x] Confirmation modal
- [x] Sticky banner


admin/assets/javascripts/jquery-modal.js
```diff
@@ -7,12 +7,18 @@
 $(function () {
   function startModal (e) {
     $('.modal-link').on('click', function (e) {
-      $('#js-modal-confirmation-button').attr('href', $(this)[0].href)
+      if ($('#js-modal-confirmation-button').is('button')) {
```
GuyHarwood: so this now works with both layouts i take it?

GeorgeChatzigiannis: @GuyHarwood that is correct.
Will amend accordingly when all the relevant usages are moved to the new toolkit

PR #1317, opened 2019-08-08
Feature/35668 new govuk pupils not taking the check update

GeorgeChatzigiannis:
Pages update:
- [x] Pupils not taking the check overview
- [x] Select pupils for not taking the check

Components update:
- [x] Group filter


admin/views/pupils-not-taking-the-check/select-pupils.ejs
```diff
@@ -1,97 +1,72 @@
...
+            <p class="govuk-body">
+                All pupils must be considered for the multiplication tables check at the end of year 4.
+                If a pupil is not going to take the check, a reason must be provided.
+            </p>
+        <% if (hdfSubmitted) { %>
+            <button type="submit" disabled="disabled" aria-disabled="true" class="govuk-button govuk-button--disabled" data-module="govuk-button">
+                Select pupil and reason
+            </button>
+        <% } else { %>
+            <button type="button" onclick="window.location.replace('/pupils-not-taking-the-check/select-pupils')" class="govuk-button" data-module="govuk-button">
```
jon-shipley: I think if we want to send the user to another location then we should use a href styled to look like a button.

GeorgeChatzigiannis: @jon-shipley faking a link as a button is against both GDS and accessibility rules

GeorgeChatzigiannis: @jon-shipley updated based on discussions on all buttons with onclick handlers

PR #1321, opened 2019-08-15
Feature/35910 new govuk restarts update

GeorgeChatzigiannis:
- Moved pupil form under pupil register view directory
- Used session expiry component with updated styles on pupil register pages

- [x] Updated restarts overview page
- [x] Updated pupil restart select page


admin/views/pupil-register/pupils-list.ejs
```diff
@@ -14,7 +14,7 @@
       <% } %>
       <%- partial('partials/_readonly', {availabilityData}) %>
       <h1 class="govuk-heading-xl">Pupil register</h1>
-      <%- partial('partials/_session_expiry') %>
+      <%- partial('partials/_gds_session_expiry') %>
```
jon-shipley: Why is it called _gds_session_expiry?  This is not vendored code.

GeorgeChatzigiannis: @jon-shipley it's the prefix I am using for any component that is using the new toolkit markup and styles. Happy to use another convention, any suggestions?

jon-shipley: _mtc works for me

GeorgeChatzigiannis: @jon-shipley I will amend all the component clones that have received new markup and style on the next PR as agreed.

PR #1332, opened 2019-08-27
Feature/35944 generate input files for anomaly report

jon-shipley:
Dump the PS data to disk as a pre-req for the analysis


functions-app/lib/mtc-fs-utils.js
```diff
@@ -0,0 +1,55 @@
...
+const path = require('path')
+
+const base = require('./logger')
+const functionName = 'mtc-fs-utils'
+
+const mtcFsUtils = {
+  /**
+   * Work out if we can use the supplied path as a writeable directory.  If not, use os.tmpdir
+   * @param {String} path - the directory we want to write to
+   * @return {Promise<(() => string)|*>}
```
GeorgeChatzigiannis: is that valid jsdoc syntax?
usually the functions are described like `{function(): string}`

jon-shipley: Good spot - we returning a string, not a function

PR #1336, opened 2019-08-29
Feature/35911 pin generation feature toolkit update

GeorgeChatzigiannis:
Page updates:
- [x] Pin generation overview
- [x] Select pupils to generate pins
- [x] View and print pins
- [x] View and print custom pins
- [x] Print Pins


admin/views/partials/_gds_sticky_confirmation_banner.ejs
```diff
@@ -9,6 +9,15 @@
                 <button id="stickyConfirm" data-prevent-double-click="true" class="govuk-button govuk-form-button govuk-!-margin-bottom-0" type="button" data-module="govuk-button">
                     <%= submitLabel %>
                 </button>
+                <% if (typeof submitPrint !== 'undefined') { %>
+                    <script>
+                      document.querySelector('#stickyConfirm').addEventListener('click', function(event) {
```
jon-shipley: Inline scripts should be avoided.  Can we move this out please?

admin/views/pupil-pin/generate-pins-list.ejs
```diff
@@ -1,107 +1,99 @@
...
+                </a>
+            </div>
+        </div>
     </div>
-</div>
+</main>
 <script>
-  $(function () {
-    window.GOVUK.tableSort.applySorting(window.document, 'generatePins', {
+  document.addEventListener('DOMContentLoaded', function() {
```
jon-shipley: This ought to be moved out too.

admin/views/pupil-pin/view-and-print-pins.ejs
```diff
@@ -1,124 +1,121 @@
...
-
+</main>
 <script>
- $(document).ready(function () {
-   $('#printPinsButton').click(function (event) {
-     event.preventDefault()
-     window.print();
-   })
- })
+document.addEventListener('DOMContentLoaded', function() {
```
jon-shipley: This ought to be moved out too.

PR #1339, opened 2019-09-02
Feature/36196 fix jsdoc issues on admin/services/data-access

GuyHarwood:
down to 46 issues


admin/services/data-access/sce.data.service.js
```diff
@@ -37,15 +37,15 @@ sceDataService.sqlFindSceSchools = async () => {
   FROM ${sqlService.adminSchema}.[school]
   LEFT JOIN ${sqlService.adminSchema}.[sce]
     ON sce.school_id = school.id
-  WHERE school.leaCode = 702 
+  WHERE school.leaCode = 702
```
jon-shipley: I realise this is existing code - but why is there a magic number in there?

GuyHarwood: that's bizarre!

GuyHarwood: @GeorgeChatzigiannis any idea on this one?

GeorgeChatzigiannis: seems like all the sce schools have a lea code of 702 as per https://github.com/DFEAGILEDEVOPS/MTC/pull/1064/files

Aside from the magic number I don't really like the naming convention used on this service as well.

GuyHarwood: ok, ill refactor that out to a variable to make it obvious.  thanks

PR #1350, opened 2019-09-17
Design/pupil state flags

GuyHarwood:
Ground work for pupil state optimisations.  The intention is to completely replace the pupil status table.

- migrations to include state flags on `mtc_admin.pupil` and `mtc_admin.check`
- updated and re-organised readme files
- sequence diagrams for each pupil check journey


admin/data/sql/migrations/20190918140848.do.pupil-state.sql
```diff
@@ -0,0 +1,21 @@
+alter table mtc_admin.pupil
+	add currentCheckId int
+
+alter table mtc_admin.pupil
+	add checkComplete bit
+	constraint DF_checkComplete_Default default (0)
+
+alter table mtc_admin.pupil
+	add restartActive bit
```
jon-shipley: Just reading this cold I am not sure whether `restartActive` indicates that permission for a restart has been granted, or whether a restart has been granted *and* the pin has been generated, or something else.

Of course, it's a good name, so perhaps I will learn more later on in the PR.

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
```diff
@@ -0,0 +1,21 @@
...
+	constraint DF_restartActive_Default default (0)
+
+alter table mtc_admin.pupil
+	add attendanceId int
```
jon-shipley: Will we still be able to hold the history as teachers can remove the not attending flag?

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
```diff
@@ -0,0 +1,21 @@
...
+
+alter table mtc_admin.pupil
+	add constraint pupil_check_id_fk
+		foreign key (currentCheckId) references mtc_admin.[check]
+
+alter table mtc_admin.pupil
+	add constraint pupil_attendanceCode_id_fk
+		foreign key (attendanceId) references mtc_admin.attendanceCode
```
jon-shipley: I am slightly surprised.  Linking directly to the `attendanceCode` removes the pupil-attendance join table so that it will be quicker.   But don't we need to know any context about this?  At the moment we store the user that made the non-attendance record and the timestamp.  Are these no longer required, or will they be stored elsewhere?

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
```diff
@@ -0,0 +1,21 @@
+alter table mtc_admin.pupil
+	add currentCheckId int
```
jon-shipley: 👍

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
GuyHarwood: Good point, i will add a readme to supplement the changes.

restartActive is populated when the restart is created.  This serves to indicate that the pupil can be included in the list of pupils eligible for pin generation.

prior to this, the following will happen...
- a restart record will be created in pupilRestart
- the restart id will be added to the check that is now void
- this `restartActive` flag will be set to true

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
GuyHarwood: yes.  pupilAttendance will be retained and serve as a history table.  This simply gives us an optimal way to know whether to include the pupil in pin generation or not, and to see what the attendance reason is.

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
GuyHarwood: see above.  same strategy as restart - pupilAttendance is retained as a history log, this merely serves as a way to query the pupils state without partitioning the attendance table into the register/status page queries.

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
jon-shipley: That's great 😄 

I wonder if I might suggest another name then...`restartPermission` or `permissionToRestart` ?  As it _seems_ to be a permission to generate another pin despite the pupil already having a check in an end-state.

If this was _code_ we would most likely prefix it with `has` or `is` to indicate that it is a boolean.

I also like the fact that with this system a DBA could grant permission for an extra restart in the case where all system-allowed restarts had been used, by following the above steps manually.  The pin generation code would show the pupil in the list just on the basis of the flag. 👏

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
jon-shipley: That's great 😄

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
jon-shipley: That works for me 😄

admin/data/sql/migrations/20190918140848.do.pupil-state.sql
GeorgeChatzigiannis: @GuyHarwood wouldn't it be better to add the restart id to the new check so that it is indicated that this check is a restart?

GuyHarwood: then you have this dangling restart waiting for a new check.  I have never liked the existing design, as it requires looking up the restart when you generate a pin.  The restart is essentially just a token that grants a new check to be taken, and has more to do with the existing check being voided.

GuyHarwood: @jon-shipley @GeorgeChatzigiannis how about `restartAvailable` instead of `restartActive` ?

GeorgeChatzigiannis: Personally I like `restartActive` as it points to the fact that this pupil has a restart that can be consumed/used.
Perhaps we could make it a bit more 'boolean-looking' by naming it hasActiveRestart but I still prefer `restartActive` over all the recommendations.

GuyHarwood: i prefer `restartAvailable` as that is more indicative of something that is ready to use. 🙂

admin/data/sql/migrate-sql.js
```diff
@@ -80,6 +80,7 @@ const runMigrations = async (version) => {
     logger.info('SQL Migrations complete')
   } catch (error) {
     logger.error('Error executing migration...')
+    console.dir(error)
```
GeorgeChatzigiannis: needs to be removed?

GuyHarwood: good catch, yes

GuyHarwood: updated

admin/data/sql/migrations/20190918140901.do.check-state.sql
```diff
@@ -0,0 +1,21 @@
...
+alter table mtc_admin.[check]
+	add complete bit not null
+  constraint DF_complete_Default default 0
+
+alter table mtc_admin.[check]
+	add completedAt datetimeoffset(3)
+
+alter table mtc_admin.[check]
+	add processingFailed bit not null
+  constraint DF_processingFailed_Default default 0
```
jon-shipley: If processing has failed do we record why it failed? E.g. failed validation field X missing.  This is not for teachers but for support/classification and reporting.

design/pupil-state.md
```diff
@@ -3,6 +3,19 @@
...
+
+- can take check: can have a pin & check activated by a teacher
+- taking check: pin & check allocated
+- check received: submitted check received, awaiting processing
+- complete: check is valid and marked.  no further action necessary
+- processing failed: submitted check could not be processed
+- restart available: teacher has activated restart for pupil, not yet consumed
+- restart consumed: pupil logged in and collected the check made available by restart
+  - is this essential? could we not allow a restart even if not logged in? as recording login to sql is expensive
+- not attending: pupil will not be taking the check
```
jon-shipley: Seems like Check not received is missing?

admin/data/sql/migrations/20190918140901.do.check-state.sql
GuyHarwood: yes, that is in receivedCheck (table storage).

design/pupil-state.md
GuyHarwood: i'm working on this right now, off the back of the conversation we had in the restarts channel.

effectively we need an 'expired' flag

jon-shipley: Expired is separate to Not Received.  Not receives requires a restart.

GuyHarwood: yes.  the expiry service needs to accommodate the fact that submitted checks may be in a queue waiting to be acknowledged as received.   The initial idea is to wait for until the next day, and if it still has not been marked as received, then it should be expired.  We can never be sure whether they did in fact submit, or just closed the check.  See `design/pupil-state.md` for notes on this situation.

PR #1351, opened 2019-09-17
Feature/36409 analyse files for anomaly report

jon-shipley:
Refactor anomaly report to be run from file

First two detections


functions-app/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "GPL-3.0",
   "scripts": {
     "start": "func host start --port 7072",
-    "test": "yarn lint && yarn run nyc -r html jasmine **/*.spec.js */**/*.spec.js",
+    "test": "yarn lint && yarn run nyc -r html jasmine **/*.spec.js */**/*.spec.js */**/**/*.spec.js",
```
GuyHarwood: would `**/*.spec.js` not cover all your subdirectories in this particular case or am i overlooking something subtle?

functions-app/package.json
```diff
@@ -25,6 +25,7 @@
     "mssql": "^5.1.0",
     "object-hash": "^1.3.1",
     "ramda": "^0.26.1",
+    "ramda-adjunct": "^2.20.0",
```
GeorgeChatzigiannis: do we really need this package? methods used can be replaced with vanilla ones?

functions-app/report-psychometrician/service/detections/get-check-started-date.js
```diff
@@ -0,0 +1,24 @@
...
+const R = require('ramda')
+
+function getCheckStartedDateFromAuditLog (payload) {
+  if (typeof payload !== 'object') return undefined
+  const audits = R.prop('audit', payload)
+  if (!audits) return undefined
+  const checkStartedEvent = R.find(R.propEq('type', 'CheckStarted'), audits)
+  if (!checkStartedEvent) return undefined
+  const ts = R.prop('clientTimestamp', checkStartedEvent)
+  if (!ts) return undefined
```
GeorgeChatzigiannis: is specifically undefined required here as a returned value?

functions-app/report-psychometrician/service/detections/remove-duplicate-anomaly-reports.js
```diff
@@ -0,0 +1,23 @@
...
+ * For example, perhaps because multiple inputs were recorded before the question was shown
+ * @usage removeDuplicates(anomalyReports)
+ */
+const removeDuplicates = R.reduce(
+  (accumulator, newElement) => {
+    const props = ['Question number', 'Message']
+    const found = R.filter(e => R.equals(
+      R.pick(props, e),
+      R.pick(props, newElement)),
+    accumulator)
```
GeorgeChatzigiannis: 🏅

functions-app/package.json
jon-shipley: It was needed as due to the extra level of sub-directory nesting with the 'detections' directory.  So, apparently not.

GuyHarwood: that's odd.  i was under the impression that the first one `**/*.js` covers all recursively

functions-app/package.json
jon-shipley: I suspect the answer, at the moment, is that we could leave it behind. And yet, it offers such a wide range of useful ancillary functions to Ramda that it seems logical to include the library, especially as we get such great use from ramda for the processing piece.

We have many more detections to add, these will most likely make use of ramda so I want to include at this time.  Please note that ramda-adjunct does not have any dependencies other than ramda.

functions-app/report-psychometrician/service/detections/get-check-started-date.js
jon-shipley: What would you suggest?

functions-app/package.json
GeorgeChatzigiannis: Ok then it does make sense to keep it.

functions-app/report-psychometrician/service/detections/get-check-started-date.js
GeorgeChatzigiannis: Wouldn't plain return be sufficient?I'd expect the caller to handle it properly.

jon-shipley: Adjusted as you suggest :)

PR #1355, opened 2019-09-24
Feature/36273 service bus queue creation util script

GuyHarwood:
- [x] node script to create & delete service bus queues
- [x] set queue policies on creation
- [x] add connection string to key vault in dev environment
- [x] add connection string to travis as secure env var
- [x] call this script from travis to ensure environment is in sync
- [x] update ./start.sh, ./stop.sh and ./restart.sh with queue purge


deploy/service-bus/create.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  })
+}))
+
+async function main () {
+  try {
+    const promises = queues.map(q => createQueue(q, defaultQueueOptions))
+    await Promise.all(promises)
+  } catch (error) {
+    throw new Error(`Error creating queue: ${error.message}`)
```
jon-shipley: Not sure that catching the error only to throw a new error is adding much.   Do we really need the try/catch here?

deploy/service-bus/delete.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+  })
+}))
+
+async function main () {
+  try {
+    const promises = queues.map(q => deleteQueue(q))
+    await Promise.all(promises)
+  } catch (error) {
+    throw new Error(`Error deleting queue: ${error.message}`)
```
jon-shipley: Is this throw required?

deploy/service-bus/create.js
GuyHarwood: good catch (see what i did there)

PR #1358, opened 2019-09-26
fix dot net version

activemq:


GuyHarwood:
@activemq can be merged as soon as travis passes 🏆

PR #1359, opened 2019-09-26
feature/36389 check marker

GuyHarwood:
- [x] define unit tests for implementation
- [x] re-use existing marking service code where applicable
- [x] check validator now persists answers to entity so marker does not have to decompress and validate
- [x] marking errors are persisted against receivedCheck object
- [x] re-used JS is modified to follow strict TS policies
- [x] notifications are sent to check notification queue when check mark finishes/fails

## Notes
- the persistence of the answers to sql has not been retained.  This is because we are looking at an alternative to SQL which will be covered in another ticket.
- the lookup of the check form still talks to SQL.  This could change once we have implemented pin & form allocation.
- the notification message is intentionally primitive, as requirements are to be refined as check notifier requirements mature.


func-complete-check/check-marker/check-marker.v1.ts
```diff
@@ -0,0 +1,166 @@
...
+declare interface MarkingData {
+  answers: Array<Answer>
+  formQuestions: Array<CheckFormQuestion>
+  results: any
+}
+
+declare interface Mark {
+  mark: number
+  maxMarks: number
+  processedAt: Date
```
jon-shipley: I am surprised to maxMarks on a Mark object.  And that the question is not represented.  I will read on...

func-complete-check/check-marker/check-marker.v1.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  async mark (functionBindings: ICheckMarkerFunctionBindings): Promise<void> {
+
+    const validatedCheck = this.findValidatedCheck(functionBindings.receivedCheckTable)
+    const markingData = await this.validateData(functionBindings, validatedCheck)
+    if (markingData === undefined) {
+      functionBindings.checkNotificationQueue.push({
+        checkCode: 'checkCode',
```
jon-shipley: I may be missing something here - but this message doesn't seem to have the actual checkCode in the message.

func-complete-check/check-marker/index.ts
```diff
@@ -1,8 +1,28 @@
...
+const functionName = 'check-marker'
+import * as V1 from './check-marker.v1'
 
 const serviceBusQueueTrigger: AzureFunction = async function (context: Context, markCheckMessage: MarkCheckMessageV1): Promise<void> {
-  context.log.warn(`check marking not yet implemented`)
+  const start = performance.now()
+  const version = markCheckMessage.version
+  context.log.info(`${functionName}: version:${version} message received for checkCode ${markCheckMessage.checkCode}`)
+  try {
+    if (version !== '1') {
```
jon-shipley: I am slightly concerned that the number could be passed as a number rather than a string.  Perhaps we could mitigate it with:

```suggestion
    if (Number(version) !== 1) {
```

func-complete-check/lib/to-bool.ts
```diff
@@ -0,0 +1,9 @@
+export function primitiveToBoolean (value: string | number | boolean | null | undefined): boolean {
+  if (value === 'true') {
```
GeorgeChatzigiannis: @GuyHarwood why not using the library `To-Bool`?

func-complete-check/check-validator/check-validator.v1.ts
```diff
@@ -38,18 +34,19 @@ export class CheckValidatorV1 implements ICheckValidator {
...
       this.validateCheckStructure(checkData)
     } catch (error) {
       await this.setReceivedCheckAsInvalid(error.message, receivedCheck)
       logger.error(error.message)
       return
     }
 
-    await this.setReceivedCheckAsValid(receivedCheck)
+    await this.setReceivedCheckAsValid(receivedCheck, checkData)
     logger.info(`receivedCheck successfully validated`)
```
jon-shipley: This message will be useless in production without a checkCode.  It might be better at `debug` level or removed.

func-complete-check/check-marker/check-marker.v1.ts
```diff
@@ -0,0 +1,166 @@
...
+      return
+    }
+
+    if (!RA.isArray(parsedAnswersJson)) {
+      await this.updateReceivedCheckWithMarkingError(validatedCheck, 'answers data is not an array')
+      return
+    }
+
+    const checkCode = validatedCheck.RowKey
+    const rawCheckForm = await this._sqlService.getCheckFormDataByCheckCode(checkCode)
```
GeorgeChatzigiannis: what if RowKey is not present in the validatedCheck?
shouldn't we prevent this transaction to happen in that case, call `updateReceivedCheckWithMarkingError` and return?

func-complete-check/lib/storage-helper.ts
```diff
@@ -4,41 +4,32 @@ import bluebird from 'bluebird'
 
 export interface IAsyncTableService {
   replaceEntityAsync (table: string, entity: any): Promise<any>
-  queryEntitiesAsync (table: string, tableQuery: az.TableQuery, currentToken?: az.TableService.TableContinuationToken): Promise<any>
-  deleteEntityAsync (table: string, entityDescriptor: any): Promise<any>
-  insertEntityAsync (table: string, entityDescriptor: unknown, options?: az.TableService.InsertEntityRequestOptions): Promise<any>
+  // queryEntitiesAsync (table: string, tableQuery: az.TableQuery, currentToken?: az.TableService.TableContinuationToken): Promise<any>
```
jon-shipley: Why comment them out?  Looks like they are not needed anymore.

func-complete-check/lib/storage-helper.ts
```diff
@@ -4,41 +4,32 @@ import bluebird from 'bluebird'
...
+        if (error) {
+          reject(error)
+        } else {
+          resolve(result)
+        }
+      })
+    })
   }
-  queryEntitiesAsync (table: string, tableQuery: az.TableQuery, currentToken?: az.TableService.TableContinuationToken | undefined): Promise<any> {
+/*   queryEntitiesAsync (table: string, tableQuery: az.TableQuery, currentToken?: az.TableService.TableContinuationToken | undefined): Promise<any> {
```
jon-shipley: Can this be removed?

func-complete-check/lib/to-bool.ts
GuyHarwood: because it is not supported by typescript, nor does it have a `@types` file.  it behaves differently with typescript.  i spent a while working on it.

func-complete-check/check-marker/check-marker.v1.ts
GuyHarwood: thats the whole point of the check validator - the marker only receives it once its been validated and doesn't repeat all the work that the validator has done.

func-complete-check/check-marker/index.ts
GuyHarwood: this is the same problem i encountered with the other TS functions - if i type it as a number, the binding/message still brings it in as a string directly off the message.  Happy to look at workarounds as it doesn't feel right.

func-complete-check/check-marker/check-marker.v1.ts
GuyHarwood: yep, that's definitely a typo.  there is nothing consuming those messages, as we have not implemented the check notification service yet.  will fix

func-complete-check/check-marker/check-marker.v1.ts
GuyHarwood: the Mark object represents the overall mark.  is this a wording/naming issue?  regards answers - they will be stored row by row, likely in table storage.  although that is a separate decision

GuyHarwood: as discussed, this is total mark.  answers will be persisted in separate ticket

jon-shipley: My bad.  Thanks for the explanation.

func-complete-check/check-marker/check-marker.v1.ts
```diff
@@ -0,0 +1,176 @@
...
+  factor2: number
+  answer: string
+  sequenceNumber: number
+  question: string
+  clientTimestamp: string
+}
+
+declare interface CheckFormQuestion {
+  f1: number
+  f2: number
```
jon-shipley: A question has only factor1 and factor2, but a CheckFormQuestion should also have a sequence - e.g. a CheckForm is a series of questions asked in a particular order.

GuyHarwood: raw JSON from checkForm.formData...

```
[ 
        {
            "f1" : 2,
            "f2" : 5
        }, 
        {
            "f1" : 11,
            "f2" : 2
        }, 
        {
            "f1" : 5,
            "f2" : 10
        }, 
        {
            "f1" : 4,
            "f2" : 4
        }, 
        {
            "f1" : 3,
            "f2" : 9
        }, 
        {
            "f1" : 2,
            "f2" : 4
        }, 
        {
            "f1" : 3,
            "f2" : 3
        }, 
        {
            "f1" : 4,
            "f2" : 9
        }, 
        {
            "f1" : 6,
            "f2" : 5
        }, 
        {
            "f1" : 12,
            "f2" : 12
        }
    ]
```

GuyHarwood: any enhancements to this can be formulated into the ticket i created and shared with you earlier.

func-complete-check/check-marker/check-marker.v1.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    let questionNumber = 1
+    for (let question of markingData.formQuestions) {
+      const currentIndex = questionNumber - 1
+      const answerRecord = markingData.answers[currentIndex]
+      const answer = (answerRecord && answerRecord.answer) || ''
+      questionNumber += 1
+
+      if (answer && question.f1 * question.f2 === parseInt(answer, 10)) {
```
GeorgeChatzigiannis: wouldn't this be more readable as a method? `private isAnswerCorrect..`

PR #1360, opened 2019-09-30
Feature/add new anomaly report detection check took too long

jon-shipley:
Add new anomaly detection for checks that take longer than expected


functions-app/report-psychometrician/service/detections/detect-check-took-too-long.js
```diff
@@ -0,0 +1,41 @@
...
+  const numberOfQuestions = R.length(markedAnswers) || 0
+
+  // Calculate the max total time allowed for the check
+  const maxCheckSeconds = (numberOfQuestions * config.loadingTime) +
+    (numberOfQuestions * config.questionTime) +
+    (config.questionReader ? numberOfQuestions * 2.5 : 0)
+
+  const checkStartedDate = getCheckStartedDate(
+    null, // always get the date from the audit log
+    R.prop('checkPayload', data)
```
GeorgeChatzigiannis: shouldn't that be R.pathOr to provide a default if `checkPayload` isn't present?

jon-shipley: If the payload is not present the implementation of getCheckStartedDate will work correctly.  In fact, passing in undefined makes it easier to check than an empty object.

PR #1361, opened 2019-09-30
Feature/36963 Helpdesk impersonation & home page

GeorgeChatzigiannis:
Introduced helpdesk site that includes impersonation form and helpdesk home page.
Helpdesk user can remove school session data by clicking on remove impersonation on home page.

Screenshot 2019-09-30 at 10 05 57

Screenshot 2019-09-30 at 10 07 47


admin/authentication/local-strategy.js
```diff
@@ -28,7 +28,7 @@ module.exports = async function (req, email, password, done) {
     // school with dfeNumber 1234567.
     // NOTE that this is not a generic implementation: you actually need to be called 'helpdesk' or 'service-manager'
     // rather than simply have the role.  This is fine - it's only for local-dev
-    const matches = email.match(/^(helpdesk|service-manager):(\d{7})$/)
+    const matches = email.match(/^(service-manager):(\d{7})$/)
```
GeorgeChatzigiannis: @GuyHarwood @jon-shipley I removed the feature to directly impersonate a school from login page as this really serves no purpose now that the helpdesk site exists. Happy to discuss this further if needed.

admin/controllers/helpdesk.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Renders the helpdesk (role) landing page
+ * @param {object} req
+ * @param {object} res
+ * @param {function} next
+ * @returns {Promise.<void>}
+ */
+controller.getHelpdeskHome = async (req, res, next) => {
```
GuyHarwood: is this meant to be a helpdesk version of the school home page?  if so, i would expect us to simply use the existing one.

admin/authentication/local-strategy.js
GuyHarwood: makes sense

admin/controllers/helpdesk.js
GeorgeChatzigiannis: @GuyHarwood there are going to be additional features appearing on this home page.

GuyHarwood: ok, so why don't we just have an optional partial for the PIN history link that is only rendered when its helpdesk?  that feels easier than duplicating all of the ongoing changes that we will make to the school homepage...

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+const R = require('ramda')
+
+const schoolDataService = require('../services/data-access/school.data.service')
+const schoolImpersonationEmptyValueValidator = require('../lib/validator/school-impersonation/school-impersonation-empty-value-validator')
```
jon-shipley: I am not sure what an 'EmptyValue' validator is?  Isn't this the SchooImpersonatorFormValidator?  Why should the implementation be a part of the name?

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,48 @@
...
+  let validationError
+  validationError = schoolImpersonationEmptyValueValidator.validate(dfeNumber)
+  if (validationError.hasError()) {
+    return validationError
+  }
+  let school
+  try {
+    school = await schoolDataService.sqlFindOneByDfeNumber(dfeNumber)
+  } catch (error) {}
+  validationError = schoolImpersonationDfeNumberValidator.validate(school)
```
jon-shipley: Why do we need two validators?

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ * Remove impersonation data from request.user object
+ * @param {object} user
+ * @returns {void}
+ */
+schoolImpersonationService.removeImpersonation = (user) => {
+  const impersonationKeys = ['School', 'schoolId', 'timezone']
+  if (user.School || user.schoolId || user.timezone) {
```
jon-shipley: What is the point of the `if` condition here?  I suspect it is not needed as `delete` will always return true.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

(unless deleting a property from the prototype chain; in which case it will throw)

admin/spec/back-end/lib/validator/school-impersonation/school-impersonation-empty-value-validator.spec.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+/* global describe, it expect */
+const schoolImpersonationEmptyValueValidator = require('../../../../../lib/validator/school-impersonation/school-impersonation-empty-value-validator')
+
+describe('schoolImpersonationEmptyValueValidator', function () {
+  it('should not add a validation error if dfeNumber is not empty', function () {
+    const dfeNumber = '1230000'
```
jon-shipley: When do you convert the form input (string) into the required type for the DB lookup (number)?  Don't you also need to validate the post-conversion result to ensure it is an integer?

admin/controllers/helpdesk.js
GeorgeChatzigiannis: @GuyHarwood resolving this as it is now using the school home page

admin/authentication/local-strategy.js
```diff
@@ -49,7 +49,13 @@ module.exports = async function (req, email, password, done) {
 
     const [school, role] = await Promise.all([schoolPromise, rolePromise])
 
-    if (!user || !role || !school) {
+    if (user && user.identifier !== 'helpdesk' && !school) {
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley this has been significantly changed. Resolving this as I am now awaiting new feedback.

admin/controllers/helpdesk.js
```diff
@@ -0,0 +1,89 @@
...
+ * Renders the helpdesk school impersonation page
+ * @param {object} req
+ * @param {object} res
+ * @param {function} next
+ * @param {object} error
+ * @returns {Promise.<void>}
+ */
+controller.getSchoolImpersonation = async (req, res, next, error = null) => {
+  res.locals.pageTitle = 'MTC Helpdesk School Impersonation'
+  schoolImpersonationService.removeImpersonation(req.user)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley this is to ensure the school session data are wiped. It is not a mandatory step as the form itself will populate the session with new school data however I liked the idea of having a clean session object instead of replacing existing key values.

admin/lib/errors/school-impersonation.js
```diff
@@ -0,0 +1,6 @@
+'use strict'
+
+module.exports = {
+  'noInput': 'Please type a dfe number',
+  'incorrectInput': 'The dfe number provided does not match a school'
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley indeed and it is updated.

admin/controllers/helpdesk.js
```diff
@@ -0,0 +1,89 @@
...
+ * Submit helpdesk school impersonation
+ * @param {object} req
+ * @param {object} res
+ * @param {function} next
+ * @returns {Promise.<void>}
+ */
+controller.postSchoolImpersonation = async (req, res, next) => {
+  const { dfeNumber } = req.body
+  try {
+    const outcome = await schoolImpersonationService.validateCreateImpersonation(req.user, dfeNumber)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley done

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley updated as per our earlier discussion

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley the first validator checks if the value is non empty where the second checks if the dfeNumber from the client does match a school. 
In between these two validators there is a database transaction to match the school record.
The alternative would be to have one validator with two different methods.

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley merged to a single validator

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley merged to one as it seems to be cleaner.

admin/controllers/helpdesk.js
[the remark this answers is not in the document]
jon-shipley: Perhaps I wasn't clear.  You are changing state on the results of a GET which shouldn't happen.  You should do a POST to change state.

admin/controllers/helpdesk.js
```diff
@@ -0,0 +1,89 @@
...
+    const outcome = await schoolImpersonationService.validateImpersonationForm(req.user, dfeNumber)
+    if (outcome.hasError && outcome.hasError()) {
+      return controller.getSchoolImpersonation(req, res, next, outcome)
+    }
+    res.redirect('home')
+  } catch (error) {
+    next(error)
+  }
+  return next()
+}
```
jon-shipley: Reading this I can't see where any work actually gets done.

admin/services/data-access/user.data.service.js
```diff
@@ -24,6 +24,36 @@ const userDataService = {
...
+    const sql = `
+      SELECT
+        u.id,
+        u.passwordHash,
+        u.identifier,
+        r.title AS roleName,
+        u.school_id AS schoolId,
+        s.dfeNumber,
+        sce.timezone
+      FROM mtc_admin.${table} u
```
jon-shipley: We have decided to drop this variable and use the string of the table name instead.

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,44 @@
...
+schoolImpersonationService.validateImpersonationForm = async (user, dfeNumber) => {
+  let validationError
+  validationError = schoolImpersonationDfeNumberValidator.isDfeNumberEmpty(dfeNumber)
+  if (validationError.hasError()) {
+    return validationError
+  }
+  let school
+  try {
+    school = await schoolDataService.sqlFindOneByDfeNumber(dfeNumber)
+  } catch (error) {}
```
jon-shipley: What do you think to naming it `ignore` where the catch error variable is unused?

admin/lib/validator/school-impersonation-dfe-number-validator.js
```diff
@@ -0,0 +1,22 @@
...
+  if (!dfeNumber) {
+    validationError.addError('dfeNumber', SchoolImpersonationMessages.noInput)
+    return validationError
+  }
+  return validationError
+}
+
+const isDfeNumberValid = (school) => {
+  let validationError = new ValidationError()
+  if (!school || !school.id) {
```
jon-shipley: `{
  id: 'foo'
}`

passes validation but is not valid.

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,44 @@
...
+  validationError = schoolImpersonationDfeNumberValidator.isDfeNumberValid(school)
```
jon-shipley: Not sure what the point of this statement is?  Either school is undefined from line 21, or it was found in the db.  Why do we need a validator to detect undefined?  Just use `if (school === undefined)`

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,44 @@
...
+  if (validationError.hasError()) {
+    return validationError
+  }
+  user.School = school.dfeNumber
+  user.schoolId = school.id
+  user.timezone = school.timezone
```
jon-shipley: It seems to me that we lose the knowledge that the user is impersonating.  We should be able to know:
* that the user is impersonating
* the true role for the user

Can you confirm?

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+  let school
+  try {
+    school = await schoolDataService.sqlFindOneByDfeNumber(dfeNumber)
+  } catch (error) {}
+  validationError = schoolImpersonationDfeNumberValidator.isDfeNumberValid(school)
+  if (validationError.hasError()) {
+    return validationError
+  }
+  user.School = school.dfeNumber
```
jon-shipley: I don't think that a  function named `validateImpersonationForm` should affect the impersonation.  That would be more likely named `impersonate` or something similar.

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,44 @@
...
+  return user
+}
+
+/**
+ * Remove impersonation data from request.user object
+ * @param {object} user
+ * @returns {void}
+ */
+schoolImpersonationService.removeImpersonation = (user) => {
+  R.forEach(k => delete user[k], ['School', 'schoolId', 'timezone'])
```
jon-shipley: Should we add a message to the user?

admin/views/school/school-home.ejs
```diff
@@ -1,6 +1,16 @@
+
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-        <%- partial('partials/_name_and_sign_out') %>
+        <% if (user.role === 'HELPDESK') { %>
```
jon-shipley: Might be better to remove this magic string.  Something as important as role names might be kept in a frozen object literal.

admin/views/school/school-home.ejs
```diff
@@ -1,6 +1,16 @@
...
+            <div id="sign-out">
+                <a class="govuk-link" href="/helpdesk/school-impersonation">Remove Impersonation</a>
+            </div>
+            <% if (typeof user != 'undefined') { %>
```
jon-shipley:
```suggestion
            <% if (typeof user !== 'undefined') { %>
```

GuyHarwood:
@activemq @sinhaamit1980 this is not ready for testing yet as its not dev complete

admin/controllers/helpdesk.js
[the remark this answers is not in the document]
GuyHarwood: also - what if they want to remain as that school?  the removal of impersonation needs to be explicit.

admin/controllers/helpdesk.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Renders the helpdesk (role) landing page
+ * @param {object} req
+ * @param {object} res
+ * @param {function} next
+ * @returns {Promise.<void>}
+ */
+controller.getSchoolLandingPage = async (req, res, next) => {
```
GuyHarwood: i thought the duplicate of the school home page was being removed in favour of reusing the existing school home page? or is this a different page?

admin/controllers/helpdesk.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley @GuyHarwood considering this indeed should have been an isolated action this has now changed and is an explicit POST route.

admin/spec/back-end/lib/validator/school-impersonation/school-impersonation-empty-value-validator.spec.js
GeorgeChatzigiannis: @jon-shipley resolving this as the caller service implementation and naming has changed.

admin/controllers/helpdesk.js
GeorgeChatzigiannis: @jon-shipley changed as described on the other comment

admin/services/data-access/user.data.service.js
GeorgeChatzigiannis: @jon-shipley updated to use the table name.

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley as discussed this is updated.

admin/lib/validator/school-impersonation-dfe-number-validator.js
GeorgeChatzigiannis: @jon-shipley at that stage the school either has been fetched from the database or it doesn't exist at all.if a 'foo' value is stored as id that would be fundamentally wrong.

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley knowledge so far about if the user is already impersonating is not needed and the role itself is exposed already in the authentication part.

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley changed as per discussion.

admin/views/school/school-home.ejs
GeorgeChatzigiannis: @jon-shipley replaced as per our discussion and your feedback.

admin/controllers/helpdesk.js
GeorgeChatzigiannis: @GuyHarwood this has been updated to use school controller.

admin/authentication/middleware.js
```diff
@@ -2,18 +2,22 @@
...
+      logger.debug(`checking authorisation on ${req.url} for role:${roles} against userRole:${userRole}`)
+      // declare single role variable if input is not array of roles
+      const role = !Array.isArray(roles) && roles
+      if (role === undefined ||
+        (role !== undefined && userRole && role === userRole) ||
+        (Array.isArray(roles) && roles.some(r => r === userRole))) {
         return next()
       } else {
-        logger.warn(`could not authorise ${role} against userRole:${userRole}, UserName:${req.user.UserName} ID:${req.user.id}`)
+        logger.warn(`could not authorise ${roles} against userRole:${userRole}, UserName:${req.user.UserName} ID:${req.user.id}`)
```
GuyHarwood: how does `roles` render if it's an array?

admin/services/helpdesk.service.js
```diff
@@ -0,0 +1,20 @@
...
+
+const roles = require('../lib/consts/service-roles.js')
+const helpdeskService = {}
+
+/**
+ * Identify if helpdesk user has not received impersonation
+ * @param {object} user
+ * @returns {boolean}
+ */
+helpdeskService.hasHelpdeskNotReceivedImpersonation = (user) => {
```
GuyHarwood: that method name doesn't read very well.  maybe flip it to `isImpersonatingSchool()` ?

admin/services/school-impersonation.service.js
```diff
@@ -12,24 +12,29 @@ const schoolImpersonationService = {}
  * @param {string} dfeNumber
  * @returns {Object}
  */
-schoolImpersonationService.validateImpersonationForm = async (user, dfeNumber) => {
+schoolImpersonationService.processImpersonationForm = async (user, dfeNumber) => {
```
GuyHarwood: again, odd choice of method name.  maybe `setSchoolImpersonation()` ?

admin/controllers/school.js
```diff
@@ -20,6 +21,11 @@ const controller = {}
 controller.getSchoolLandingPage = async (req, res, next) => {
   res.locals.pageTitle = 'School Homepage'
   try {
+    const hasHelpdeskNotReceivedImpersonation = helpdeskService.hasHelpdeskNotReceivedImpersonation(req.user)
```
jon-shipley: I think naming things in the negative is generally harder to understand.  Would be better as `!isHelpdeskImpersonator()` IMHO.  Happy to discuss.

admin/lib/consts/service-roles.js
```diff
@@ -0,0 +1,6 @@
+module.exports = Object.freeze({
+  helpdesk: 'HELPDESK',
+  serviceManager: 'TEACHER',
+  teacher: 'TEACHER',
+  testDeveloper: 'TEST-DEVELOPER'
+})
```
jon-shipley: Whilst it is much better having the object 👍 I think the serviceManager role needs to be set to SERVICE-MANAGER 😉

admin/roles-config.js
```diff
@@ -1,8 +1,10 @@
 module.exports = {
   ROLE_SERVICE_MANAGER: 'SERVICE-MANAGER',
   ROLE_TEST_DEVELOPER: 'TEST-DEVELOPER',
+  ROLE_HELPDESK: 'HELPDESK',
```
jon-shipley: Not sure what this object is doing.  It's a mess - some sort of combination of roles and home pages? 😱

admin/routes/helpdesk.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const express = require('express')
+const router = express.Router()
+const isAuthenticated = require('../authentication/middleware')
+const rolesConfig = require('../roles-config')
+const helpdeskController = require('../controllers/helpdesk')
+
+router.get('/', isAuthenticated(rolesConfig.ROLE_HELPDESK), (req, res, next) => helpdeskController.getSchoolImpersonation(req, res, next))
```
jon-shipley: Shouldn't we be using the new roles.js object that defines all the roles in the system?  Prefix with ROLE seems superfluous to me.

admin/lib/consts/service-roles.js
```diff
@@ -0,0 +1,6 @@
+module.exports = Object.freeze({
```
jon-shipley: This files should just be named roles.js - it should be all roles in the system.

admin/services/school-impersonation.service.js
jon-shipley: @GeorgeChatzigiannis - do you think it is worth adding a flash message for the user?

admin/views/helpdesk/school-impersonation.ejs
```diff
@@ -0,0 +1,68 @@
...
+                                    <li><a href="#<%= f %>"><%= errors %></a></li>
+                                <% } %>
+                            <% }) %>
+                        </ul>
+                    </div>
+                </div>
+            <% } %>
+            <h1 class="govuk-heading-xl"><%= pageTitle %></h1>
+            <%- partial('partials/_gds_session_expiry') %>
+            <form action="/helpdesk/add-school-impersonation" id="schoolImpersonationForm" method="post">
```
jon-shipley: I think it would be better to prevent browsers caching dfeNumbers here.  In normal operation, this could be used hundreds of times a day.  I think it would be safer to make the user type out the whole number.

`autocomplete=off`

admin/roles-config.js
GuyHarwood: this is one thing i forgot to bring up - there should be just one canonical set of declared roles in the admin app

admin/authentication/middleware.js
GeorgeChatzigiannis: @GuyHarwood changed so that it displays the message correctly regardless of string or array type.

admin/authentication/middleware.js
```diff
@@ -2,18 +2,22 @@
...
     if (req.isAuthenticated()) {
       let userRole
       if (req.user) {
         userRole = ((req.user).role || {})
       }
-      logger.debug(`checking authorisation on ${req.url} for role:${role} against userRole:${userRole}`)
-      if (role === undefined || (role !== undefined && userRole && role === userRole)) {
+      logger.debug(`checking authorisation on ${req.url} for role:${roles} against userRole:${userRole}`)
+      // declare single role variable if input is not array of roles
+      const role = !Array.isArray(roles) && roles
```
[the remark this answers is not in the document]
GuyHarwood: also - why not add tests around this?  its a critical part of the system

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @GuyHarwood updated

admin/services/helpdesk.service.js
GeorgeChatzigiannis: @GuyHarwood this method has now been split to 2 different ones for readability and simplification.

admin/services/helpdesk.service.js
```diff
@@ -4,17 +4,21 @@ const roles = require('../lib/consts/service-roles.js')
...
+  return user.role === roles.helpdesk
+}
+
+/**
+ * Identify if current user has impersonation properties set
+ * @param {object} user
+ * @returns {boolean}
+ */
+helpdeskService.isImpersonating = (user) => {
+  return user.School && user.schoolId && typeof user.timezone === 'string'
```
GuyHarwood: i would just do a typeof check for number on schoolId.  its more explicit

admin/controllers/school.js
GeorgeChatzigiannis: @jon-shipley this is now simplified and split to two methods.

admin/lib/consts/service-roles.js
GeorgeChatzigiannis: oh wow that was a good one. resolved!

admin/routes/helpdesk.js
GeorgeChatzigiannis: @jon-shipley I could tackle that on the follow up PBI. seems like this piece of work is already doing too much.

admin/lib/consts/service-roles.js
GeorgeChatzigiannis: @jon-shipley done

admin/views/helpdesk/school-impersonation.ejs
GeorgeChatzigiannis: @jon-shipley that's a fair point. I have added the attribute.

admin/services/helpdesk.service.js
GeorgeChatzigiannis: @GuyHarwood fair point. amended

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley it's already included. Please have a look at the helpdesk controller method `postRemoveSchoolImpersonation`.

admin/roles-config.js
GeorgeChatzigiannis: @GuyHarwood @jon-shipley already made a note for the follow-up about this.

admin/authentication/middleware.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @GuyHarwood @jon-shipley please have another look on this.

admin/authentication/middleware.js
```diff
@@ -2,18 +2,20 @@
...
+      let roles = typeof arg === 'string' && arg.length > 1
+        ? [arg] : arg
       let userRole
       if (req.user) {
         userRole = ((req.user).role || {})
       }
-      logger.debug(`checking authorisation on ${req.url} for role:${role} against userRole:${userRole}`)
-      if (role === undefined || (role !== undefined && userRole && role === userRole)) {
+      logger.debug(`checking authorisation on ${req.url} for roles:${roles && roles.join()} against userRole:${userRole}`)
+      if (roles === undefined || (Array.isArray(roles) && roles.some(r => r === userRole))) {
```
jon-shipley: roles will now always be an array so this can be simplified.  userRole should be a valid role or undefined.

admin/controllers/school.js
```diff
@@ -20,6 +21,12 @@ const controller = {}
 controller.getSchoolLandingPage = async (req, res, next) => {
   res.locals.pageTitle = 'School Homepage'
   try {
+    const isHelpdeskRole = helpdeskService.isHelpdeskRole(req.user)
+    const isImpersonating = helpdeskService.isImpersonating(req.user)
+    // Prevent helpdesk users from accessing school home when impersonation data are not populated
+    if (isHelpdeskRole && !isImpersonating) {
```
jon-shipley: 👍 clear

admin/lib/consts/roles.js
```diff
@@ -0,0 +1,6 @@
+module.exports = Object.freeze({
+  helpdesk: 'HELPDESK',
+  serviceManager: 'SERVICE-MANAGER',
+  teacher: 'TEACHER',
+  testDeveloper: 'TEST-DEVELOPER'
+})
```
jon-shipley: 👍

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,49 @@
...
+const schoolImpersonationService = {}
+
+/**
+ * Validate and assign school to helpdesk session
+ * @param {object} user
+ * @param {string} dfeNumber
+ * @returns {Object}
+ */
+schoolImpersonationService.setSchoolImpersonation = async (user, dfeNumber) => {
+  let validationError
```
jon-shipley: Why can't this be const?

admin/services/school-impersonation.service.js
```diff
@@ -0,0 +1,49 @@
...
+  // returns a validation error if dfeNumber provided is empty
+  if (validationError.hasError()) {
+    return validationError
+  }
+  let school
+  try {
+    school = await schoolDataService.sqlFindOneByDfeNumber(dfeNumber)
+  } catch (ignore) {}
+  // returns a validation error if the school record is not valid
+  validationError = schoolImpersonationValidator.isSchoolRecordValid(school)
```
jon-shipley: This is another variable, e.g. schoolValidationError.  Should be const.

admin/spec/back-end/authentication/middleware.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      const req = httpMocks.createRequest(reqParams)
+      spyOn(req, 'isAuthenticated').and.returnValue(true)
+      spyOn(res, 'redirect')
+      const allowedRoles = ['SERVICE-MANAGER', 'TEACHER']
+      const func = isAuthenticated(allowedRoles)
+      func(req, res, next)
+      expect(next).not.toHaveBeenCalled()
+      expect(res.redirect).toHaveBeenCalledWith('/unauthorised')
+    })
+    it('calls next if the role passed is undefined while the request is authenticated', () => {
```
jon-shipley: Also need same test for null, empty string, function, and object types.

admin/spec/back-end/authentication/middleware.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      const req = httpMocks.createRequest(reqParams)
+      spyOn(req, 'isAuthenticated').and.returnValue(false)
+      spyOn(res, 'redirect')
+      const func = isAuthenticated(undefined)
+      func(req, res, next)
+      expect(next).not.toHaveBeenCalled()
+      expect(res.redirect).toHaveBeenCalledWith('/sign-in')
+    })
+  })
+})
```
jon-shipley: Should be a test where the userRole is not defined.

admin/authentication/middleware.js
```diff
@@ -2,18 +2,20 @@
 
 const logger = require('../services/log.service').getLogger()
 
-function isAuthenticated (role) {
+function isAuthenticated (arg) {
   return function (req, res, next) {
     if (req.isAuthenticated()) {
+      let roles = typeof arg === 'string' && arg.length > 1
```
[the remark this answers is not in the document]
GuyHarwood: Ramda?

admin/authentication/middleware.js
GuyHarwood: ramda contains?

admin/spec/back-end/authentication/local-strategy.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      spyOn(userDataService, 'sqlFindUserInfoByIdentifier')
+      spyOn(adminLogonEventDataService, 'sqlCreate')
+      const doneFunc = (err, res) => err || res
+      await validateAndSave(req, 'teacher1', 'password', doneFunc)
+      expect(adminLogonEventDataService.sqlCreate).toHaveBeenCalledWith(
+        { sessionId: 1, body: '{}', remoteIp: 'remoteAddress', userAgent: undefined, loginMethod: 'local', errorMsg: 'Invalid user', isAuthenticated: false }
+      )
+    })
+  })
+})
```
GuyHarwood: 🏆

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/services/school-impersonation.service.js
GeorgeChatzigiannis: @jon-shipley updated

admin/spec/back-end/authentication/middleware.spec.js
GeorgeChatzigiannis: @jon-shipley done

admin/spec/back-end/authentication/middleware.spec.js
GeorgeChatzigiannis: @jon-shipley all done and are now redirecting instead of calling next.

admin/authentication/middleware.js
```diff
@@ -2,18 +2,20 @@
...
+        ? [arg] : arg
       let userRole
       if (req.user) {
         userRole = ((req.user).role || {})
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley that is correct. updated.

admin/authentication/middleware.js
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley updated so they are now triggering a redirect.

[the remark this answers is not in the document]
GeorgeChatzigiannis: I prefer if it only checks for an array structure and construct an array if not since the following code checks are made on that assumption.

admin/authentication/middleware.js
GeorgeChatzigiannis: @jon-shipley the undefined check is now removed.

admin/spec/back-end/authentication/local-strategy.spec.js
GeorgeChatzigiannis: @GuyHarwood you mean ramda.includes ;)

PR #1369, opened 2019-10-08
Feature/37186 sign on

GuyHarwood:
### DfE Signin integration

- [x] refactor existing auth strategies into uniform structure
- [x] add dfe sign-in strategy, as per example
- [x] local working poc 
- [x] add issuer timeout to config
- [x] link in school org id
- [x] persist new users
- [x] issuer discovery is fault tolerant
- [x] sign out implemented

## KNOWN ISSUE
Signout results in a 404 for the user when they land back at dfe signin.
This will be covered in a new ticket once we are unblocked on this


admin/authentication/dfe-signon-strategy.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
+
+const config = require('../config')
+const { Strategy, Issuer } = require('openid-client')
+const logger = require('../services/log.service').getLogger()
+const asyncRetry = require('login.dfe.async-retry')
+
+const initSignOn = async () => {
+  Issuer.defaultHttpOptions = { timeout: 10000 }
```
jon-shipley: Are the units milliseconds?

admin/authentication/dfe-signon-strategy.js
```diff
@@ -0,0 +1,42 @@
...
+  })
+  if (config.Auth.dfeSignIn.clockTolerance && config.Auth.dfeSignIn.clockTolerance > 0) {
+    client.CLOCK_TOLERANCE = config.Auth.dfeSignIn.clockTolerance
+  }
+
+  return new Strategy({
+    client,
+    params: {
+      // redirect_uri: `http://localhost:3001/auth`,
+      scope: 'openid profile email'
```
jon-shipley: magic string?

admin/config.js
```diff
@@ -136,5 +135,18 @@ module.exports = {
     Port: process.env.REDIS_PORT || 6379,
     Key: process.env.REDIS_KEY,
     useTLS: getEnvironment() !== 'Local-Dev'
+  },
+  Auth: {
+    mode: process.env.AUTH_MODE || 'Local', // see ./lib/consts/auth-modes.js for valid options
```
jon-shipley: ✅

admin/config.js
```diff
@@ -136,5 +135,18 @@ module.exports = {
...
+    dfeSignIn: {
+      authUrl: process.env.DFE_SIGNON_AUTH_URL,
+      clientId: process.env.DFE_SIGNON_CLIENT_ID,
+      clientSecret: process.env.DFE_SIGNON_CLIENT_SECRET,
+      apiSecret: process.env.DFE_SIGNON_API_SECRET,
+      clockTolerance: process.env.DFE_SIGNON_CLOCK_TOLERANCE || 300
```
jon-shipley: Are the units milliseconds?

admin/controllers/ping.js
```diff
@@ -0,0 +1,32 @@
...
+async function getPing (req, res) {
+  // get build number from /build.txt
+  // get git commit from /commit.txt
+  let buildNumber = 'NOT FOUND'
+  let commitId = 'NOT FOUND'
+  try {
+    buildNumber = await getBuildNumber()
+  } catch (error) {
+
+  }
```
jon-shipley: Your code is fine, but I quite like:

```suggestion
  } catch (ignore) {}
```

admin/controllers/ping.js
```diff
@@ -0,0 +1,32 @@
...
+  res.setHeader('Content-Type', 'application/json')
+  let obj = {
+    'Build': buildNumber,
+    'Commit': commitId,
+    'CurrentServerTime': Date.now()
+  }
+  return res.status(200).send(obj)
+}
+
+module.exports = getPing
```
jon-shipley: ✅

admin/routes/index.js
```diff
@@ -52,45 +52,19 @@ router.get('/privacy', (req, res) => getPrivacyPage(req, res))
...
 router.get('/ping', (req, res) => getPing(req, res))
 
-/* NCA Tools Authentication Endpoint */
+/* NCA Tools or DfeSignIn Authentication Endpoint */
 router.post('/auth',
   function (req, res, next) {
-    // Only allow post requests if NCA TOOLS is enabled
-    if (!config.NCA_TOOLS_AUTH_URL) {
+    // Only allow post requests if not in local auth mode
+    if (config.Auth.mode === authModes.local) {
```
jon-shipley: The comment and the code disagree here.

admin/authentication/dfe-signon-strategy.js
GuyHarwood: taken verbatim from provided example.  will figure out what this is supposed to be and action appropriately

admin/authentication/dfe-signon-strategy.js
GuyHarwood: taken verbatim from provided example, will formalise this soon.

admin/authentication/dfe-signon-strategy.js
GuyHarwood: @jon-shipley these are the 'sets' of data you want included in the payload returned from the auth provider.  such as their email, org info etc.

This has been added as config with a default.

admin/authentication/dfe-signon-strategy.js
GuyHarwood: now a formal setting with unit postfix

admin/config.js
GuyHarwood: seconds.  config updated with postfix

admin/app.js
```diff
@@ -210,19 +211,32 @@ passport.deserializeUser(function (user, done) {
...
-  new LocalStrategy({
-    passReqToCallback: true
-  },
-  require('./authentication/local-strategy')
-  )
-)
+// initialise chosen auth strategy only
+switch (config.Auth.mode) {
+  case authModes.dfeSignIn:
+    ;(async function () {
```
jon-shipley: This IIFE is not synchronised with the switch statement at all.  Is there any way to remove the await on the `dfe-signin-strategy` and make this synchronous?

admin/controllers/authentication.js
```diff
@@ -27,65 +54,107 @@ const getSignIn = (req, res) => {
...
+  logger.info(`postSignIn: User ID logged in:
+    id:${req.user.id} \n
+    displayName:${req.user.displayName} \n
+    role:${req.user.role} \n
+    timezone:"${req.user.timezone}"`)
+
+  switch (req.user.role) {
+    case 'TEACHER':
+    case 'HEADTEACHER':
     case 'HELPDESK':
```
jon-shipley: Don't the helpdesk have their own page now?

admin/app.js
GuyHarwood: yes, we could make the initialisation of OpenIdConnect synchronous, which is a fairly valid ask given that the app should not even start if DfE signin service is not available when enabled.

admin/controllers/authentication.js
GuyHarwood: indeed they do.  I tested the helpdesk role yesterday and it worked, however I will walk through the code today to ensure its valid.  Maybe it does a redirect after establishing that no school impersonation is set...

admin/services/dfe-signin.service.js
```diff
@@ -6,23 +6,75 @@ const bluebird = require('bluebird')
...
+  getRoleInfo: async (user) => {
+    try {
+      // get role info...
+      const token = await createJwtForDfeApi()
+      const userInfo = await getUserInfoFromDfeApi(token, user)
+      logger.debug('response received from dfe.API:')
+      logger.debugObject(userInfo)
+      // TODO array check
+      const mtcRole = roleService.mapDfeRoleToMtcRole(userInfo.roles[0].code)
+      user.role = mtcRole
```
jon-shipley: This function takes an object as the param, so it is passed by reference.  I find it slightly misleading that it also returns the object which is then re-assigned by to `req.user` - which is the original object that was passed in.

admin/controllers/authentication.js
```diff
@@ -59,11 +57,11 @@ const getSignIn = (req, res) => {
 }
 
 const postDfeSignIn = async (req, res) => {
-  req.user = await dfeSignInService.process(req.user)
+  req.user = await dfeSignInService.getRoleInfo(req.user)
```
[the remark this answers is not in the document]
GuyHarwood: this is now obsolete, i will remove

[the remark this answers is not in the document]
GuyHarwood: this method is now obsolete, and no longer used.  will remove

admin/app.js
GuyHarwood: i've added a synchronous implementation, but it fails sporadically on discovery, likely due to the signin service configuring short request timeouts.

admin/controllers/authentication.js
GuyHarwood: the implementation in `master` from @GeorgeChatzigiannis recent PR always routes helpdesk to the school home page, which in turn redirects to the impersonation page, if necessary - https://github.com/DFEAGILEDEVOPS/MTC/blob/43b8288ceb7ba4c374a59140317669c60e4b8f4d/admin/controllers/school.js#L27

Seems like a valid solution to me

admin/services/dfe-signin.service.js
GuyHarwood: good catch.  i will remove the return and superfluous re-assignment.

admin/app.js
GuyHarwood: switched back to async, with promise based call in app.js

admin/app.js
```diff
@@ -210,19 +212,34 @@ passport.deserializeUser(function (user, done) {
...
-  )
-)
+if (config.Auth.mode === authModes.dfeSignIn) {
+  dfeSignInStrategy.initialiseAsync()
+    .then((strategy) => {
+      passport.use(authModes.dfeSignIn, strategy)
+    })
+    .catch((error) => {
+      logger.error(`unable to configure passport for dfeSignin:${error.message}`)
+      process.exit(1)
```
jon-shipley: Let's test this under load later on and see how it works out.  My concern is that at high-loads we could see our dependency (the Sign-in service) causing our app to fail.

I would prefer our app to handle lack of availability on the Sign-in service by doing something reasonable like showing the user a message asking them to try again later.  Getting that into the start-up phase seems like a lot more work if it's even possible at all.

This is fine for a version 1 though.

admin/app.js
```diff
@@ -210,19 +212,34 @@ passport.deserializeUser(function (user, done) {
...
+      passport.use(authModes.local,
+        new LocalStrategy({
+          passReqToCallback: true
+        },
+        require('./authentication/local-strategy')
+        )
+      )
+      break
+  }
+}
```
jon-shipley: tidy 💯

admin/package.json
```diff
@@ -22,10 +22,11 @@
     "parse-school-changes-csv": "node bin/parse-school-changes-csv.js"
   },
   "engines": {
-    "node": ">= 8.11"
+    "node": ">= 10.16.3"
```
jon-shipley: Thanks for doing this

admin/services/data-access/dfe-signin.data.service.js
```diff
@@ -0,0 +1,66 @@
...
+    logger.error(`dfe API call failed: statusCode:${response.status} - ${response.statusText}`)
+    throw new Error(`unsatisfactory response returned from DfE API. statusCode:${response.status}`)
+  }
+}
+
+/**
+ * @description creates a JWT signed with the api secret using the HS256 algorithm.
+ * @returns {Promise<string>} the signed Json Web Token
+ */
+const createJwtForDfeApi = async () => {
```
jon-shipley: Why is this method async?  It doesn't await.  The promise will still be returned from the `signAsync` method.

admin/spec/back-end/service/dfe-signin.service.spec.js
```diff
@@ -0,0 +1,247 @@
...
+    try {
+      await sut.initialiseUser({ organisation: { urn: 12345 } }, token)
+      fail('expected error to be thrown')
+    } catch (error) {
+      expect(error).toBeDefined()
+      expect(error.message).toBe('unable to find user record')
+      done()
+    }
+  })
+  // failing WIP
```
jon-shipley: Is this comment still valid?

admin/services/data-access/sql.service.js
```diff
@@ -37,7 +37,7 @@ let pool
  * @return {string | undefined}
  */
 const findDataType = (type) => Object.keys(sqlService.TYPES).find(k => {
-  logger.debug(`findDataType('${type}'): called`)
+  // logger.debug(`findDataType('${type}'): called`)
```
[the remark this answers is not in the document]
GuyHarwood: yes i would prefer to - as i mentioned on the session earlier this week, i want a new logger category for sql output, at the debug level.  These would come back into play at that point.

admin/services/data-access/dfe-signin.data.service.js
GuyHarwood: fixed

PR #1372, opened 2019-10-11
[admin-app] Add script to check for corrupt data in PS report

jon-shipley:


admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
+#!/usr/bin/env node
+'use strict'
+const fs = require('fs')
+const csv = require('fast-csv')
+const startTime = Date.now()
+const R = require('ramda')
+const RA = require('ramda-adjunct')
+
+// const timeRegex =  /^\d{1,2}:\d{2}:\d{2} (AM|PM)$/i // Not quite right - PM instead pm and single digit hours
```
GuyHarwood: delete comment if obsolete please

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+  // console.log()
+  console.log(`${prop} field not matched for ${row['AttemptId']} val was '${val}'`)
+}
+
+function reportMissing (row, prop) {
+  process.stderr.write(`${prop} field missing for ${row['AttemptId']}\n`)
+}
+
+function reportProgress (count) {
+  // process.stdout.write('.')
```
GuyHarwood: required?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+  const regex = /^\d{1,2} x \d{1,2}$/
+  regexTest(row, prop, regex)
+}
+
+const checkResponse = function (row, prop) {
+  const regex = /^\d{0,5}$/
+  regexTest(row, prop, regex)
+}
+
+const checkK = function (row, prop) {
```
GuyHarwood: mis-spelling?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+
+function checktLoad (row, prop) {
+  regexTest(row, prop, timeRegex)
+}
+
+function checktFirstKey (row, prop) {
+  regexTest(row, prop, timeRegex)
+}
+
+function checktLastKey (row, prop) {
```
GuyHarwood: checkt?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+  if (val !== '' && val !== '1' && val !== '0') {
+    report(prop, row, val)
+  }
+}
+
+function checktLoad (row, prop) {
+  regexTest(row, prop, timeRegex)
+}
+
+function checktFirstKey (row, prop) {
```
GuyHarwood: checkt?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+}
+
+function checkTimeOutSco (row, prop) {
+  const val = row[prop]
+  if (val !== '' && val !== '1' && val !== '0') {
+    report(prop, row, val)
+  }
+}
+
+function checktLoad (row, prop) {
```
GuyHarwood: checkt?

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: I'd prefer to leave it as it details the currently passing regex.  Also, when run against a dataset of 400K lines, it reports multiple times per line.  That's when you use the commented out one, as a sanity saver. 

I would suggest our next steps are to confirm with the PS team if this is acceptable and then consolidate the spec, script and output to conform to the standard.

admin/bin/check-ps-report-for-corrupt-data.js
GuyHarwood: if it's potentially useful then keep it.  Just pointing out in case you forgot about some dead code 👍

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: No :)

Naming convention giving us an odd name.  Field in the spec is called 'QnK' where n is the question number, so it becomes checkK.

Happy?

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: Field is called QntLastKey in the spec, so it becoms checktLastKey here.
Happy?

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: ditto

admin/bin/check-ps-report-for-corrupt-data.js
GuyHarwood: all good 👍

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+
+function reportMissing (row, prop) {
+  process.stderr.write(`${prop} field missing for ${row['AttemptId']}\n`)
+}
+
+function reportProgress (count) {
+  // process.stdout.write('.')
+}
+
+function regexTest (row, prop, regex) {
```
GeorgeChatzigiannis: jsdoc missing?
this method seems to be more than regex testing right?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+}
+
+function checkTimeOutResponse (row, prop) {
+  const val = row[prop]
+  if (val !== '' && val !== '1' && val !== '0') {
+    report(prop, row, val)
+  }
+}
+
+function checkTimeOutSco (row, prop) {
```
GeorgeChatzigiannis: this method name is not really readable.
is sco supposed to be something about score?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+  const val = row[prop]
+  if (!row.hasOwnProperty(prop)) {
+    return reportMissing(row, prop)
+  }
+  if (val.length !== 13) {
+    report(prop, row, val)
+  }
+}
+
+function checkForename (row, prop) {
```
GeorgeChatzigiannis: can you just reuse some of these checks for some fields? i.e. checkForename and checkSurname

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+  const val = Number(row[prop])
+  if (!row.hasOwnProperty(prop)) {
+    return reportMissing(row, prop)
+  }
+  if (val !== 3) {
+    report(prop, row, row[prop])
+  }
+}
+
+function checkDeviceType (row, prop) {
```
GeorgeChatzigiannis: same reusability comment with checkDeviceType and checkDeviceTypeModel and further down.
They implementation seems quite the same to me.
Am i missing something?

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
...
+  const regex = /^[0-9A-Za-z]{4,}$/
+  regexTest(row, prop, regex)
+}
+
+function checkTestDate (row, prop) {
+  const regex = /^\d{8}$/
+  regexTest(row, prop, regex)
+}
+
+function checkTimeStart (row, prop) {
```
GeorgeChatzigiannis: these could also be merged to one that fits the whole group of properties to be checked.

admin/bin/check-ps-report-for-corrupt-data.js
GuyHarwood: i would argue its much clearer to have a separate method that explicitly dictates what is required for the checking of each row.  otherwise you end up with a maintenance headache when you need to change one subtle thing for a particular case.  this is easy to read, and understand.

admin/bin/check-ps-report-for-corrupt-data.js
```diff
@@ -0,0 +1,380 @@
+#!/usr/bin/env node
+'use strict'
+const fs = require('fs')
+const csv = require('fast-csv')
+const startTime = Date.now()
+const R = require('ramda')
+const RA = require('ramda-adjunct')
+
+// const timeRegex =  /^\d{1,2}:\d{2}:\d{2} (AM|PM)$/i // Not quite right - PM instead pm and single digit hours
+const timeRegex = /^\d{2}:\d{2}:\d{2} (AM|PM)$/ // Correct
```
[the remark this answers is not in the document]
jon-shipley: Done

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: Done

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: The field is called `QnTimeOutSco`, so this method is consistent with the property name, which is defined in the specification.

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: Code refactored and clarified.

admin/bin/check-ps-report-for-corrupt-data.js
jon-shipley: Refactored and clarified.

PR #1382, opened 2019-10-19
feature/36677 check allocation/tslib

GuyHarwood:
Core project for all typescript functions and services

- contains all typescript function implementations and associated services
- uses jest for testing, type `yarn watch` for continual testing during development
- previous PR & branch are closed in favour of this to reduce amount of noise in one PR
- copies existing functions from `func-complete-check` into `tslib`
  - `check-marker`, `check-validator` and `check-receiver` can be ignored due to this

Once this PR is merged...
- `func-check-gen` will be deleted as no longer required
- `func-complete-check` will be deleted as no longer required
- a 'consumption function' shell project will be created to run all the core functions
  - util scripts will be created to build & copy the outputs to the function host for local development and deployment to azure
- work will continue on the implementation of check allocation


tslib/src/caching/redis-service.ts
```diff
@@ -0,0 +1,66 @@
...
+   */
+  get (key: string): Promise<any | null>
+  /**
+   *
+   * @param {string} key the unique string key of the redis entry to persist
+   * @param {object | string} value the item to persist in redis cache
+   * @param {number} ttl how long to store the item in seconds
+   * @returns {Promise<void} an awaitable promise
+   */
+  setex (key: string, value: string | object, ttl: number): Promise<void>
```
jon-shipley: Why is this called `setex` rather than `set`?

tslib/src/common/DateTimeService.ts
```diff
@@ -0,0 +1,50 @@
...
+      return moment.utc(d)
+    }
+    return d as Moment
+  }
+
+  /**
+   * Given an object will convert all Moment values to Javascript Date
+   * Useful for converting Data during sql UPDATES and INSERTS
+   */
+  convertMomentToJsDate (m: Moment): Date {
```
jon-shipley: There is a moment method to do this.  
https://momentjs.com/docs/#/displaying/as-javascript-date/

tslib/src/functions/check-marker/check-marker.v1.ts
```diff
@@ -0,0 +1,176 @@
...
+      await this.updateReceivedCheckWithMarkingError(validatedCheck, 'answers property not populated')
+      return
+    }
+
+    let parsedAnswersJson: any
+    try {
+      parsedAnswersJson = JSON.parse(validatedCheck.answers)
+    } catch (error) {
+      await this.updateReceivedCheckWithMarkingError(validatedCheck, 'answers data is not valid JSON')
+      return
```
jon-shipley: I am surprised the preceding `await` was not picked up by a linter.  This is an `async` function so we could 
`return this.updateReceivedCheckWithMarkingError(validatedCheck, 'answers data is not valid JSON')`
and save an `await`.

Am I missing something?

tslib/src/caching/redis-service.ts
GuyHarwood: it matches the underlying redis command 'set with an explicit expiration' - https://redis.io/commands/setex

tslib/src/common/DateTimeService.ts
GuyHarwood: excellent, will replace.  This was copied verbatim from the sql service, so it would be good to remove maintained code where we can

tslib/src/functions/check-marker/check-marker.v1.ts
GuyHarwood: i noticed quite a few inconsistencies like that while reviewing the PR yesterday evening.  I will go through the entire project and rectify them

tslib/src/caching/redis-service.ts
jon-shipley: Perfect! 😊

PR #1383, opened 2019-10-21
Feature/37277 add new anomaly report detection responses after the time limit

jon-shipley:


functions-app/.nvmrc
```diff
@@ -1 +1 @@
-v10.14.1
+v10.16.3
```
GuyHarwood: is there a particular reason for this?  i ask because 14.1 is recommended by MS

jon-shipley: Overzealous version upgrading 🙄 Now reverted.

functions-app/report-psychometrician/service/detections/detect-answer-responses-after-cutoff.js
```diff
@@ -0,0 +1,44 @@
...
+  const answers = R.path(['checkPayload', 'answers'], data)
+  const audits = R.pathOr([], ['checkPayload', 'audit'], data)
+
+  const output = markedAnswers.map(ma => {
+    const questionTimerStartedEvent = getQuestionTimerStartEvent(audits, ma.questionNumber, ma.factor1, ma.factor2)
+    const questionTimerTimestamp = R.prop('momentTimestamp', questionTimerStartedEvent)
+    if (!moment.isMoment(questionTimerTimestamp)) { return }
+    const cutoff = moment(questionTimerTimestamp).add(6.1, 'seconds')
+    const answer = getAnswer(answers, ma.questionNumber)
+    if (answer && moment.isMoment(answer.momentTimestamp)) {
```
GeorgeChatzigiannis: why not `if (answer && moment.isMoment(answer.momentTimestamp) && answer.momentTimestamp.isAfter(cutoff))`
Rather avoid nesting?
You could alternatively return if answer is not moment type as guard clause.

GuyHarwood: how would that be more readable?

GuyHarwood: does that aid readability?  i would say no... :/

GeorgeChatzigiannis: Then how about splitting in two statements, one that covers the guard clause `!answer || !moment.isMoment(answer.momentTimestamp)` that returns and the other one that contains the currently nested statement?

GuyHarwood: i just don't see an issue with how it currently is.  my perspective - if you understand what it does, and you can easily interpret the code, then the author has done a fair job.  

there is always more than one way to write these constructs, and if the change offers little value then i see no point in requesting it.  its not like what you are suggesting will optimise, nor aid maintainability.  Unless i'm overlooking something?

GeorgeChatzigiannis: Even though there's only 1 level of nesting in this block, I believe it's more readable and understandable if it's done with a guard clause.
I am going to approve as it has nothing to do with optimisation and it is rather my opinion since generally I am against nesting.
https://refactoring.guru/replace-nested-conditional-with-guard-clauses

jon-shipley: I've added the guard clause as I'm not opposed to it.

GuyHarwood: cool, as long as everyone is happy :)

PR #1384, opened 2019-10-22
Feature/37472 check window form validation enhancements

GeorgeChatzigiannis:
- [x]  Add active check window validation in create/edit scenarios 
- [x]  Reduced existing validation to allow current date in all user date inputs and start dates to potentially match end dates


admin/services/date.service.js
```diff
@@ -153,6 +153,20 @@ const dateService = {
...
+  /**
+   * Checks if a moment date is between a given timeframe with additional parameters
+   * @param {object} date
+   * @param {object} startDate
+   * @param {object} endDate
+   * @param {null|string} granularity
+   * @param {boolean} isInclusive
+   * @returns {boolean}
+   */
+  isBetween: function (date, startDate, endDate, granularity = null, isInclusive = false) {
```
jon-shipley: This is workable code, but passing the boolean for `isInclusive` as the 5th parameter is asking the reader to do a lot of work.

I would name the function `isBetweenInclusive()` and not have the 5th parameter.  

Alternatively, instead of passing a raw boolean primitive, make the code more readable:

```
const inclusive = true
...
if (dateService.isBetween(dt, activeCheckWindow.adminStartDate, activeCheckWindow.adminEndDate, null, inclusive)) {
```

GeorgeChatzigiannis: @jon-shipley I favoured the  `isBetweenInclusive`. Done

admin/lib/validator/check-window-v2/active-check-window-validator.js
```diff
@@ -0,0 +1,50 @@
...
+  const requestDates = R.pick([
+    'adminStartDate',
+    'adminEndDate',
+    'familiarisationCheckStartDate',
+    'familiarisationCheckEndDate',
+    'checkStartDate',
+    'checkEndDate'], requestData
+  )
+  R.forEach(dt => {
+    if (dateService.isBetweenInclusive(dt, activeCheckWindow.adminStartDate, activeCheckWindow.adminEndDate)) {
```
jon-shipley: 👍

admin/services/date.service.js
```diff
@@ -153,6 +153,18 @@ const dateService = {
 
   formatFileName: function (date) {
     return moment(date).format(filenameFriendly)
+  },
+
+  /**
+   * Checks if a moment date is inclusively between a given time frame
+   * @param {object} date
```
jon-shipley: I believe the type is `Moment.moment`

admin/services/date.service.js
```diff
@@ -157,9 +157,9 @@ const dateService = {
 
   /**
    * Checks if a moment date is inclusively between a given time frame
-   * @param {object} date
-   * @param {object} startDate
-   * @param {object} endDate
+   * @param {Moment} date
```
jon-shipley: Should be `moment.Moment`

PR #1387, opened 2019-10-23
add retry-async tests and return on retry fix

GuyHarwood:
- [x] add tests for retry-async component
- [x] fix issue where retry does not return after 1 or more failures


admin/services/data-access/retry-async.js
```diff
@@ -16,10 +16,11 @@ const defaultConfiguration = {
  */
 const asyncRetryHandler = async (asyncRetryableFunction, retryConfiguration = defaultConfiguration, retryCondition = defaultRetryCondition) => {
   const retryPolicy = {}
+  let result
   try {
     Object.assign(retryPolicy, retryConfiguration)
     logger.debug('asyncRetryHandler: executing retryable method...')
-    const result = await asyncRetryableFunction()
```
jon-shipley: I don't see the need to turn the `const` into a `let`.  It's block scoped.

GuyHarwood: i assign `result` further down, hence the `let`

jon-shipley: It's in a different block, so could also be const I think.

PR #1388, opened 2019-10-23
Feature/37498 tslib gulp

GuyHarwood:
- [x] new function host app `func-consumption` to run `tslib` transpiled functions
- [x] remove obsolete `func-check-gen` and `func-complete-check` apps
- [x] script to sync `package.json` dependencies from `tslib`
- [x] gulp tasks to copy build outputs from `tslib` to `func-consumption`
- [x] function definition for each existing function in `tslib/functions`
- [x] fix uses of asyncTableService to use storage helper
- [x] all functions can locally and be deployed
- [x] readme in func-consumption describes usage


tslib/src/functions/check-validator/check-validator.v1.ts
```diff
@@ -1,12 +1,23 @@
...
+import { ValidateCheckMessageV1, ReceivedCheck, MarkCheckMessageV1 } from '../../schemas/Models'
 import { ILogger } from '../../common/ILogger'
-import * as R from 'ramda'
 import * as RA from 'ramda-adjunct'
-import checkSchema from '../../schemas/complete-check.v1.json'
 import Moment from 'moment'
 import { ICompressionService, CompressionService } from '../../common/compression-service'
 
+const requiredSubmittedCheckProperties = [
+  'answers',
```
jon-shipley: Should be in alpha order please.

tslib/src/functions/check-validator/index.ts
```diff
@@ -1,5 +1,5 @@
 import { AzureFunction, Context } from '@azure/functions'
-import { ValidateCheckMessageV1 } from '../../schemas'
+import { ValidateCheckMessageV1 } from '../../schemas/Models'
```
jon-shipley: Capitalised directory names are so Microsoft.  I'd prefer all lowercase, but can live with it.

GuyHarwood: that's actually a file, called `Models.ts` and is inconsistent to other files in same project, will lower case them

GuyHarwood:
@sinhaamit1980 i'm getting consistent failures on this PR with the following test...
`cucumber features/check.feature:28 # Scenario: Users can complete the test using the on screen keyboard and the check data is stored in the DB`

the contents of this PR is code that doesn't run yet, and certainly does not affect the failing test functionality.

is this a known issue/ common failure?

PR #1389, opened 2019-10-24
[admin] upgrade tedious to v6

jon-shipley:


admin/data/sql/createDatabase.js
```diff
@@ -6,17 +6,27 @@ const config = require('../../config')
...
-  password: config.Sql.Migrator.Password,
   server: config.Sql.Server,
-  debug: true,
   options: {
     database: 'master',
     encrypt: config.Sql.Encrypt,
     requestTimeout: config.Sql.Migrator.Timeout,
     port: config.Sql.Port,
     connectTimeout: config.Sql.Migrator.Timeout
+  },
```
GuyHarwood: do we need any of this in the main `admin/config.js`?

jon-shipley: The `admin` app uses `mssql` as the main driver whereas this script is using `tedious` directly.  So it appears fine to me.

I am not sure why this script is using `tedious`.

File introduced in this PR: https://github.com/DFEAGILEDEVOPS/MTC/pull/333/files#diff-ec8f8dec3a18eec6cb6db2f63c2b0fad

GuyHarwood: looks like it wasn't migrated when we moved to `mssql`.  i would assume it's the only place where tedious is directly used in admin now.  Would you like me to raise a ticket to change this?

PR #1390, opened 2019-10-24
Chore/functions upgrade deps

jon-shipley:


functions/lib/sql/retry-async.js
```diff
@@ -16,7 +16,7 @@ const defaultConfiguration = {
  * @param {function(Error):Boolean} retryCondition - predicate function to determine if the function should be retried.  Defaults to true
  */
 const asyncRetryHandler = async (asyncRetryableFunction, retryConfiguration = defaultConfiguration, retryCondition = defaultRetryCondition) => {
-  let retryPolicy = {}
+  const retryPolicy = {}
```
GuyHarwood: this will need updating with the same patch as admin, will create ticket

PR #1391, opened 2019-10-25
Feature/37595 add two new anomaly report detections

jon-shipley:
Add two new anomaly report detections:

* Pupil prefs out of sequence
* Duplicates

Remove eslint packages from admin dev dependencies - they don't work for monorepos


admin/services/data-access/retry-async.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 const logger = require('../log.service').getLogger()
-const pause = (duration) => new Promise(res => setTimeout(res, duration), noReject => undefined)
+const pause = (duration) => new Promise(resolve => setTimeout(resolve, duration))
```
GuyHarwood: this is a change i made in the retry-async PR - do you need to refresh from master?

GuyHarwood: looks like it's a duplicate change, because i see a conflict with master on this file

jon-shipley: Yes, I made the same change as you.  Now resolved. 😁

admin/server.js
```diff
@@ -67,11 +67,9 @@ function onError (error) {
     case 'EACCES':
       console.error(bind + ' requires elevated privileges')
       process.exit(1)
-      break
     case 'EADDRINUSE':
       console.error(bind + ' is already in use')
       process.exit(1)
-      break
```
GeorgeChatzigiannis: Isn't the linter complaining about this being removed?

jon-shipley: No.  Checked with standard v14.3.1, and v12.0.1

PR #1393, opened 2019-10-28
Feature/37638 check window and forms unused dependencies removal

GeorgeChatzigiannis:
- [x] Removed unused v1 dependencies from check window and check form
- [x] Refactored generic check-form controller(now as test-developer)


admin/controllers/test-developer.js
```diff
@@ -0,0 +1,90 @@
...
+ * Download pupil check data ZIP file.
+ * @param req
+ * @param res
+ * @param next
+ * @returns {Promise.<void>}
+ */
+const getFileDownloadPupilCheckData = async (req, res, next) => {
+  let psychometricianReport
+  try {
+    psychometricianReport = await testDeveloperReportService.getReportMeta(req.params.urlSlug)
```
GuyHarwood: should we parse to UUID first?

....or does the service take care of that?

PR #1394, opened 2019-10-28
Bug/Map datetimeoffset to mssql datetimeoffset instead of datetime

GeorgeChatzigiannis:


admin/tests-integration/sql.service.spec.js
```diff
@@ -205,27 +205,73 @@ describe('sql.service:integration', () => {
     }
   })
 
-  xit('should store the timezone offset with the datetime value', async () => {
-    const updatedAtDate = moment('2017-12-01T15:00:00.000-08:00')
+  it('should store UTC timezone when provided with a given timezone offset as moment object', async () => {
```
GuyHarwood: can you rename these tests to something more specific about how it normalises to UTC, such as...
`should normalise a non UTC offset to UTC when stored as moment`

admin/services/data-access/sql.service.js
```diff
@@ -196,7 +196,9 @@ const sqlService = {
       confusing when checking data manually, as many GUIs (e.g. DBeaver) will then show
       the datetime with the server's TZ offset.
     */
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley ✅

admin/tests-integration/sql.service.spec.js
GeorgeChatzigiannis: done

PR #1399, opened 2019-11-05
Feature/37817 prepare checks for redis

GuyHarwood:
- [x] prepare-check service to store in redis
- [x] preserve existing default implementation
- [x] add jest for easy watch based testing
- [x] feature toggle to opt between prepared checks in table storage or redis (defaults to table storage)


admin/services/prepare-check.service.js
```diff
@@ -0,0 +1,63 @@
...
+    config: check.config,
+    createdAt: new Date(),
+    pinExpiresAt: moment(check.pupil.pinExpiresAt).toDate(),
+    pupil: R.omit(['id', 'checkFormAllocationId', 'pinExpiresAt'], check.pupil),
+    pupilId: check.pupil.id,
+    questions: check.questions,
+    school: check.school,
+    schoolId: check.school.id,
+    tokens: check.tokens,
+    updatedAt: new Date()
```
jon-shipley: Would be good to have the 'valid from' information.  We have `pinExpiresAt`, so perhaps call it `pinValidFrom`?

admin/services/redis-cache.service.js
```diff
@@ -87,4 +87,29 @@ redisCacheService.drop = async (caches = []) => {
   return true
 }
 
+/**
+ * @param {[{key:string, value:object}]} batch a dictionary of items to add to redis
+ * @param {number} ttl time to expiry in seconds, optional.
+ * @returns {Promise<void>}
+ */
+redisCacheService.addBatch = async (batch, ttl = 0) => {
```
jon-shipley: I think it's more accurate if ttl defaults to undefined or null if not set.  We might actually want to set the ttl to 0.

admin/services/prepare-check.service.js
```diff
@@ -0,0 +1,63 @@
...
+      const check = checks[index]
+      const preparedCheck = constructPreparedCheck(check)
+      const cacheKey = buildKey(check.schoolPin, check.pupilPin)
+      batch.push({
+        key: cacheKey,
+        value: preparedCheck
+      })
+      const ttl = secondsBetweenNowAnd4pm()
+      return redisService.addBatch(batch, ttl)
+    }
```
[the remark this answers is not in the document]
GuyHarwood: much cleaner :) thanks for the example code 👍

admin/services/prepare-check.service.js
```diff
@@ -0,0 +1,67 @@
...
+    return redisService.setMany(preparedChecks)
+  }
+}
+
+/**
+ *
+ * @param {Date} pinExpiry
+ * @returns {number} number of seconds between now and the pin expiry time
+ */
+function secondsBetweenNowAndPinExpiryTime (pinExpiry) {
```
jon-shipley: If pinExpiry is coming from another object it is very likely a Moment.moment object

admin/services/prepare-check.service.js
```diff
@@ -0,0 +1,67 @@
...
+    createdAt: new Date(),
+    pinExpiresAt: moment(check.pupil.pinExpiresAt).toDate(),
+    pupil: R.omit(['id', 'checkFormAllocationId', 'pinExpiresAt'], check.pupil),
+    pupilId: check.pupil.id,
+    questions: check.questions,
+    school: check.school,
+    schoolId: check.school.id,
+    tokens: check.tokens,
+    updatedAt: new Date(),
+    pinValidFrom: new Date()
```
jon-shipley: Pins are only valid from some known time in the morning. 8AM?  This time should be local time for the school's timezone.

GuyHarwood: sounds like new functionality is being introduced.
This will need to be raised with the PO

sinhaamit1980: This is what Paul was saying when we were discussing about MOD schools . He says even though we allowed to generate pin before 8 AM...but those pin should be valid from 8:00AM. As far as i know as soon as we generate pin, pupil can login from that point.

jon-shipley: fair enough

GuyHarwood: I’ll remove this in light of what we discussed earlier.  It has its own ticket now

GuyHarwood: `pinValidFrom` removed, and `updatedAt` and `createdAt` set to moments.

PR #1400, opened 2019-11-06
Feature/37819 pupil api redis

GuyHarwood:
- [x] switch from jasmine -> jest
- [x] add feature toggle for redis/table storage
- [x] look up redis when toggle enabled


pupil-api/package.json
```diff
@@ -4,11 +4,12 @@
   "private": false,
   "license": "GPL-3.0",
   "scripts": {
-    "build": "gulp scripts assets",
-    "coverage": "nyc -r text -r html ./run-jasmine.js",
+    "build": " tsc -b --v && gulp json",
```
jon-shipley: Preceeding space before command

pupil-api/src/App.ts
```diff
@@ -27,10 +29,40 @@ class App {
...
+    // tslint:disable-next-line: no-empty
+    } catch (ignore) {} // missing feature files throw intentionally
+
+    try {
+      featureTogglesDefaultPath = './feature-toggles/feature-toggles.default'
+      featureTogglesDefault = require(featureTogglesDefaultPath)
+    // tslint:disable-next-line: no-empty
+    } catch (ignore) {} // missing feature files throw intentionally
+
+    const featureTogglesMerged = R.mergeRight(featureTogglesDefault, featureTogglesSpecific)
```
jon-shipley: Nothing to do with this PR but this merging of the feature toggles seems unnecessary.   Most people add new feature toggles to every config file rather than expecting merge behaviour where more specific config options overwrite the defaults.

Perhaps it would be better just to look for the feature toggle files in the order of most specific to least specific for a particular environment?

pupil-api/src/services/redis-pupil-authentication.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor (redisService: IRedisService) {
+    if (redisService === undefined) {
+      redisService = new RedisService()
+    }
+    this.redisService = redisService
+  }
+
+  async authenticate (schoolPin: string, pupilPin: number): Promise<object> {
+    await this.redisService.get(`preparedCheck:${schoolPin}:${pupilPin}`)
```
jon-shipley: This appears to be a straightforward reimplementation of the table storage auth.

So, does the `redisService.get` method throw if the key isn't found?

What if the current time is outside of the valid period between (`pinValidFrom`, `pinExpiresAt`)?

We are hardcoding the key format here.  Would it be better to encapsulate this is a function?

pupil-api/src/services/redis.service.ts
```diff
@@ -0,0 +1,176 @@
...
+      this.logger.info('redis:end')
+    }) */
+    this.logger = new Logger()
+  }
+
+  async get (key: string): Promise<any | null> {
+    try {
+      const cacheEntry = await this.redis.get(key)
+      if (cacheEntry === null) return Promise.resolve(null)
+      const cacheItem: RedisCacheItem = JSON.parse(cacheEntry)
```
jon-shipley: Currently, the admin app is not setting the payload to be a typed RedisCacheItem.  It would be nice if it did though.

pupil-api/src/App.ts
GuyHarwood: that's the implementation taken verbatim from admin.  we should only have one file for a particular environment, and then a default, with them all in.

happy to review as a separate piece of work

pupil-api/src/services/redis-pupil-authentication.service.ts
GuyHarwood: this method is at the point of 'make the test pass after failing first time'.
should hopefully start work on the implementation today

pupil-api/src/services/redis.service.ts
GuyHarwood: yes, it would.  🤔

GuyHarwood: separate ticket created for this

pupil-api/src/services/feature.service.ts
```diff
@@ -0,0 +1,52 @@
...
+     * Load feature toggles
+     */
+    logger.info('ENVIRONMENT_NAME : ' + config.Environment)
+    const environmentName = config.Environment
+    let featureTogglesSpecific: string
+    let featureTogglesDefault: string
+    let featureTogglesSpecificPath: string
+    let featureTogglesDefaultPath: string
+    try {
+      featureTogglesSpecificPath = '../feature-toggles/feature-toggles.' + environmentName
```
GeorgeChatzigiannis: shouldn't it throw in case `config.Environment` is not set?

GuyHarwood: no, because it defaults to `Local-Dev` if not set

GeorgeChatzigiannis: ok fair enough

pupil-api/src/controllers/auth.controller.ts
```diff
@@ -13,15 +32,21 @@ class AuthController {
...
-      const data = await pupilAuthenticationService.authenticate(pupilPin, schoolPin)
+      if (this.useRedisAuth) {
+        data = await this.redisAuthService.authenticate(schoolPin, pupilPin)
+        if (data === null) {
+          return apiResponse.unauthorised(res)
+        }
+      } else {
+        data = await pupilAuthenticationService.authenticate(pupilPin, schoolPin)
+      }
+      console.log(typeof data)
```
jon-shipley: debug?

pupil-api/src/services/redis.service.ts
```diff
@@ -0,0 +1,232 @@
...
+    })
+    .on('end', () => {
+      this.logger.info('redis:end')
+    }) */
+    this.logger = new Logger()
+  }
+
+  async get (key: string): Promise<any | null> {
+    try {
+      console.log('cache key:', key)
```
jon-shipley: Could these `console.logs` be removed?

pupil-api/src/controllers/auth.controller.ts
GuyHarwood: added in the throes of despair, and forgotten in the moment of joy when the fix was realised 😄

PR #1402, opened 2019-11-08
Try it out and mtc(live) pin slip changes

GeorgeChatzigiannis:
- [x] Try it out pin slips changes
- [x] MTC check pin slips changes


admin/views/partials/_print_pupils.ejs
```diff
@@ -33,19 +41,25 @@
...
+                                <label class="pin-text">PIN:</label>
                                 <span class="pin-information"><%= pupils[i].pin %></span>
                             </div>
                             <% if (qrDataURL) { %>
                                 <div class="qr"><img src="<%= qrDataURL %>" /></div>
                             <% } %>
                             <div class="url"><%= url %></div>
                         </td>
+                         <% if (pinEnv === 'live') {%>
+                            <td class="line-checktype-1"><div>Official</div></td>
```
jon-shipley: It's at this line that I can see what you are doing 😉 I would consider changing the suffix on these classes to be something more descriptive.  Perhaps `live` and something else.  TBH, I am not sure what the difference actually is.  Perhaps the code could be clearer? If there are two classes that are so similar perhaps the common elements could be extracted?

GeorgeChatzigiannis: @jon-shipley I changed the class names to more descriptive and I extracted the common properties into a shared class.
The only difference between these two (live and try out) is the black background on live and additional width on the try out ending block.

PR #1405, opened 2019-11-09
feature/36759-check-notifier

GuyHarwood:
- [x] create check-notifier implementation in `tslib`
- [x] create function runtime definition in `func-consumption`
- [x] implement check complete notification


tslib/src/functions/check-notifier/check-notifier.v1.ts
```diff
@@ -0,0 +1,31 @@
...
+        throw new Error('not yet implemented')
+      case CheckNotificationType.checkComplete:
+        return this.checkNotifierDataService.updateCheckAsComplete(notification.checkCode)
+    }
+  }
+}
+
+export enum CheckStatus {
+  New = 1,
+  // Expired = 2, To be removed
```
jon-shipley: What happens to NEW  checks that are not collected?

tslib/src/functions/check-notifier/index.ts
```diff
@@ -0,0 +1,29 @@
...
+const functionName = 'check-notifier'
+
+const serviceBusTrigger: AzureFunction = async function (context: Context, checkNotification: ICheckNotificationMessage): Promise<void> {
+  const start = performance.now()
+  const version = checkNotification.version
+  context.log.info(`${functionName}: version:${version} message received for checkCode ${checkNotification.checkCode}`)
+  try {
+    if (version !== 1) {
+      // dead letter the message as currently we only support v1
+      throw new Error(`Message schema version:${version} unsupported`)
```
jon-shipley: If this error is not caught the message will be re-injected into the queue after a delay, again and again until the limit is reached.

If we don't want to re-process this, we should silently swallow the message by just returning.  It would be worthwhile logging a message too.

tslib/src/functions/check-notifier/check-notifier.v1.ts
GuyHarwood: they would remain NEW, never collected.  Does anything else actually need to be done?
we can discuss this in more detail later today if you wish.

tslib/src/functions/check-notifier/index.ts
GuyHarwood: good idea. will review

tslib/src/functions/check-notifier/check-notifier.v1.ts
GeorgeChatzigiannis: what's the outcome of that comment?

GuyHarwood: it's outside of this PR, but we will be addressing it soon.  It was discussed on this mornings planning session and will emerge with the pupil state analysis.

tslib/src/functions/check-notifier/index.ts
GuyHarwood: in order for it to end up on the poison queue for later inspection i have retained the error throw, but also added logging.

PR #1406, opened 2019-11-12
Chore/upgrade pupil spa to angular 8

jon-shipley:
Upgrade pupil app to angular 8


pupil-spa/src/app/services/config/config.service.ts
```diff
@@ -86,22 +86,29 @@ export function loadConfigMockService(): Function {
...
-        APP_CONFIG = Object.assign(t, envResponse);
-        resolve(true);
-      });
+    return new Promise(async (resolve, reject) => {
+      await this.http.get('/public/config.json', { observe: 'response' })
+        .pipe(first())
+        .toPromise()
+        .then(data => {
+          const t = new AppConfig();
+          console.log('data ', data)
```
GeorgeChatzigiannis: can we get rid of the log?

pupil-spa/src/polyfills.ts
```diff
@@ -11,35 +11,19 @@
...
+ * Learn more in https://angular.io/guide/browser-support
  */
 
 /***************************************************************************************************
- * BROWSER POLYFILLS
- */
+* BROWSER POLYFILLS
+*/
 
-/** IE9, IE10 and IE11 requires all of the following polyfills. **/
```
GeorgeChatzigiannis: Are these polyfill imports:
a) not used on our codebase
b) not required as angular 8 is exposing these in a different way
c)any other reason that is not obvious to me

pupil-spa/src/app/services/config/config.service.ts
jon-shipley: Yes, good spot.

pupil-spa/src/polyfills.ts
jon-shipley: Angular 8 uses differential loading - see the `browserlist` file to see what browsers we are targetting.  The polyfills will be created to support these browsers as required.

GeorgeChatzigiannis: ok good to know.

GuyHarwood:
yes, its a good point @GeorgeChatzigiannis 
will need thorough testing across the browser space

PR #1407, opened 2019-11-12
feature/37957 set expiry on prepared check after pickup

GuyHarwood:
When redis prepared checks are enabled we set the expiry to 30 minutes by default


pupil-api/src/services/redis-pupil-auth.service.ts
```diff
@@ -26,13 +27,16 @@ export class RedisPupilAuthenticationService implements IPupilAuthenticationServ
...
+    // Emit a successful login to the queue
     const pupilLoginMessage = {
       checkCode: hydratedCacheItem.checkCode,
       loginAt: new Date(),
       version: 1
     }
     azureQueueService.addMessage('pupil-login', pupilLoginMessage)
+
+    // update TTL to 30 minutes now it's been collected...
+    await this.redisService.setex(cacheKey, cacheItem, +config.RedisPreparedCheckExpiryInSeconds)
```
jon-shipley: It strikes me as odd that we need to coerce our own config into a number here.  Should we be _sure_ of that already?

GuyHarwood: historically you know why this is - we use JSON.

we could move toward an interface or class based approach where they are all pre-parsed and validated, but that would be a ticket on its own.  i would prefer that, as it's a lot neater.

jon-shipley: We don't use JSON for config the pupil-api.  Maybe you were thinking of the SPA?

In config.ts
```
  RedisPreparedCheckExpiryInSeconds: process.env.PREPARED_CHECK_EXPIRY_SECONDS || 1800
```

Sure it makes more sense to coerce it here, than where it is consumed?

E.g.

```
 RedisPreparedCheckExpiryInSeconds: parseInt(process.env.PREPARED_CHECK_EXPIRY_SECONDS, 10) || 1800
```

GuyHarwood: ah yes, thanks for the pointer 💯

PR #1408, opened 2019-11-12
feature/37916-check-started v2

GuyHarwood:
- [x] create preparedCheck cache key lookup in pupil-api for check-started function 
- [x] new check started function to invalidate preparedCheck immediately
- [x] integrate structured redis cache across admin and pupil-api, as per tslib
- [x] add check-started entry to new 'checkStarted' azure table

## Notes for QA
This PR introduces a system wide change to the way redis cache items are structured.  It is essential that your redis container is recreated or flushed prior to testing this work.


tslib/src/functions/check-started/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  const start = performance.now()
+  const version = checkStartedMessage.version
+  context.log.info(`${functionName}: version:${version} message received for checkCode ${checkStartedMessage.checkCode}`)
+  try {
+    if (version !== 1) {
+      // dead letter the message as we only support v1
+      throw new Error(`Message schema version:${version} unsupported`)
+    }
+    const checkStartedService = new CheckStartedService()
+    console.dir(context.bindings)
```
jon-shipley: Looks like debug... remove?

admin/services/redis-cache.service.js
```diff
@@ -32,7 +32,7 @@ redisCacheService.get = async key => {
   redisConnect()
   try {
     const cacheEntry = await redis.get(key)
-    if (cacheEntry === null) return Promise.resolve(null)
+    if (cacheEntry === null) return null
```
jon-shipley: Why prefer null over undefined?  Generally, I would prefer to not use null in services.   And on that subject, shouldn't this service be in the data-access layer?

GuyHarwood: this is a straight copy from the `tslib` version. will switch to undefined.

regards placement of the service - i agree, its a generic data service.  will be interesting to see how vscode handles the move 😄

GuyHarwood: hang on a minute - i think there was a reason i used null, because `ioredis` returns null.  i can obviously return undefined, but the null check was necessary due to what the library returned.

will debug it.

GuyHarwood: now returning `undefined` across all implementations.

pupil-spa/src/app/services/check-start/check-start.service.spec.ts
```diff
@@ -47,12 +47,19 @@ describe('CheckStartService', () => {
...
     const addEntrySpy = spyOn(auditService, 'addEntry');
     spyOn(tokenService, 'getToken').and.returnValue({ url: 'url', token: 'token'});
-    spyOn(azureQueueService, 'addMessage')
-      .and.returnValue(Promise.resolve());
+    // spyOn(azureQueueService, 'addMessage').and.returnValue(Promise.resolve());
+    let actualPayload;
+    spyOn(azureQueueService, 'addMessage').and
+    .callFake((queueName, url, token, payload, retryConfig) => {
+      actualPayload = payload;
+      return Promise.resolve();
```
GeorgeChatzigiannis: simple return should work too I believe.
Also you might want to remove the comment above.

GeorgeChatzigiannis: simple return should work too I believe.
Also you might want to remove the comment above.

GuyHarwood: updated, thanks

PR #1410, opened 2019-11-13
automation/update-automation-for-new-check-process

sinhaamit1980:


test/admin-hpa/features/support/azure_table_helper.rb
```diff
@@ -16,4 +16,25 @@ def self.wait_for_prepared_check(school_password, pin)
...
 
+  #### -- Enable below method when new check process is enabled
+  # def self.wait_for_prepared_check(school_password, pin)
+  #   p school_password, pin
+  #   found = false
+  #   begin
+  #     retries ||= 0
+  #     sleep 2
+  #     REDIS_CLIENT.keys.each do |key|
+  #       if key.include?("#{pin}")
```
GuyHarwood: this doesn't seem very precise given that a single 4 digit pin can be used across more than one school.
why not construct the exact prepared check key and look it up with `get`
also, enumerating all the redis keys is not optimal

sinhaamit1980: Updated, please review again

PR #1412, opened 2019-11-15
Feature/37980 remove status from pupil register

GeorgeChatzigiannis:
Added v2 toggle, services and view for pupil register without status


admin/services/data-access/pupil-register-v2.data.service.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+const { TYPES } = require('./sql.service')
+
+const sqlService = require('./sql.service')
+
+const service = {
+  getPupilRegister: async function (schoolId) {
```
GuyHarwood: please add JSDoc comments

GuyHarwood: also - you should be able to remove the async and just document the return value as a promise, as you don't await anything

GeorgeChatzigiannis: thanks for that, added.

admin/services/data-access/pupil-register-v2.data.service.js
```diff
@@ -0,0 +1,38 @@
...
+         p.urlSlug,
+         p.dateOfBirth,
+         p.school_id,
+         p.group_id,
+         ISNULL(g.name, '-') as groupName
+      FROM
+           [mtc_admin].[pupil] p LEFT JOIN
+           [mtc_admin].[group] g ON (p.group_id = g.id)
+      WHERE p.school_id = @schoolId
+      ORDER BY p.lastName, p.foreName
```
jon-shipley: In my view, we should do the ordering on the admin app servers and not the database.  There is one DB server; the app layer is horizontally scalable.

admin/services/pupil-register-v2.service.js
```diff
@@ -0,0 +1,29 @@
...
+      return {
+        urlSlug: d.urlSlug,
+        foreName: d.foreName,
+        lastName: d.lastName,
+        middleNames: d.middleNames,
+        dateOfBirth: d.dateOfBirth,
+        group: d.groupName
+      }
+    })
+    pupilIdentificationFlagService.addIdentificationFlags(pupilRegister)
```
jon-shipley: Adding identification flags requires an ordered array.  As this is a pre-condition anyway, why not add the ordering here and drop it from the database?

admin/views/pupil-register/pupils-list-v2.ejs
```diff
@@ -0,0 +1,151 @@
...
+            <% if (incompletePupils && incompletePupils.length > 0) { %>
+                <div class="govuk-warning-message govuk-!-margin-bottom-7">
+                    <%= incompletePupils.length %> <%= incompletePupils.length > 1 ? 'pupils have' : 'pupil has'  %> an Incomplete status
+                </div>
+            <% } %>
+            <%- partial('partials/_readonly', {availabilityData}) %>
+            <h1 class="govuk-heading-xl">Pupil register</h1>
+            <%- partial('partials/_gds_session_expiry') %>
+            <p class="govuk-body">
+                This list is taken from your school's census. Review your list and add pupils, if required. If a pupil is not going to participate, go to <a href="/pupils-not-taking-the-check">Pupils not taking the check</a>.
```
jon-shipley: Please wrap long lines.

admin/views/pupil-register/pupils-list-v2.ejs
```diff
@@ -0,0 +1,151 @@
...
+                        <li>
+                            <a class="govuk-link" href="/access-arrangements/overview" rel="noopener noreferrer">Access arrangements</a>
+                        </li>
+                    </ul>
+                </nav>
+            </aside>
+        </div>
+    </div>
+</main>
+<script>
```
jon-shipley: We should be removing all inline script tags for security reasons.

GeorgeChatzigiannis: @jon-shipley this should be a separate piece of work that tackles the removal of all the script tags we currently have in our views.
Function calls are related to sorting, autocompletion, check forms, printing and access arrangements.

GuyHarwood: but if you are just creating this page it would be best to do it here, rather than creating extra debt?

admin/services/data-access/pupil-register-v2.data.service.js
GeorgeChatzigiannis: @jon-shipley removed and added admin backend sorting

admin/services/pupil-register-v2.service.js
```diff
@@ -0,0 +1,30 @@
...
+      return {
+        urlSlug: d.urlSlug,
+        foreName: d.foreName,
+        lastName: d.lastName,
+        middleNames: d.middleNames,
+        dateOfBirth: d.dateOfBirth,
+        group: d.groupName
+      }
+    })
+    pupilIdentificationFlagService.addIdentificationFlags(tableSorting.applySorting(pupilRegister, 'lastName'))
```
jon-shipley: Shouldn't this sort by lastName, firstname, dob ?

PR #1414, opened 2019-11-18
Prioritise aliases on pupil spa welcome screen

GeorgeChatzigiannis:
Added new fields on prepare check config, relevant views and in display with priority on pupil spa

### DIFF
--- a/admin/views/partials/_print_pupils.ejs
+++ b/admin/views/partials/_print_pupils.ejs
@@ -26,10 +26,10 @@
                             <% } %>
                             <div class="pupil-content">
                             <span class="fullname"><%=
-                                    `${pupils[i].foreName} ${pupils[i].showMiddleNames
+                                    `${pupils[i].foreNameAlias || pupils[i].foreName} ${pupils[i].showMiddleNames
                                             ? pupils[i].middleNames
                                             : ''
-                                    } ${pupils[i].lastName}` %>
+                                    } ${pupils[i].lastNameAlias || pupils[i].lastName}` %>
                                 <% if (pupils[i].group) { %>
                                     (<%= pupils[i].group %>)
                                 <% } %>
--- a/admin/views/partials/_print_pupils.ejs
+++ b/admin/views/partials/_print_pupils.ejs
@@ -26,10 +26,10 @@
                             <% } %>
                             <div class="pupil-content">
                             <span class="fullname"><%=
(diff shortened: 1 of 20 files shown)


admin/views/partials/_print_pupils.ejs
```diff
@@ -26,10 +26,10 @@
                             <% } %>
                             <div class="pupil-content">
                             <span class="fullname"><%=
-                                    `${pupils[i].foreName} ${pupils[i].showMiddleNames
+                                    `${pupils[i].foreNameAlias || pupils[i].foreName} ${pupils[i].showMiddleNames
```
jon-shipley: I would prefer to have a simple field printed to screen here, populated by the presentation layer.

We only have two views, a pupil view of the name, and the staff view.

pupil-spa/src/app/login-success/login-success.component.html
```diff
@@ -4,17 +4,17 @@
         <div class="grid-row">
             <div class="column-full">
                 <header class="page-header">
-                    <h1 class="aa-title-size" tabindex="0">Welcome, {{ pupil.firstName }}</h1>
+                    <h1 class="aa-title-size" tabindex="0">Welcome, {{ pupil.firstNameAlias || pupil.firstName }}</h1>
```
jon-shipley: Why store two properties in the SPA?  Why not just call it `firstName`?  It is always the pupil view of their name.

GuyHarwood: as long as this is implemented in the payload generation code, and not the underlying view, i agree

GuyHarwood: as long as this is implemented in the payload generation code, and not the underlying view, i agree

admin/views/partials/_print_pupils.ejs
GeorgeChatzigiannis: There is no requirement for the staff view to display anything else rather than the actual names.
I added a pupil name presenter to capture the pupil view level.

pupil-spa/src/app/login-success/login-success.component.html
GeorgeChatzigiannis: This is now reverted to display `firstName` and `lastName` since it's processed now prior to message queue submission.

PR #1415, opened 2019-11-18
Bug/38070 tio checks expiring

GuyHarwood:
- [x] fix check-started function to only drop preparedCheck if `practice: false`
- [x] fix pupil-api to only set expiry if `practice: false`

### DIFF
--- a/pupil-api/src/services/redis-pupil-auth.service.ts
+++ b/pupil-api/src/services/redis-pupil-auth.service.ts
@@ -37,7 +37,9 @@ export class RedisPupilAuthenticationService implements IPupilAuthenticationServ
     azureQueueService.addMessage('pupil-login', pupilLoginMessage)
     const checkStartedLookupKey = this.buildCheckStartedLookupKey(preparedCheckEntry.checkCode)
     await this.redisService.setex(checkStartedLookupKey, cacheKey, this.eightHoursInSeconds)
-    await this.redisService.expire(cacheKey, config.RedisPreparedCheckExpiryInSeconds)
+    if (preparedCheckEntry.config.practice === false) {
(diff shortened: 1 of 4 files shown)


pupil-api/src/services/redis-pupil-auth.service.ts
```diff
@@ -37,7 +37,9 @@ export class RedisPupilAuthenticationService implements IPupilAuthenticationServ
     azureQueueService.addMessage('pupil-login', pupilLoginMessage)
     const checkStartedLookupKey = this.buildCheckStartedLookupKey(preparedCheckEntry.checkCode)
     await this.redisService.setex(checkStartedLookupKey, cacheKey, this.eightHoursInSeconds)
-    await this.redisService.expire(cacheKey, config.RedisPreparedCheckExpiryInSeconds)
+    if (preparedCheckEntry.config.practice === false) {
```
jon-shipley: This reads like a double negative to me.  I realise it is existing code though.

GeorgeChatzigiannis: Regarding Jon's comment for this, is it worth checking if the `config` has a property `practice` using something like `hasOwnProperty` or if it is `undefined`?

GuyHarwood: that's a good point George, because typescript will suggest that if its a non-optional boolean, there is no point checking for undefined.  However in this case, we know it could be undefined at runtime, causing it to go into live mode, just by not being present, which is not what we want.  will add some structure to this. thanks

GuyHarwood: i've added unit tests for the above case.  the existing line is still valid, even when `config.practice` is undefined or non existent

PR #1416, opened 2019-11-19
Feature/38093 feature toggle files -> config

GuyHarwood:
## admin app
- [x] create matching default toggles under config node
- [x] load from config

## pupil api
- [x] create matching default toggles under config node
- [x] load from config


pupil-api/src/services/feature.service.ts
```diff
@@ -24,29 +23,7 @@ export class FeatureService implements IFeatureService {
...
-    // tslint:disable-next-line: no-empty
-    } catch (ignore) {} // missing feature files throw intentionally
-
-    const featureTogglesMerged = R.mergeRight(featureTogglesDefault, featureTogglesSpecific)
-
-    if (featureTogglesMerged) {
-      logger.info(`Loading merged feature toggles from '${featureTogglesSpecificPath}', '${featureTogglesDefaultPath}': `, featureTogglesMerged)
-      featureToggles.load(featureTogglesMerged)
-    }
+    featureToggles.load(config.FeatureToggles)
```
jon-shipley: Can we have a line from `logger.info` saying what feature toggles are loaded please?  

E.g. 
```
logger.info(`Loading feature toggles from: `, config.FeatureToggles)
```

GuyHarwood: good idea. added to both.

PR #1420, opened 2019-11-21
Feature/37977 Pupil filter

GeorgeChatzigiannis:
- [x] Added reusable pupil filter based on name and upn on pupil register, generate pins and view print pins pages
- [x] Altered db views to provide upns
- [x] Added test spec for existing pupil filter


admin/assets/javascripts/pupil-filter-name.js
```diff
@@ -1,43 +1,57 @@
...
 /* global $ stickyBanner */
-$(function () {
-  if ($('.filter-name').length > 0) {
-    $('#search-name').on('change keyup', function () {
-      var input = $.trim($(this).val()).toLowerCase()
-      var selAllTr = '#generatePins > tbody > tr'
-      if (input.length === 0) {
+(function () {
+  window.MTCAdmin.pupilFilter = function () {
+    if ($('.filter-name').length > 0) {
```
GuyHarwood: how does this perform on larger sets on the first couple of inputs?  might be worth checking and only invoking after 2nd or 3rd character.

GeorgeChatzigiannis: @GuyHarwood that's existing logic from Dan's filter. I haven't notice any performance impact so far.

admin/views/pupil-pin/generate-pins-list.ejs
```diff
@@ -40,14 +40,15 @@
                     <% pupils.forEach( (pupil, index) => { %>
                         <tr class="govuk-table__row <% if (pupil && pupil.group_id) { %>group-id-<%= pupil.group_id %><% } %>">
                             <td scope="row">
-                                <label for="pupil-<%= pupil.id %>">
-                                    <%= pupil.fullName %>
+                                <label id="pupilName" for="pupil-<%= pupil.id %>">
+                                    </span><%= pupil.fullName %>
```
jon-shipley: This closing SPAN tag looks unmatched.

GeorgeChatzigiannis: nice catch! 🦅 👀 
span removed

PR #1421, opened 2019-11-25
Bug/38215 Group name missing from pin slips

GeorgeChatzigiannis:
Call presenter after group name has been retrieved for pupils


GeorgeChatzigiannis:
@GuyHarwood that's a good point.
Existing tests did not cover unfortunately, I made a push to improve them.

PR #1422, opened 2019-11-26
upgrade sql linux to 2019 edition

GuyHarwood:


GuyHarwood:
@activemq @sinhaamit1980 FYI - this will require a full local test suite run

PR #1423, opened 2019-11-26
Feature/38148 check look up redis key

GuyHarwood:
Creates a secondary lookup for prepared checks based on checkCode, rather than all services requiring the pupil & school pin.

This was originally conceived in the pupil-api, and has now been moved to the prepare check service, at the point when the checks are created.

### DIFF
--- a/admin/services/prepare-check.service.js
+++ b/admin/services/prepare-check.service.js
@@ -35,7 +35,7 @@ const service = {
 }
 
 function buildLookupKey (checkCode) {
-  return `check-started-check-lookup:${checkCode}`
+  return `prepared-check-lookup:${checkCode}`
 }
 
 /**
--- a/admin/services/prepare-check.service.js
+++ b/admin/services/prepare-check.service.js
@@ -13,16 +13,29 @@ const service = {
     if (!Array.isArray(checks)) {
       throw new Error('checks is not an array')
     }
-    const preparedChecks = checks.map(check => {
+    const lookupKeys = []
+    const cacheItems = checks.map(check => {
       const preparedCheck = constructPreparedCheck(check)
+      const preparedCheckKey = buildKey(check.schoolPin, check.pupilPin)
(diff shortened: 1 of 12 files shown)


admin/spec/back-end/service/prepare-check.service.spec.js
```diff
@@ -50,15 +51,20 @@ describe('prepare-check.service', () => {
...
-    let actualKey
+  it('should add each check with the expected cache key and a lookup item', async () => {
+    let actualPreparedCheckKey, actualLookupKey
     check.schoolPin = 'schoolPin'
     check.pupilPin = 'pupilPin'
+    const expectedLookupKey = `check-started-check-lookup:${check.checkCode}`
+    const expectedPreparedCheckKey = `preparedCheck:${check.schoolPin}:${check.pupilPin}`
     spyOn(redisService, 'setMany').and.callFake((batch, ttl) => {
-      actualKey = batch[0].key
+      console.dir(batch)
```
GeorgeChatzigiannis: do you need that extra noise in your test?

admin/services/prepare-check.service.js
```diff
@@ -13,18 +13,31 @@ const service = {
...
       return {
-        key: buildKey(check.schoolPin, check.pupilPin),
+        key: preparedCheckKey,
         value: preparedCheck,
-        ttl: secondsBetweenNowAndPinExpiryTime(preparedCheck.pinExpiresAt)
+        ttl: ttl
       }
     })
-    return redisService.setMany(preparedChecks)
+    cacheItems.push(...lookupKeys)
```
GeorgeChatzigiannis: nice way of building all the items that needs to be set on redis

admin/spec/back-end/service/prepare-check.service.spec.js
jon-shipley: Please delete the console.dir

PR #1424, opened 2019-11-27
Feature/38206 pupil register cache

GeorgeChatzigiannis:
- Introduced service responsible for setting the pupil register in cache for a school
- Drop cache on pupil add(single-multiple)/edit, group add/edit/remove


admin/services/pupil-register-caching.service.js
```diff
@@ -0,0 +1,41 @@
...
+   * @return {Array}
+   */
+  setPupilRegisterCache: async function (schoolId, pupilRegisterViewData) {
+    if (!schoolId) {
+      throw new Error('School id not found in session')
+    }
+    if (!pupilRegisterViewData || pupilRegisterViewData.length === 0) {
+      throw new Error('Pupil register view data not provided')
+    }
+    const pupilRegisterRedisKey = `school:${schoolId}`
```
GeorgeChatzigiannis: Need thoughts on redis key name.
I am guessing it's not very descriptive 🤔

admin/services/pupil-register-caching.service.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+
+const featureToggles = require('feature-toggles')
+
+const pupilRegisterV2Service = require('./pupil-register-v2.service')
+const pupilRegisterService = require('./pupil-register.service')
+const redisCacheService = require('./data-access/redis-cache.service')
+
+const pupilRegisterCachingService = {
```
GuyHarwood: i don't believe we need a dedicated pupil register cache, all the logic should be encapsulated in the pupil register service.

admin/services/pupil-register-v2.service.js
```diff
@@ -11,6 +13,23 @@ const pupilRegisterV2Service = {
    * @return {Promise<*>}
    */
   getPupilRegister: async function (schoolId) {
+    if (!schoolId) {
+      throw new Error('School id not found in session')
+    }
+    const pupilRegisterRedisKey = `school:${schoolId}`
```
GuyHarwood: please add more specific key

admin/services/pupil-register-v2.service.js
```diff
@@ -22,7 +41,10 @@ const pupilRegisterV2Service = {
         group: d.groupName
       }
     })
-    return pupilIdentificationFlagService.addIdentificationFlags(tableSorting.applySorting(pupilRegister, 'lastName'))
+    const pupilRegisterViewData = pupilIdentificationFlagService.addIdentificationFlags(tableSorting.applySorting(pupilRegister, 'lastName'))
+    const pupilRegisterRedisKey = `school:${schoolId}`
+    await redisCacheService.set(pupilRegisterRedisKey, pupilRegisterViewData)
```
GuyHarwood: you should check if the data exists in the cache, and return it.  if it doesnt exist, get from sql, add to cache and return

admin/controllers/pupil.js
```diff
@@ -62,6 +63,7 @@ const postAddPupil = async (req, res, next) => {
   res.locals.pageTitle = 'Add pupil'
   try {
     const pupil = await pupilAddService.addPupil(req.body, req.user.schoolId)
+    await pupilRegisterCachingService.refreshPupilRegisterCache(req.user.schoolId)
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @GuyHarwood this method and service is removed and replaced with redisservice.drop
When making a request to get the list of pupils in pupil register view the cache will be recreated.

admin/services/pupil-register-caching.service.js
GeorgeChatzigiannis: @GuyHarwood yes, I removed it.

admin/services/pupil-register-v2.service.js
GeorgeChatzigiannis: @GuyHarwood This is done by the `getPupilRegister` which calls `getPupilRegisterViewData` once it identifies that no data exist in the cache.
If `getPupilRegister` detects data in cache `getPupilRegisterViewData` won't be called at all.

admin/services/pupil-register-v2.service.js
```diff
@@ -11,6 +13,24 @@ const pupilRegisterV2Service = {
    * @return {Promise<*>}
    */
   getPupilRegister: async function (schoolId) {
+    if (!schoolId) {
+      throw new Error('School id not found in session')
+    }
+    const pupilRegisterRedisKey = `pupilRegisterViewData:${schoolId}`
```
jon-shipley: I am slightly concerned that there isn't a buildKey() method or something.  If you GET the key here; how is it set? Duplicate code?

GuyHarwood: seems like a good opportunity to implement the `redisKeyService` we talked about.

`getPupilRegisterViewDataKey(schoolid: number): string`

admin/controllers/pupil.js
```diff
@@ -62,6 +61,8 @@ const postAddPupil = async (req, res, next) => {
   res.locals.pageTitle = 'Add pupil'
   try {
     const pupil = await pupilAddService.addPupil(req.body, req.user.schoolId)
+    const pupilRegisterRedisKey = `pupilRegisterViewData:${req.user.schoolId}`
+    await redisCacheService.drop(pupilRegisterRedisKey)
```
GuyHarwood: we should be invalidating cache in the service, not the controller

admin/controllers/pupil.js
```diff
@@ -160,6 +161,8 @@ const postAddMultiplePupils = async (req, res, next) => {
     res.fileErrors = uploadResult.fileErrors
     return getAddMultiplePupils(req, res, next)
   } else {
+    const pupilRegisterRedisKey = `pupilRegisterViewData:${req.user.schoolId}`
+    await redisCacheService.drop(pupilRegisterRedisKey)
```
GuyHarwood: we should be invalidating cache in the service, not the controller

admin/services/pupil-register-v2.service.js
GeorgeChatzigiannis: added

admin/controllers/pupil.js
GeorgeChatzigiannis: @GuyHarwood updated

admin/services/group.service.js
```diff
@@ -90,6 +93,8 @@ groupService.update = async (id, group, schoolId) => {
     // only update pupils if list has changed
     await groupDataService.sqlModifyGroupMembers(id, pupils)
   }
+  const pupilRegisterRedisKey = redisKeyService.getPupilRegisterViewDataKey(schoolId)
+  await redisCacheService.drop(pupilRegisterRedisKey)
```
jon-shipley: This reads very well ⭐️

PR #1428, opened 2019-11-28
Feature/37958 admin app changes for pupil state

jon-shipley:
Pupil state : part 1

Update the admin app to set the new pupil state fields: currentCheckId, checkComplete, restartAvailable, attendanceId

Another PR will update the views to use the new fields.

ToDo
 - [x] remove restart
 - [x] add restart
 - [x] add attendance
 - [x] remove attendance


admin/services/check-window-v2.service.js
```diff
@@ -25,8 +25,17 @@ checkWindowV2Service.getCheckWindow = async (urlSlug) => {
...
-checkWindowV2Service.getActiveCheckWindow = async () => {
-  return checkWindowDataService.sqlFindActiveCheckWindow()
+let cachedActiveCheckWindow
+let cachedActiveCheckWindowExpiresAt
+checkWindowV2Service.getActiveCheckWindow = async (cacheBust = false) => {
+  const now = (new Date()).getTime()
+
+  if (cacheBust || !cachedActiveCheckWindow || !cachedActiveCheckWindowExpiresAt || now > cachedActiveCheckWindowExpiresAt) {
+    cachedActiveCheckWindow = await checkWindowDataService.sqlFindActiveCheckWindow()
+    cachedActiveCheckWindowExpiresAt = (new Date()).getTime() + 60 * 1000 // +60 seconds
```
GuyHarwood: why not use redis TTL and pass it to `redisService.set` or am i misunderstanding this?

admin/services/restart.service.js
```diff
@@ -200,12 +206,16 @@ restartService.markDeleted = async (pupilUrlSlug, userId, schoolId) => {
...
+  }
+
   // see if there is a check associated with this restart, ideally the slug
   // would refer to the restart itself, and not the pupil.
   if (restart.check_id) {
     const check = await pupilRestartDataService.sqlFindCheckById(restart.check_id, schoolId)
-    await checkStateService.changeState(check.checkCode, checkStateService.States.Expired)
-    await azureQueueService.addMessageAsync('prepared-check-delete', {
+    // await checkStateService.changeState(check.checkCode, checkStateService.States.Expired)
+    azureQueueService.addMessage('prepared-check-delete', {
```
GuyHarwood: you will need to use the redis prepared check feature toggle here and drop from redis if true, otherwise send queue message

admin/services/check-window-v2.service.js
jon-shipley: The issue, for me at least, was discovering that an express middleware function is called on every request which makes a network call to get the current active check window.  This modification stores it in memory for 60 seconds reducing network traffic.  It requires a trivial amount of memory, and the data is very slow-moving.  

After 60 seconds it will refresh itself from Redis (unfortunately the Redis cache implementation is at the data service level, so it is not clear.)

If the Redis cache is missing it will read from the Database.

admin/controllers/restart.js
```diff
@@ -145,15 +144,6 @@ controller.postSubmitRestartList = async (req, res, next) => {
   const pupilUrlSlugs = pupilsRestarted && pupilsRestarted.map(p => encodeURIComponent(p.urlSlug))
   const pupilsToHighlight = pupilUrlSlugs.join()
   req.flash('info', restartInfo)
-
```
jon-shipley: These status recalculations have moved to the service layer and will be deleted once the new pupil state work is complete.

admin/data/sql/migrations/20191126095851.do.add-check-status-calc-func.sql
```diff
@@ -0,0 +1,117 @@
...
+            @startedAt = startedAt
+    FROM
+        [mtc_admin].[check] c LEFT OUTER JOIN
+        [mtc_admin].[checkPin] cp ON (c.id = cp.check_id)
+    WHERE
+            c.id = @checkId;
+
+    -- NEW
+    IF (@pupilLoginDate IS NULL AND
+        @startedAt IS NULL AND
```
jon-shipley: @GuyHarwood - given the imminancy of the new world so we want to remove these startedAt vars?

admin/services/check-start.service/check-start.service.js
```diff
@@ -146,7 +167,7 @@ checkStartService.prepareCheck2 = async function (
   if (prepareCheckServiceEnabled) {
     prepareCheckService.prepareChecks(pupilChecks)
```
jon-shipley: This is ultimately an async operation...

admin/tests-integration/udf-calc-check-status-id.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    it('does not class a try it out check as not received', async () => {
+      const id = await createCheck('NTR1', 0)
+      const res = await sql.query(createQuery(id))
+      const checkStatusId = res[0].code
+      expect(checkStatusId).toBe('STD') // Started
+    })
+  })
+
+  describe('unusual conditions:', () => {
+    it('does not class a try it out check as complete', async () => {
```
jon-shipley: @GeorgeChatzigiannis @GuyHarwood - can you think of anything to add here?

admin/data/sql/migrations/20191126095851.do.add-check-status-calc-func.sql
GuyHarwood: yes, makes sense to me

admin/services/check-state.service.js
```diff
@@ -19,11 +19,13 @@ module.exports.States = Object.freeze({
...
  */
 const checkStatusCodes = Object.freeze({
-  New: { code: 'NEW' },
-  Expired: { code: 'EXP' },
+  Collected: { code: 'COL' },
   Complete: { code: 'CMP' },
+  Expired: { code: 'EXP' },
+  New: { code: 'NEW' },
+  NotReceived: { code: 'NTR' },
   Started: { code: 'STD' },
```
GuyHarwood: should we get rid of service? or at the very least identify code that may still be using it?

admin/data/sql/migrations/20191126095851.do.add-check-status-calc-func.sql
GuyHarwood: also the enum entry, which ive commented on separately

admin/services/check-state.service.js
jon-shipley: Yes good idea.   Once we have moved over to the new world we can remove any dead code left behind.

admin/data/sql/migrations/20191126095851.do.add-check-status-calc-func.sql
jon-shipley: Okay- let's queue these up for the next PR as otherwise, we will break the build with this PR.  Is that okay with you @GuyHarwood?

The next PR will move the views and other consumers of the pupil state over to the new pupil state fields so we can keep the build green.

GuyHarwood: yes, fine with me.

admin/services/check-start.service/check-start.service.js
GuyHarwood: as discussed, please add the await in this PR

admin/data/sql/migrations/20191126095851.do.add-check-status-calc-func.sql
```diff
@@ -0,0 +1,117 @@
...
+        @pupilLoginDate IS NOT NULL AND
+        DATEADD(minute, @checkTimeLimit, @pupilLoginDate) < GETUTCDATE())
+        BEGIN
+            SELECT @ret = id FROM [mtc_admin].[checkStatus] where code = 'NTR';
+            RETURN @ret;
+        END
+
+    -- If none of the IF conditions were met this will return the original
+    -- check status id.
+    RETURN @ret;
```
GeorgeChatzigiannis: shouldn't an error be thrown if there is no match?

admin/tests-integration/udf-calc-check-status-id.spec.js
GeorgeChatzigiannis: you could say not received for try out is also unusual state even though you have added the test in the block above?

admin/data/sql/migrations/20191126095851.do.add-check-status-calc-func.sql
```diff
@@ -0,0 +1,117 @@
...
+    IF (@isLiveCheck = 1 AND
+        @pupilLoginDate IS NOT NULL AND
+        DATEADD(minute, @checkTimeLimit, @pupilLoginDate) < GETUTCDATE())
+        BEGIN
+            SELECT @ret = id FROM [mtc_admin].[checkStatus] where code = 'NTR';
+            RETURN @ret;
+        END
+
+    -- If none of the IF conditions were met this will return the original
+    -- check status id.
```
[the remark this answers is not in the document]
jon-shipley: Possibly.  @GuyHarwood what do you think?  It would be clearer that the check was in an odd state?  It wouldn't be able to do anything about it though.

admin/tests-integration/udf-calc-check-status-id.spec.js
jon-shipley: I think that's standard - all try it out checks get expired at the end of the day.  Negative tests are useful too?

GeorgeChatzigiannis: ok then the test is in the right block.

PR #1434, opened 2019-12-04
Bug/38324 admin logon event

GuyHarwood:
as per existing auth provider implementations, dfe sign-in module should add entry to adminLogonEvent table

**Known issues**
No access to request context without extensive work, so remote IP, user agent, request body and session id are unavailable


GuyHarwood:
@sinhaamit1980 @activemq i will show you how to setup local env for testing this

PR #1439, opened 2019-12-05
spike/38422 read replica for queries

GuyHarwood:
- [x] separate connection pool for readonly replica.
- [x] new `sqlService.readonlyQuery` method for distinction
- [x] refactor a number of key data-access services to use readonlyQuery for testing against azure
- [x] implement kill switch that diverts back to a single connection


admin/config/sql.readonly.config.js
```diff
@@ -0,0 +1,29 @@
...
+  password: config.Sql.Application.Password,
+  pool: {
+    min: process.env.SQL_POOL_MIN_COUNT_REPLICA || 0,
+    max: process.env.SQL_POOL_MAX_COUNT_REPLICA || 5,
+    acquireTimeoutMillis: parseInt(process.env.SQL_POOL_ACQUIRE_TIMEOUT_REPLICA, 10) || thirtySecondsInMilliseconds
+  },
+  options: {
+    appName: `${config.Sql.Application.name}:Readonly`,
+    encrypt: config.Sql.Encrypt,
+    readOnlyIntent: {}.hasOwnProperty.call(process.env, 'SQL_READONLY_INTENT') ? toBool(process.env.SQL_READONLY_INTENT) : true
```
jon-shipley: What's this?

admin/services/data-access/sql.service.js
```diff
@@ -288,6 +318,44 @@ sqlService.query = async (sql, params = [], redisKey) => {
...
 
+/**
+ * Query data from SQL Server over a readonly connection
+ * @param {string} sql - The SELECT statement to execute
+ * @param {array} params - Array of parameters for SQL statement
+ * @param {string} redisKey - Redis key to cache resultset against
+ * @returns {Promise<*>}
+ */
+sqlService.readonlyQuery = async (sql, params = [], redisKey) => {
+  // logger.debug(`sql.service.readonlyQuery(): ${sql}`)
```
GeorgeChatzigiannis: I guess these debug logs need to be removed

admin/config/sql.readonly.config.js
GuyHarwood: Sql azure allows you to specify your intention of this being a connection for reads against a replica

GuyHarwood: https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/configure-read-only-routing-for-an-availability-group-sql-server

admin/services/data-access/sql.service.js
GuyHarwood: Yeah, they were preserved in case we wanted them at some point,  but a sql specific category would be cleaner

admin/config/sql.readonly.config.js
jon-shipley: Yes, but I'm confused as to why it is configurable?  Is it to support development?

GuyHarwood: initially, yes.  It does work OK in development, even with both connections pointing at a linux docker instance.  i would imagine it's just ignored when connecting to a non cloud instance 🤷‍♂

PR #1443, opened 2019-12-12
Feature/37648 Removed await from pupil pref calls

GeorgeChatzigiannis:


GeorgeChatzigiannis:
@GuyHarwood aside from confirming if that particular method is called, tests are checking the navigation path in a stable component state so 'awaiting' did not affect them.

PR #1447, opened 2019-12-13
Feature/36390 check notifier unhappy path

GuyHarwood:
marks a check with new status code of 'ERR' when processing fails.


tslib/src/functions/check-marker/check-marker.v1.ts
```diff
@@ -33,16 +33,16 @@ export class CheckMarkerV1 {
...
     functionBindings.checkNotificationQueue = []
     if (markingData === undefined) {
-      const notification: ICheckNotificationMessage = {
-        checkCode: validatedCheck.RowKey,
-        notificationType: CheckNotificationType.checkInvalid,
-        version: 1
-      }
-      functionBindings.checkNotificationQueue.push(notification)
+      this.notifyProcessingFailure(validatedCheck, functionBindings)
+      return
```
GeorgeChatzigiannis: isn't the return redundant here?

tslib/src/functions/check-notifier/check-notifier.data.service.ts
```diff
@@ -13,6 +13,21 @@ export class CheckNotifierDataService implements ICheckNotifierDataService {
...
   }
 
+  markCheckAsProcessingFailed (checkCode: string): Promise<void> {
+    const checkCodeParam: ISqlParameter = {
+      type: mssql.UniqueIdentifier,
+      name: 'checkCode',
+      value: checkCode
+    }
+    const sql = `UPDATE [mtc_admin].[check]
+      SET checkStatus_id=
```
jon-shipley: Shouldn't we also set the `processingFailed` flag?

tslib/src/functions/check-notifier/check-notifier.data.service.ts
```diff
@@ -21,14 +36,11 @@ export class CheckNotifierDataService implements ICheckNotifierDataService {
     }
     const checkRequest: ITransactionRequest = {
       sql: `UPDATE [mtc_admin].[check]
-      SET checkStatus_id=@checkStatusId
+      SET checkStatus_id=(SELECT cs.id FROM
+        [mtc_admin].[checkStatus] cs
+        WHERE cs.code='CMP')
```
jon-shipley: If we are setting the status to complete, should we also set `pupil.checkComplete`, `check.complete`, `check.completedAt` ?

tslib/src/functions/pupil-prefs/pupil-prefs.service.ts
```diff
@@ -84,7 +84,7 @@ export class PupilPrefsDataService implements IPupilPrefsDataService {
     return this.sqlService.modifyWithTransaction(requests)
   }
 
-  // TODO why not add a lookup to redis on check creation to avoid sql hits?
+  // TODO : PERF why not add a lookup to redis on check creation to avoid sql hits?
```
jon-shipley: 👍

tslib/src/functions/check-marker/check-marker.v1.ts
GuyHarwood: no, because then the next block will run and notify the teacher of completion, which is not wanted

tslib/src/functions/pupil-prefs/pupil-prefs.service.ts
GuyHarwood: i had this as one item in a PBI, but ive promoted it to its own card now

PR #1454, opened 2019-12-18
Feature/38622 load test setup

GuyHarwood:
Azure CLI script to create a database (with optional replica), run migrations and seed data, update apps to point at new database and clean up after load test complete.

- [x] load test setup script
- [x] dedicated `infra` app for infrastructure management
- [x] migrations & seeds under `infra` app (example at this point)

**Known issues**
admin migrations are executed against the database as defined in the `SQL_DATABASE` environment variable (defaults to `mtc`).  The load test setup script (`prep-up.sh`) creates a dynamically named database, which will need passing to the migrator.  Rather than implement this into the admin migrations this will be added to the `infra/sql` standalone migrator.


load-test/prep-up.sh
```diff
@@ -0,0 +1,87 @@
...
+RES_GROUP_FUNCTIONS=$2
+SQL_SERVER=$3
+DB_SCALE=$4
+ADMIN_APP=$5
+FUNC_CONSUMP=$6
+FUNC_APPSVC=$7
+SQL_SERVER_REPLICA=$8
+
+# Create database with unique name
+DB_SUFFIX=$(openssl rand -hex 4)
```
jon-shipley: Is `openssl` guaranteed to be installed?  What happens if it isn't?

GuyHarwood: that's a very good point....

GuyHarwood: the idea is that this will be run on a testers/developers machine, so i think we can just demand it as a requirement...

jon-shipley: That's fine then 👍

PR #1455, opened 2019-12-19
Feature/38591 pupiluuid lookup

GuyHarwood:
- [x] Remove obsolete check payload items (school.id & pupil.id)
- [x] Add pupil uuid lookup by check code to redis
- [x] utilise lookup in pupil prefs function to avoid sql call


admin/services/prepare-check.service.js
```diff
@@ -61,10 +66,14 @@ const service = {
   }
 }
 
-function buildLookupKey (checkCode) {
+function buildPreparedCheckLookupKey (checkCode) {
```
jon-shipley: There is a `redisKeyService` that seems like a good fit for this code.

GuyHarwood: ah yes, thanks for the reminder 👍

admin/services/redis-key.service.js
```diff
@@ -20,6 +20,23 @@ const redisKeyService = {
...
+    return `pupil-uuid-lookup:${checkCode}`
+  },
+
+  /**
+   * return the key used to store the prepared check entry for pupil API
+   * @param {string} schoolPin
+   * @param {number} pupilPin
+   */
+  getPreparedCheckKey (schoolPin, pupilPin) {
+    return `preparedCheck:${schoolPin}:${pupilPin}`
```
jon-shipley: 👍

PR #1466, opened 2020-01-13
Feature/37901 register input question data

GeorgeChatzigiannis:
Question data should be passed from component to service instead of having the service reading the DOM to acquire them.


pupil-spa/src/app/services/register-input/registerInput.service.spec.ts
```diff
@@ -70,8 +80,13 @@ describe('RegisterInputService', () => {
...
         type: 'mousedown', which: 1, currentTarget: null
       };
-      registerInputService.addEntry(event);
-      registerInputService.addEntry(event);
+      const questionData = {
+        questionNumber: '1',
+        factor1: '1',
+        factor2: '12',
+      };
+      registerInputService.addEntry(event, questionData);
```
GuyHarwood: are 2 subsequent calls intentional?

GeorgeChatzigiannis: I didn't find any particular value or reason behind having a second call so I removed it.

GeorgeChatzigiannis: thank you for spotting it

GuyHarwood:
@jon-shipley as long as azure does not use `yarn start` they should be ok.

GeorgeChatzigiannis:
`prestart` has been added back

PR #1470, opened 2020-01-15
Feature/37898 Utilise event timestamp for register input event

GeorgeChatzigiannis:
Rely on event timestamps during storing a register input otherwise default to standard date timestamp


pupil-spa/src/app/services/register-input/registerInput.service.spec.ts
```diff
@@ -68,7 +68,7 @@ describe('RegisterInputService', () => {
   it('calls the storage service', inject([TestRegisterInputService],
     (registerInputService: TestRegisterInputService) => {
       const event = {
-        type: 'mousedown', which: 1, currentTarget: null
+        type: 'mousedown', which: 1, currentTarget: null, timeStamp: 1519211809934
```
jon-shipley: I would expect to see a test with an undefined timestamp too, to mimic those browsers that do not add event.timestamp, and undefined is the expected output from accessing a missing property.  A timestamp should be added.

GeorgeChatzigiannis: Added in the `storeEntry` tests block

PR #1474, opened 2020-01-17
bug/39032 check started function - add null guard on prepared check lookup

GuyHarwood:
add a null guard when looking up prepared check, as it could have been deleted.


tslib/src/functions/check-started/check-started.service.ts
```diff
@@ -26,7 +26,8 @@ export class CheckStartedService {
       clientCheckStartedAt: checkStartedMessage.clientCheckStartedAt
     })
     const preparedCheck = await this.redisService.get(preparedCheckKey)
-    if (preparedCheck.config.practice === false) {
+
+    if (preparedCheck && preparedCheck.config.practice === false) {
```
GeorgeChatzigiannis: possible enhancement:  `preparedCheck && preparedCheck.config &&preparedCheck.config.practice === false`

jon-shipley: I think this should either be written in pure js style:

`if (preparedCheck && preparedCheck.config && preparedCheck.config.practice === false) {`

Or, something like

`if(R.path(['config', 'practice'], preparedCheck) === true`

GeorgeChatzigiannis: The ramda approach is 👌🏼

PR #1479, opened 2020-01-21
Test support

activemq:
Cache bust the local check window on the school landing page

This is for test support, to help identify an issue.


GuyHarwood:
few questions...
- why do we need to do this?
- is this a temporary or permanent change?
- what does the `true` flag change in behaviour?

jon-shipley:
> few questions...
> 
> * why do we need to do this?
> * is this a temporary or permanent change?
> * what does the `true` flag change in behaviour?

The cacheBust flag (boolean) provides an easy way to remove the checkWindow cache from the webserver, allowing it to fallback to retrieving it from Redis, or the DB if it is not in Redis.

I added it when I saw that we were making a round trip to Redis from the middleware for every single route, and quite often looking it up again in the controller, and again in the services. 😱

GuyHarwood:
@jon-shipley thanks for clarification.  i was concerned it was dropping it from redis.
@activemq could you let me know about the other 2 questions please :)

activemq:
- is this a temporary or permanent change?
@GuyHarwood this is likely to be a perm change (if you agree), as it means that when the cache is cleared it repopulates Redis with the value as soon as the homepage is loaded. We need this for our tests as currently it is cached for approx 60 secs. Thus causing flakiness in the tests.

The other option is to have the tests to refresh the page until the Redis cache is populated. This will add more wait time for every scenario.

- what does the true flag change in behaviour?
It means it repopulates the cache if the 'checkWindow.sqlFindActiveCheckWindow' entry is not in Redis whenever the homepage is loaded.

GuyHarwood:
@activemq thanks Mo.  happy to approve given the context

PR #1480, opened 2020-01-21
Feature/39076 loadtest live pin gen

GuyHarwood:
- [x] beanshell script to support multi-node load tests with unique teachers across each node
- [x] configurable school count for dummy data injection
- [x] script to purge and rebuild local db with dummy data injection
- [x] try it out pin gen scenario
- [x] live pin gen scenario
- [x] live pin gen with check start and check submission


deploy/sql/config.js
```diff
@@ -36,5 +36,8 @@ module.exports = {
     MaxRetryAttempts: parseInt(process.env.RETRY_MAX_ATTEMPTS, 10) || 3,
     InitialPauseMs: parseInt(process.env.RETRY_PAUSE_MS, 10) || 5000,
     PauseMultiplier: parseFloat(process.env.RETRY_PAUSE_MULTIPLIER) || 1.5
+  },
+  DummyData: {
```
jon-shipley: I think I would prefer this to be in the main config as this is not strictly speaking SQL config.

load-test/check-submit-proxy/.vscode/launch.json
```diff
@@ -0,0 +1,12 @@
+{
```
jon-shipley: This vscode dir should probably not be checked in.

GuyHarwood: its optional.  the vscode team do check it in, so they can share debugger config etc.
there is nothing in here that is user specific, it provides support for the azure functions tooling.

i'd be happy with `.idea` and `.vscode` folders, as long as they only include settings that are useful & non-invasive to other team members.

not fussed either way

jon-shipley: I did caveat it with "probably".  If you intentionally checked it in then by all means leave it.

deploy/sql/config.js
GuyHarwood: it is in the main config file, for the sql app.

PR #1484, opened 2020-01-23
feature/39138 name pin elements on pin list

GuyHarwood:
add explicit element ids for pin values to aid xpath lookups during load tests

### DIFF
--- a/admin/views/pupil-pin/view-and-custom-print-pins.ejs
+++ b/admin/views/pupil-pin/view-and-custom-print-pins.ejs
@@ -96,16 +96,18 @@
                                 <% } %>
                                 <input type="hidden" id="pupilUpn" name="pupilUpn" value="<%= pupil.upn %>"/>
                                 <div>
-                                    <div class="pin-content">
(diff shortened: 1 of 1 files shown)


admin/views/pupil-pin/view-and-custom-print-pins.ejs
```diff
@@ -96,16 +96,18 @@
...
-                                    <div class="pin-content">
-                                        <span class="label">School Password: </span>
-                                        <span class="information"><%= (!!school ? school.pin : 'No active password') %></span>
-                                        <span class="separator"></span>
-                                        <span class="label">PIN: </span>
-                                        <span class="information">
-                                          <%= pupil.pin %>
+                                  <div class="pin-content">
+                                      <span class="label">School Password: </span>
+                                      <span class="information" id="schoolPin" name="schoolPin">
```
jon-shipley: The `name` attribute is not a valid attribute for `span`.  In any case, `id` should be enough?

PR #1486, opened 2020-01-27
39150 - pupil is not removed from pin list

jon-shipley:
The view vewPupilsWithActiveLivePins will now only show pins for pupils who have not logged in.  This closes a gap in the system as:

1) Logging in will set the expiry on the prepared check to 30 mins
2) Clicking "Start Now" will expire the prepared check immediately

This removes the pupil from the pin gen screen as soon as they log in.


admin/data/sql/migrations/20200127112400.do.alter-view-vewpupilswithactivelivepins.sql
```diff
@@ -0,0 +1,24 @@
+ALTER VIEW [mtc_admin].[vewPupilsWithActiveLivePins] AS
```
GeorgeChatzigiannis: Just to follow the strategy that we have for other migrations shouldn't that be `CREATE OR ALTER` ?

jon-shipley: Good point.  Amended.  Thanks.

PR #1487, opened 2020-01-27
Feature/39126 Pupil status UI p1

GeorgeChatzigiannis:
- [x] Added basic pupil status UI page
- [x] Introduced relevant admin backend services (including support for additional pupil states)
- [x] Removed pupil register v1 implementation and config
- [x] Point existing pupil status calls to the new service
- [x] Removed incomplete check calculation from school home


func-consumption/package.json
```diff
@@ -54,4 +54,4 @@
     "random-number-csprng": "^1.0.2",
     "uuid": "^3.3.3"
   }
-}
+}
```
GuyHarwood: newline?

admin/services/pupil-status.service.js
```diff
@@ -0,0 +1,142 @@
...
+    })
+    return newPupil
+  },
+
+  /**
+   * Return the process status using new pupil status fields
+   * @param {ProcessStatusArg} arg object containing the parameters
+   * @return {string}
+   */
+  getProcessStatusV2: function (arg) {
```
GuyHarwood: this looks familiar.  have you updated other call sites?

GuyHarwood: please ignore, i now see its on your todo list

admin/controllers/school.js
```diff
@@ -36,6 +36,7 @@ controller.getSchoolLandingPage = async (req, res, next) => {
     const currentDate = moment.tz(req.user.timezone || config.DEFAULT_TIMEZONE)
     const resultsOpeningDay = resultPageAvailabilityService.getResultsOpeningDate(currentDate, checkWindowData.checkEndDate)
     const isResultsFeatureAccessible = resultPageAvailabilityService.isResultsFeatureAccessible(currentDate, resultsOpeningDay)
+    // TODO: Determine if the below call will still be needed
```
GeorgeChatzigiannis: @GuyHarwood @jon-shipley I believe the below call will be redundant on school home page but I'd like to get confirmation before removing it.

admin/services/pupil-status.service.js
```diff
@@ -0,0 +1,145 @@
...
+      processingFailed,
+      reason
+    } = arg
+
+    if (pinExpiresAt && !moment.isMoment(pinExpiresAt)) {
+      throw new Error('pinExpiresAt must be null or a Moment.moment datetime')
+    }
+
+    if (isPositive(attendanceId)) {
+      status = reason
```
GeorgeChatzigiannis: This is more of a UI demand from both HDF and pupil status page and it can be moved to a presenter rather than keeping it in the service.
My question is do we maintain the status at that level as `Not taking the Check` or do we allow storing the actual reason?

admin/services/data-access/pupil-status.data.service.js
```diff
@@ -0,0 +1,101 @@
+'use strict'
+
+const { TYPES } = require('./sql.service')
+const R = require('ramda')
+const sqlService = require('./sql.service')
+const pupilStatusDataService = {}
+
+pupilStatusDataService.sqlFindPupilsFullStatus = async function sqlFindPupilsAndAttendanceReasons (schoolId) {
```
GuyHarwood: how comes the property and function name are different?

admin/services/pupil-status.service.js
GuyHarwood: given its shared across both controllers i would leave it as a service, rather than a particular presenter

admin/services/data-access/pupil-status.data.service.js
GeorgeChatzigiannis: good catch, fixed.

admin/services/data-access/pupil-status.data.service.js
```diff
@@ -0,0 +1,101 @@
...
+const sqlService = require('./sql.service')
+const pupilStatusDataService = {}
+
+pupilStatusDataService.sqlFindPupilsFullStatus = async (schoolId) => {
+  if (!schoolId) {
+    throw new Error('schoolId param is required')
+  }
+  const paramSchoolId = { name: 'schoolId', type: TYPES.Int, value: schoolId }
+  const sql = `
+  SELECT
```
GuyHarwood: please extract this to a view

admin/controllers/pupil-status.js
```diff
@@ -0,0 +1,29 @@
+// const config = require('../config')
```
jon-shipley: I wonder if this could be deleted?

admin/services/setting.service.js
```diff
@@ -24,18 +24,22 @@ settingService.update = async (loadingTimeLimit, questionTimeLimit, checkTimeLim
...
  * Get check settings
+ * @param {Boolean} cacheBust
  * @returns {questionTimeLimit: number, loadingTimeLimit: number, checkTimeLimit: number}
  */
-settingService.get = async () => {
-  let settings = await settingDataService.sqlFindOne()
-  if (!settings) {
-    settings = {
+let cachedSettings
+settingService.get = async (cacheBust = false) => {
```
jon-shipley: Should there be an expiry on the local cache here? Say 5 minutes?  If not, the settings are unchangeable without a web app reboot.

GuyHarwood: good catch 🥇

GeorgeChatzigiannis: yea I did miss that

PR #1491, opened 2020-01-29
Feature/38466 remove check state check started

jon-shipley:
Remove checkStatus check started state


start.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -eu
```
GuyHarwood: 💯

admin/data/sql/migrations/20200129155749.do.fix-db-routine.sql
```diff
@@ -0,0 +1,113 @@
...
+        @pinExpiresAt IS NOT NULL AND
+        @pinExpiresAt > GETUTCDATE())
+        BEGIN
+            SELECT @ret = id FROM [mtc_admin].[checkStatus] where code = 'NEW';
+            RETURN @ret;
+        END
+
+    -- EXPIRED
+    IF ((@pupilLoginDate IS NULL OR
+         @pupilLoginDate IS NOT NULL AND @isLiveCheck = 0) AND
```
GeorgeChatzigiannis: is this correct `@isLiveCheck = 0`?  that would only point to try out expired checks?

jon-shipley: Live checks only go to expired if no-one logged in.
TIO checks go to expired even if they were used.

PR #1497, opened 2020-02-04
spike/39361 - cache check forms after first fetch during pin generation

GuyHarwood:
Allocated Check forms are static during a check window and are only modified in exceptional circumstances.   They are fetched from the database every time a teacher generates one or more pins.

- Cache the check forms for 1 month after first fetch.  
- Drop cache when check window form assignments are modified.
- remove obsolete check form modification code from old check form service

### DIFF
--- a/admin/services/data-access/redis-cache.service.js
+++ b/admin/services/data-access/redis-cache.service.js
@@ -31,7 +31,7 @@ const redisCacheService = {}
 redisCacheService.get = async key => {
   redisConnect()
   try {
-    logger.info(`REDIS (get): retrieving ${key}`)
+    logger.debug(`REDIS (get): retrieving ${key}`)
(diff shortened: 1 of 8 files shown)


admin/services/data-access/redis-cache.service.js
```diff
@@ -31,6 +31,7 @@ const redisCacheService = {}
 redisCacheService.get = async key => {
   redisConnect()
   try {
+    logger.info(`REDIS (get): retrieving ${key}`)
```
jon-shipley: I am probably reading this out of context, but I think that the Redis logger calls should exist at the same level as the SQL ones, `debug`?

GuyHarwood: all changed

PR #1499, opened 2020-02-04
Feature/39288 [ARCH] change check notifier to timed batch message receiver

GuyHarwood:
the check-notifier can saturate sql with incoming updates if not controlled.
moving the function to a timer based trigger allows us to stem the flow and send updates in batches.

- default of 32 messages per batch
- default of 5 batches per timer execution
- singleton, timer based function
- manages service bus connection entirely
- verbose logging for diagnostics under load
- existing check-notifier disabled


tslib/src/functions/check-notifier-batch/index.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ICheckNotificationMessage } from '../check-notifier/check-notification-message'
+import { BatchCheckNotifier } from './batch-check-notifier.service'
+
+const batchCheckNotifier: AzureFunction = async function (context: Context, timer: any): Promise<void> {
+  const start = performance.now()
+
+  if (!config.ServiceBus.ConnectionString) {
+    throw new Error('ServiceBusConnection env var is required')
+  }
+
```
GuyHarwood: add graceful dropout if connection unavailable

tslib/src/functions/check-notifier-batch/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  const busClient = sb.ServiceBusClient.createFromConnectionString(config.ServiceBus.ConnectionString)
+  const queueClient = busClient.createQueueClient('check-notification')
+  const receiver = queueClient.createReceiver(sb.ReceiveMode.peekLock)
+  const batchNotifier = new BatchCheckNotifier()
+
+  const messageBatch = await receiver.receiveMessages(config.ServiceBus.BatchReceiveCount)
+  const notifications: ICheckNotificationMessage[] = []
+  for (let index = 0; index < messageBatch.length; index++) {
```
GuyHarwood: enhance to perform multiple batches on each timer trigger.

tslib/src/functions/check-notifier-batch/index.ts
```diff
@@ -0,0 +1,98 @@
...
+* The function is running as a singleton, and the receiver is therefore exclusive
+  we do not expect another receive operation to be in progress.
+  if the message is abandoned 10 times (the current 'max delivery count') it will be
+  put on the dead letter queue automatically.
+*/
+const batchCheckNotifier: AzureFunction = async function (context: Context, timer: any): Promise<void> {
+
+  if (timer.IsPastDue) {
+    context.log('timer is past due, exiting...')
+    return
```
jon-shipley: 👍 nice

tslib/src/functions/check-notifier-batch/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      await batchNotifier.notify(notifications)
+      context.log('batch processed successfully, completing all messages in batch')
+      messageBatch.forEach(async msg => {
+        await msg.complete()
+      })
+    } catch (error) {
+      context.log.error(error.message)
+      messageBatch.forEach(async msg => {
+        try {
+          await msg.abandon()
```
jon-shipley: If it has called  `complete` on half of the messages, this will then call `abandon` even on messages that have been completed.  Is that okay?

tslib/src/functions/check-notifier-batch/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    throw new Error(`errors occured during processing...\n${errorList}`)
+  }
+
+  const end = performance.now()
+  const durationInMilliseconds = end - start
+  const timeStamp = new Date().toISOString()
+  context.log(`${functionName}: ${timeStamp} run complete: ${durationInMilliseconds} ms`)
+}
+
+export default batchCheckNotifier
```
jon-shipley: This monolith should be broken up into functions.

tslib/src/functions/check-notifier-batch/index.ts
GuyHarwood: good catch 🥇  unfortunately i cannot complete in a transaction, so will have to figure something out

tslib/src/functions/check-notifier-batch/index.ts
GuyHarwood: agree!

GuyHarwood: in fact, it feels like the majority of the logic should be moved to the service

GuyHarwood: refactored

tslib/src/functions/check-notifier-batch/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    const msg = messageBatch[index]
+    try {
+      await msg.abandon()
+    } catch (error) {
+      context.log.error(`unable to abandon message:${error.message}`)
+    }
+  }
+}
+
+function exit (start: number, context: Context) {
```
jon-shipley: Ideally, functions should be declared before they are used.  This is calling the final callback rather process exit.  Perhaps we could call it something else?

E.g. `done()` `cleanupAndFinish()`

GeorgeChatzigiannis:
@GuyHarwood shouldn't there be a corresponding test file for `tslib/src/functions/check-notifier-batch/index.ts` ?

GuyHarwood:
@GeorgeChatzigiannis no, these are typically just for bootstrapping the service calls.  admittedly it is busier than most, however its still POC at this point, we need to get it under load test, so further work can be done if its deemed of value

PR #1506, opened 2020-02-06
bug/39419 move function to mtc_admin schema

GuyHarwood:
move `ufnCalcCheckStatusID` to `mtc_admin` schema and update calls


GuyHarwood:
@activemq @sinhaamit1980 this is a very small change that is only used when requesting a restart

GuyHarwood:
@sinhaamit1980 @activemq only affects restart creation.  light manual QA is sufficient.  release will not go past load-test env if incorrect behaviour observed

PR #1519, opened 2020-02-07
Bug/39404 only perform async retry on timeout

GuyHarwood:
- [x] implement default retry policy of 'no retry'
- [x] add additional predicate for `mssql` specific errors on db timeouts
- [x] implement in tslib
- [x] implement in admin
- [x] implement in functions-app


GuyHarwood:
@sinhaamit1980 @activemq we can only prove this in load test env.  light manual load test is sufficient.  release will not go past load-test env if incorrect behaviour observed

PR #1525, opened 2020-02-12
Feature/39198 remove check pins once expired

jon-shipley:
Add function to consumption plan to delete expired checkPins every eight hours.


func-consumption/check-pin-expiry/function.json
```diff
@@ -0,0 +1,12 @@
+{
+  "bindings": [
+    {
+      "_ref": "sec, min, hour, day, month, dow",
```
GuyHarwood: whats the `_ref` for?  is it an undocumented setting?

tslib/src/functions/check-pin-expiry/check-pin-expiry.service.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export class CheckPinExpiryService {
+  private sqlService: ISqlService
+
+  constructor (theSqlService: ISqlService) {
+    this.sqlService = theSqlService
+  }
+
+  async process () {
+    const sql = `DELETE FROM [mtc_admin].[checkPin] WHERE pinExpiresAt < GETUTCDATE()`
```
GuyHarwood: i did question why you didn't implement a dedicated data service, but this is leaner and the service is simple enough 👌

func-consumption/check-pin-expiry/function.json
jon-shipley: It's a comment to remind the viewer what the format of the cron expression is.   As this is JSON commenting has to be a field.

GuyHarwood: good idea 💡

GuyHarwood:
@jon-shipley conflict in tslib/package.json

PR #1527, opened 2020-02-12
Bug/39494 async retry predicate fix

GuyHarwood:
- filter out stack trace from retry logging to improve readability
- refactor timeout predicate based on load test findings
- introduce new 'limit reached' predicate for sql azure


GuyHarwood:
Critical path tested locally, further testing to be undertaken in load-test env

PR #1534, opened 2020-02-16
Feature/39312 Introduced wait page between POST and GET on pupil edit

GeorgeChatzigiannis:
- [x] Implement meta refresh on pupil edit post request with configurable delay and url redirect parameter
- [x] Test browser behavior: IE11, Edge, Firefox, Chrome, Safari


admin/views/redirect-delay.ejs
```diff
@@ -0,0 +1,8 @@
+<meta http-equiv="refresh" content="<%= WaitTimeBeforeRedirectInSeconds %>;url=<%= url %>" />
+<main class="govuk-main-wrapper" id="main-content" role="main">
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds wait-redirect">
+            <h1 class="govuk-heading-xl">Please wait, redirecting...</h1>
```
GuyHarwood: Not sure if you were present, but there was some discussion around the language to use here.  You should seek clarification but the view was that something along the lines of ‘saving changes’ would be best

admin/views/redirect-delay.ejs
```diff
@@ -0,0 +1,8 @@
+<meta http-equiv="refresh" content="<%= WaitTimeBeforeRedirectInSeconds %>;url=<%= url %>" />
```
jon-shipley: Can you confirm that this `meta` tag will be output in the document 'HEAD'?

admin/config.js
```diff
@@ -57,6 +57,7 @@ module.exports = {
   SESSION_SECRET: process.env.NODE_ENV === 'production' ? process.env.SESSION_SECRET : 'anti tamper for dev',
   WEBSITE_OFFLINE: {}.hasOwnProperty.call(process.env, 'WEBSITE_OFFLINE') ? toBool(process.env.WEBSITE_OFFLINE) : false,
   WaitTimeBeforeExitInSeconds: parseInt(process.env.WAIT_TIME_BEFORE_EXIT, 10) || 30,
+  WaitTimeBeforeRedirectInSeconds: parseInt(process.env.WAIT_TIME_BEFORE_REDIRECT, 10) || 5,
```
jon-shipley: 5 seconds seems a very long time to wait by default.

GuyHarwood: if i recall correctly they said 2 seconds was OK?

GeorgeChatzigiannis: yes you are right, I changed it to 2

admin/views/redirect-delay.ejs
GeorgeChatzigiannis: Moved meta tag to head

PR #1535, opened 2020-02-17
Feature/39549 cache sas tokens for 1 hour configurable

jon-shipley:
Cache implementation for SAS tokens when generating checks.


admin/services/check-start.service/check-start.service.spec.js
```diff
@@ -227,19 +239,24 @@ describe('check-start.service', () => {
         expect(Object.keys(res[0].questions[0])).toContain('factor1')
         expect(Object.keys(res[0].questions[0])).toContain('factor2')
       })
+
       it('does generate and include check complete & check submit sas tokens when live checks are generated', async () => {
         const res = await checkStartService.createPupilCheckPayloads([1], 1)
-        expect(sasTokenService.generateSasToken).toHaveBeenCalledTimes(4)
```
GeorgeChatzigiannis: perhaps replace instead of delete this expectation with:
expect(sasTokenService.getTokens).toHaveBeenCalledTimes(1)
or make an isolated test to ensure the method is being called?

admin/services/sas-token.service.js
```diff
@@ -16,7 +22,22 @@ const sasTokenService = {
...
-  generateSasToken: function (queueName, expiryDate, serviceImplementation) {
+  generateSasToken: async function (queueName, expiryDate, serviceImplementation) {
+    const start = performance.now()
+    // See if a valid token can be retrieved from redis
+    const redisKeyName = redisKeyService.getSasTokenKey(queueName)
+    try {
+      const token = await redisCacheService.get(redisKeyName)
+      if (token) {
+        const end = performance.now()
+        logger.debug(`generateSasToken(): took ${end - start} ms`)
```
GeorgeChatzigiannis: 👌

admin/services/sas-token.service.js
```diff
@@ -49,11 +70,56 @@ const sasTokenService = {
...
 
-    return {
+    const tokenObject = {
       token: sasToken,
       url: serviceImplementation.getUrl(queueName),
       queueName: queueName
     }
+
+    // Store the sasToken in redis
+    if (redisKeyName) {
```
GeorgeChatzigiannis: Isn't that if block redundant?
`redisKeyName` will always be set as the `redisKeyService.getSasTokenKey` returns a string regardless.

admin/services/check-start.service/check-start.service.spec.js
jon-shipley: @GeorgeChatzigiannis Upon reflection I have deleted the whole test as it is already covered by the `getTokens()` tests.  In this case we have already mocked the `getTokens()` call so this test was only confirming it had beeen mocked correctly.  I don't see that as adding much value.

admin/services/sas-token.service.js
jon-shipley: Good catch

PR #1546, opened 2020-02-19
Feature/39110 pupil census data import

jon-shipley:
Port the previous pupil census upload to the new long-running functions.


functions-app/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Disabling functions Locally
+
+1. You need an environment variable named `AzureWebJobs.<function-name>.Disabled` set to 'true'
+2. There is an example env file `disable-functions.env` that may be used:
+    * `env $(cat disable-functions.env | grep Azure | xargs ) yarn start`
+
+## Creating a new function
+
+From the `functions` directory, execute `func function new`.
```
GuyHarwood: didn't know that was a command, `func new` is shorter and what i've always used

functions-app/README.md
```diff
@@ -0,0 +1,17 @@
+# Functions App
```
GuyHarwood: 💯

functions-app/census-import/census-import.data.service.js
```diff
@@ -0,0 +1,100 @@
...
+const R = require('ramda')
+
+let pool
+
+module.exports.initPool = async function initPool (context) {
+  const poolConfig = {
+    database: config.Sql.Database,
+    server: config.Sql.Server,
+    port: config.Sql.Port,
+    requestTimeout: 10 * 60 * 1000,
```
GuyHarwood: please can you move this to a config variable with a default value the same as whats in admin app.

i am currently tweaking these in load tests and it has a profound effect on the database load behaviour when coupled with the retry mechanism

functions-app/census-import/census-import.data.service.js
```diff
@@ -0,0 +1,100 @@
...
+    connectionTimeout: config.Sql.Timeout,
+    user: config.Sql.PupilCensus.Username,
+    password: config.Sql.PupilCensus.Password,
+    pool: {
+      min: 1,
```
GuyHarwood: as above, please move to config

functions-app/README.md
jon-shipley: The 'function' context is optional, so both forms are correct.

functions-app/census-import/census-import.data.service.js
jon-shipley: I am happy to move it to a config variable, but the current 10-minute request timeout may be too short, rather than too long.   The default request timeout for the functions-app is five minutes which is too short.

I don't see how this function is affected by the load tests - pupil load takes place before any other load is placed on the system.

GuyHarwood: the default value of 10 minutes is ok, its just about having it configurable in case we need to support production.  we don't want to have to redeploy.  and yes, this is literally a 'run once before we go live' thing

jon-shipley: Ok.  It will almost certainly take longer than 10 mins to run the main SQL, so it's set accordingly.

functions-app/census-import/census-import.data.service.js
jon-shipley: Now using default config values

functions-app/config.js
```diff
@@ -31,6 +32,7 @@ module.exports = {
     Port: process.env.SQL_PORT || 1433,
     Timeout: process.env.SQL_TIMEOUT || oneMinuteInMilliseconds,
     requestTimeout: parseInt(process.env.SQL_REQUEST_TIMEOUT, 10) || fiveMinutesInMilliseconds,
+    censusRequestTimeout: parseInt(process.env.SQL_CENSUS_REQUEST_TIMEOUT, 10) || twoHoursInMilliseconds,
```
GuyHarwood: i think you will be pleasantly surprised by the speed of the bulk import, but let's keep this high by default to avoid issue

PR #1548, opened 2020-02-20
Feature/39655 vewpupilseligibleforlivepingeneration is expensive

jon-shipley:
* Add index to speed up view
* Move ordering from DB to web-app


jon-shipley:
Kudos to the Ramda cookbook for the recipe.
https://github.com/ramda/ramda/wiki/Cookbook#sort-a-list-by-array-of-props-if-first-prop-equivalent-sort-by-second-etc 👍

PR #1569, opened 2020-02-27
tarn pool manager requires integers

GuyHarwood:
Fix for breaking change in `mssql` lib where underlying pool no longer allows string values for pool connection range settings.


activemq:
@GuyHarwood manually tested and full set of tests have been run.

PR #1573, opened 2020-02-28
Feature/39662 New cookies form and functionality to opt out from google analytics

GeorgeChatzigiannis:
- Added lib scripts, styles and markup from newest GDS cookie banner and opt in form view.
- Added new route for opt in view and preserved existing mtc cookie report in a new dedicated route.


admin/views/cookies-mtc.ejs
```diff
@@ -80,30 +80,9 @@
             <h3 class="govuk-heading-m">Our performance survey</h3>
 
             <p class="govuk-body">
-                We use <a class="govuk-link" target="_blank" rel="noopener noreferrer" href="https://www.smartsurvey.co.uk/">SmartSurvey</a>
-                <img alt="" src="/images/icon-link.png"/> to collect responses to the survey.
-                If you take part, SmartSurvey will save extra cookies to your computer to track your progress through it.
+                To be updated.
```
jon-shipley: Just checking this is correct.  Why not just say nothing until we want to say something?

GeorgeChatzigiannis: @jon-shipley I was instructed to leave this as to be updated.

PR #1574, opened 2020-02-28
bug/39882 fix vewPupilsEligibleForTryItOutPin to exclude expired checks

GuyHarwood:


admin/data/sql/migrations/20200228150155.do.fix-vewPupilsEligibleForTryItOutPin.sql
```diff
@@ -0,0 +1,40 @@
...
+      AND p.id NOT IN (
+        SELECT p2.id
+        FROM
+             [mtc_admin].[pupil] p2 LEFT JOIN
+             [mtc_admin].[check] AS chk ON (p.id = chk.pupil_id) LEFT JOIN
+             [mtc_admin].[checkStatus] AS chkStatus ON (chk.checkStatus_id = chkStatus.id)
+        WHERE chk.isLiveCheck = 0
+          AND chkStatus.code IN ('NEW', 'COL')
+          AND p2.school_id = p.school_id
+          AND cp.pinExpiresAt < GETUTCDATE()
```
jon-shipley: 1. `cp` refers to a table from the outer expression
2. `cp.pinExpiresAt` could be null if the pin had expired and been cleaned up, but comparisons involving nulls depending on whether ansi nulls are set to on or off.  If on, then this expression will return UNKNOWN rather than false.

MS has this note
"In a future version of SQL Server, ANSI_NULLS will be ON and any applications that explicitly set the option to OFF will generate an error. Avoid using this feature in new development work, and plan to modify applications that currently use this feature."
https://docs.microsoft.com/en-us/sql/t-sql/statements/set-ansi-nulls-transact-sql?view=sql-server-ver15

Which suggests we should deal with nulls separately for future SQL versions.

GuyHarwood: Great catch.

I have moved the join to the inner sub-query where the exclusion applies.  The filter was also the wrong way round, as we want to exclude anyone with an existing TIO check.  I will put that one down to a long and busy week 😄 

regards `ANSI_NULLS` - as the filter is only interested in records where they have a TIO check with a date in the future, its a non-issue here.  However, there is an overall concern when MS switch over to `ON` by default.  A cursory search on the entire codebase suggests for the most part we are ok, as we have been in the habit of using `WHERE [colA] IS NULL` as opposed to `WHERE [colA] = NULL`.

I will create a ticket now to cover the analysis and explicit switchover.

admin/data/sql/migrations/20200228150155.do.fix-vewPupilsEligibleForTryItOutPin.sql
```diff
@@ -0,0 +1,40 @@
...
+    LEFT JOIN [mtc_admin].[checkStatus] cs ON (c.checkStatus_id = cs.id)
+  WHERE
+      -- don't allow a try it out check for anyone who has left
+      p.attendanceId IS NULL
+    -- and exclude anyone who has a check that is not new
+    AND (
+        p.currentCheckId IS NULL
+    OR
+    (p.currentCheckId IS NOT NULL AND c.pupilLoginDate IS NULL)
+      )
```
jon-shipley: unformatted brace?

admin/data/sql/migrations/20200228150155.do.fix-vewPupilsEligibleForTryItOutPin.sql
```diff
@@ -0,0 +1,40 @@
...
+        SELECT p2.id
+    FROM
+      [mtc_admin].[pupil] p2
+      LEFT JOIN [mtc_admin].[check] AS chk ON (p.id = chk.pupil_id)
+      LEFT JOIN [mtc_admin].[checkStatus] AS chkStatus ON (chk.checkStatus_id = chkStatus.id)
+      LEFT JOIN [mtc_admin].[checkPin] cp ON (c.id = cp.check_id)
+    WHERE chk.isLiveCheck = 0
+      AND chkStatus.code IN ('NEW', 'COL')
+      AND p2.school_id = p.school_id
+      AND cp.pinExpiresAt > GETUTCDATE()
```
jon-shipley: As the `checkPin` table is Left-Joined the `pinExpiresAt` could still be null - can you add in a null check preceding the comparison?

GuyHarwood:
@sinhaamit1980 @activemq  test cases...
- try it out check with no matching record in checkPin table (where cp.pinExpiresAt IS NULL)
- try it out check with checkPin.pinExpiresAt in the past

PR #1575, opened 2020-02-28
Bug/33856 Used staticify to serve static assets

GeorgeChatzigiannis:
Serve static assets with version hashes that are the MD5 of the contents of each static asset.

### DIFF
--- a/admin/helpers/healthcheck.js
+++ b/admin/helpers/healthcheck.js
@@ -3,10 +3,16 @@
 const fs = require('fs')
 const path = require('path')
 
+let cachedCommitId
+let cachedCommitIdExpiresAt
+let cachedBuildNumber
+let cachedBuildNumberExpiresAt
+
 module.exports = {
-  getCommitId: () => {
-    return new Promise(function (resolve, reject) {
-      var commitFilePath = path.join(__dirname, '..', 'public', 'commit.txt')
+  getCommitId: async (cacheBust = false) => {
+    const now = Date.now()
+    const readCommitIdFromFileAsync = new Promise(function (resolve, reject) {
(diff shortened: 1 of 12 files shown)


admin/helpers/index.js
```diff
@@ -58,7 +57,7 @@ module.exports = async function (app) {
   app.locals.crownCopyrightMessage = null
   app.locals.googleTrackingId = config.GOOGLE_TRACKING_ID
   app.locals.appInsightsClientKey = config.Monitoring.ApplicationInsights.Key
-  app.locals.deployVersion = appVersion
+  app.locals.deployVersion = buildNumber
```
GeorgeChatzigiannis: hmm do we need both locals if they have the same value? 🤔

GuyHarwood: i think buildNumber is a vso generated item, which we inject into build.txt during `dev build`
this, in my humble opinion, is a more useful value as it maps to a build, and therefore a commit

admin/helpers/healthcheck.js
```diff
@@ -15,11 +21,16 @@ module.exports = {
         }
       })
     })
+    if (cacheBust || !cachedCommitId || !cachedCommitIdExpiresAt || now > cachedCommitIdExpiresAt) {
+      cachedCommitId = await readCommitIdFromFileAsync
+      cachedCommitIdExpiresAt = Date.now() + (60 * 1000) // +60 seconds
```
jon-shipley: How does the commit ever change?  Sure it can only change on new deployment? If so, why bother with cache-busting?  Perhaps I am missing something?

GeorgeChatzigiannis:
Dropping this since this it failed for remote assets.

PR #1579, opened 2020-03-03
Feature/33856 Generated md5 assets hash and use it for cache busting

GeorgeChatzigiannis:
- Generate md5 hash from both js and css contents
- Use the hash for cache busting


admin/package.json
```diff
@@ -150,5 +156,6 @@
       "<rootDir>/services/data-access"
     ],
     "verbose": false
-  }
-}
+  },
+  "mtc[assets-version]": "2c8bde8dabf32d6b9b64059875461c22"
```
GuyHarwood: why the duplication?

GeorgeChatzigiannis: thanks for spotting this, it was a leftover from a test I made.
it's removed now.

admin/gulpfile.js
```diff
@@ -154,12 +157,37 @@ gulp.task('realclean', gulp.series('clean'), function () {
...
+  dir.readFiles('./public/', {
+    // match only filenames with a .js and .css extensions and that don't start with a `.´
+    match: /\.(js|css)(?:\?.*|)$/i,
+    exclude: /^\./
+  }, function (err, content, next) {
+    if (err) throw err
+    assetBuffer = assetBuffer + content
+    next()
+  },
+  function (err, files) {
```
jon-shipley: `files` doesn't appear to be used.

admin/gulpfile.js
```diff
@@ -154,12 +157,37 @@ gulp.task('realclean', gulp.series('clean'), function () {
     .pipe(clean())
 })
 
+gulp.task('generate-assets-version', function (done) {
+  let assetBuffer = ''
+  dir.readFiles('./public/', {
+    // match only filenames with a .js and .css extensions and that don't start with a `.´
+    match: /\.(js|css)(?:\?.*|)$/i,
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: @jon-shipley 🏅

admin/gulpfile.js
```diff
@@ -154,12 +157,37 @@ gulp.task('realclean', gulp.series('clean'), function () {
...
+    exclude: /^\./
+  }, function (err, content, next) {
+    if (err) throw err
+    assetBuffer = assetBuffer + content
```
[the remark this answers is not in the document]
GeorgeChatzigiannis: I forgot to rename the var as it is no longer a buffer while using this library. Thanks!

PR #1581, opened 2020-03-05
[CONFIG-CHANGE] Feature/37087 persist answers after marking

jon-shipley:
# Persist answer after marking

* Move the marks awarded to a new table `checkResult` that is written to if marking is successful
* PK is the checkCode
* RowKey is random uuid
* Marking errors _still_ cause the `receivedCheck` to be updated
* Once marking is complete a `checkComplete` message is sent out
* Uses functionBindings to write the marking data to the `checkResult` table

## Config change
      "connection": "ServiceBusConnection"
Is now using 
```
"connection": "AZURE_SERVICE_BUS_CONNECTION_STRING"
```

For local dev change local-settings.json
For azure environments ensure AZURE_SERVICE_BUS_CONNECTION_STRING is set


tslib/src/functions/check-marker/check-marker.v1.ts
```diff
@@ -107,39 +113,67 @@ export class CheckMarkerV1 {
...
+      answers: sortedAnswers,
       formQuestions: checkForm,
       results: []
     }
     return toReturn
   }
 
-  private markCheck (markingData: MarkingData): Mark {
-    const results: Mark = {
+  private markCheck (markingData: MarkingData, checkCode: string): CheckResult {
```
GeorgeChatzigiannis: is there a chance at that stage that markingData don't have the appropriate object structure?

jon-shipley: @GeorgeChatzigiannis - is there a specific path that would cause that?

tslib/src/functions/check-marker/check-marker.v1.ts
```diff
@@ -28,17 +28,19 @@ export class CheckMarkerV1 {
   }
 
   async mark (functionBindings: ICheckMarkerFunctionBindings, logger: ILogger): Promise<void> {
-
+    logger.verbose('mark() called')
     const validatedCheck = this.findValidatedCheck(functionBindings.receivedCheckTable)
     const markingData = await this.validateData(functionBindings, validatedCheck, logger)
```
GeorgeChatzigiannis: This method can return a different response other than undefined or the expected structure of markingData.
The implementation returns the result of `updateReceivedCheckWithMarkingError` in many cases which seems to be a problem. This will need to be addressed.

GuyHarwood: @GeorgeChatzigiannis you are missing the context of working in Typescript - the strict typing forces you to declare all possible return types (using a pipe) and you cannot write code to check for undefined if you have not specified it as one of those return types.

GeorgeChatzigiannis: @GuyHarwood I can see `updateReceivedCheckWithMarkingError` can return the result of `tableService.replaceEntityAsync` which is being referenced in the `IAsyncTableService` interface.
In there the return type is described as `Promise<any>`.
Am I missing something else in here? 🤔

jon-shipley: The Promise<any> returns type was masking the issue.  I've updated the types on storage helper.

PR #1582, opened 2020-03-05
Bug/38477 Unselect pupils that do not belong to the group when filtering by a group and fixed warnings

GeorgeChatzigiannis:


admin/views/pupil-pin/generate-pins-list.ejs
```diff
@@ -48,7 +48,7 @@
                                         Date of birth: <%= pupil.dateOfBirth %>
                                     </div>
                                 <% } %>
-                                <input type="hidden" id="pupilUpn" name="pupilUpn" value="<%= pupil.upn %>"/>
+                                <input type="hidden" id="pupilUpn-<%= pupil.upn %>" name="pupilUpn" value="<%= pupil.upn %>"/>
```
jon-shipley: Would be better to use pupil.urlSlug

admin/assets/javascripts/util-checkbox.js
```diff
@@ -152,7 +152,9 @@ var inputStatus = {
     * @param checkboxParent
     */
   countCheckedCheckboxes: function (checkboxParent) {
-    var el = $((checkboxParent || '.multiple-choice-mtc') + ' > input:checkbox:checked').not('#tickAllCheckboxes')
```
jon-shipley: @GeorgeChatzigiannis 
all that's needed here is an extra attribute filter to only count visible checked checkbox inputs.

 `var el = $((checkboxParent || '.multiple-choice-mtc') + ' > input:checkbox:checked:visible').not('#tickAllCheckboxes')`
 
Then you don't need to do the adding and removing classes dance - it's cleaner.

admin/assets/javascripts/util-checkbox.js
```diff
@@ -279,12 +281,23 @@ var checkboxUtil = {
...
+    const checkBoxEls = document.querySelectorAll(sel.concat(' > td > .multiple-choice-mtc'))
     if (paramIds.length < 1 || paramIds[0].length < 1) {
       $(sel).removeClass('filter-hidden-group')
+      for (let i = 0; i < checkBoxEls.length; i++) {
+        checkBoxEls[i].classList.remove('filter-hidden-group')
+      }
     } else {
       $(sel).addClass('filter-hidden-group')
+      for (let i = 0; i < checkBoxEls.length; i++) {
+        checkBoxEls[i].classList.add('filter-hidden-group')
```
jon-shipley: If this class `filter-hidden-group` is only applied so that javascript can use it, then it should be prefixed with `js-`

GeorgeChatzigiannis: added

admin/assets/javascripts/util-checkbox.js
```diff
@@ -278,18 +278,33 @@ var checkboxUtil = {
...
     */
   tableRowVisibility: function (param, paramIds) {
-    var sel = document.getElementsByClassName('govuk-template').length === 1 ? '.govuk-spacious > tbody > tr' : '.spacious > tbody > tr'
+    const sel = '.govuk-spacious > tbody > tr'
     if (paramIds.length < 1 || paramIds[0].length < 1) {
-      $(sel).removeClass('filter-hidden-group')
+      $(sel).removeClass('js-filter-hidden-group')
     } else {
-      $(sel).addClass('filter-hidden-group')
+      $(sel).addClass('js-filter-hidden-group')
```
jon-shipley: If we are hiding the checkbox, then should we just uncheck it too?  This seems obvious and simple to me.  Why not?

GeorgeChatzigiannis: Moved the hiding checkboxes function within that method.
The one you are commenting has scoped table rows not checkboxes that will be hidden if they don't belong in a group.

PR #1591, opened 2020-03-16
Bug/38524 Session expiry changes

GeorgeChatzigiannis:
Replace page reload with hiding and redirect to sign out with expired text on session expiry box


admin/config.js
```diff
@@ -34,8 +34,8 @@ const getLinesPerCheck = () => {
 }
 
 module.exports = {
-  ADMIN_SESSION_DISPLAY_NOTICE_AFTER: (process.env.ADMIN_SESSION_DISPLAY_NOTICE_AFTER || 5) * 60, // Display notice after 5 minutes by default
-  ADMIN_SESSION_EXPIRATION_TIME_IN_SECONDS: (process.env.ADMIN_SESSION_EXPIRATION_TIME_IN_SECONDS || 10) * 60, // Expire after 10 minutes inactivity by default
+  ADMIN_SESSION_DISPLAY_NOTICE_AFTER: parseInt(process.env.ADMIN_SESSION_DISPLAY_NOTICE_AFTER) || 300, // Display notice after 5 minutes by default
+  ADMIN_SESSION_EXPIRATION_TIME_IN_SECONDS: parseInt(process.env.ADMIN_SESSION_EXPIRATION_TIME_IN_SECONDS) || 600, // Expire after 10 minutes inactivity by default
```
GuyHarwood: just to confirm - we are showing the inactivity warning message after 5 minutes, and the express session auto expires after 10?

GeorgeChatzigiannis: Based on the default fallback values yes. @GuyHarwood not sure what values we had before for the env variables in different environments but I am guessing we were passing down minute values( like 5 or 10) but now we need to pass down the equivalent in seconds (aka 5*60=300).

GuyHarwood: i will review what is set in the environments, but i believe we just use defaults.  thanks for info

PR #1598, opened 2020-03-18
Add createdby_userId during check creation

GeorgeChatzigiannis:
In order to add the new column `createdby_userId` and populate it as part of the check creation process the type needs to be dropped and the `spCreateChecks` procedure recreated.

### DIFF
--- a/admin/data/sql/migrations/20200318120158.do.update-check-add-created-by-user-id.sql
+++ b/admin/data/sql/migrations/20200318120158.do.update-check-add-created-by-user-id.sql
@@ -1,0 +1,4 @@
+ALTER TABLE [mtc_admin].[check]
+ADD createdBy_userId INT NOT NULL
+CONSTRAINT [FK_check_createdBy_userId_fk] FOREIGN KEY([createdBy_userId])
+REFERENCES [mtc_admin].[user] ([id])
--- a/admin/data/sql/migrations/20200318120158.do.update-check-add-created-by-user-id.sql
+++ b/admin/data/sql/migrations/20200318120158.do.update-check-add-created-by-user-id.sql
@@ -1,4 +1,4 @@
 ALTER TABLE [mtc_admin].[check]
-ADD createdBy_userId INT NULL
+ADD createdBy_userId INT NULL;
 CONSTRAINT [FK_check_createdBy_userId] FOREIGN KEY([createdBy_userId])
-REFERENCES [mtc_admin].[user] ([id])
+REFERENCES [mtc_admin].[user] ([id]);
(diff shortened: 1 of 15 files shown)


admin/services/check-start.service/check-start.service.js
```diff
@@ -194,6 +201,7 @@ checkStartService.initialisePupilCheck = async function (
   availableForms,
   usedFormIds,
   isLiveCheck,
+  userId,
   schoolId = null,
```
GeorgeChatzigiannis: Not relevant to my PR but this seems to be wrong. `schoolId` will definitely not be null at that stage (the caller method already evaluates that) so providing a default to null seems to be totally wrong here.

GuyHarwood: please check existing call sites and remove default if not required

admin/data/sql/migrations/20200318120158.do.update-check-add-created-by-user-id.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE [mtc_admin].[check]
+ADD createdBy_userId INT NULL
+CONSTRAINT [FK_check_createdBy_userId] FOREIGN KEY([createdBy_userId])
+REFERENCES [mtc_admin].[user] ([id])
```
jon-shipley: Missing semi-colon

admin/data/sql/migrations/20200318120158.undo.update-check-add-created-by-user-id.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE [mtc_admin].[check]
+DROP CONSTRAINT [FK_check_createdBy_userId];
+ALTER TABLE [mtc_admin].[check]
```
jon-shipley: Is this okay like this?  Or should the two ALTERs go in separate files?

admin/data/sql/migrations/20200318133108.do.drop-spcreatechecks.sql
```diff
@@ -0,0 +1 @@
+DROP PROCEDURE IF EXISTS [mtc_admin].[spCreateChecks];
```
jon-shipley: I'm surprised this is needed - you are using the CREATE OR ALTER statement for the new definition.

GuyHarwood: i would leave it in, as it makes the operation fail safe.  although there is the argument 'surely it should fail if it's not there as there would be a wider issue?'

admin/data/sql/migrations/20200318120158.undo.update-check-add-created-by-user-id.sql
GeorgeChatzigiannis: Looks ok to me. they can happen in a single transaction.

admin/data/sql/migrations/20200318133108.do.drop-spcreatechecks.sql
jon-shipley: It won't fail if it already exists or if it doesn't exist - it's using CREATE OR ALTER PROCEDURE

PR #1601, opened 2020-03-19
Feature/39967 access arrangements not persisting when changed in spa affecting azure only

jon-shipley:
Restore data idempotently

TODO

 - [x] Rewrite the query so that it produces an error if an error occurs
 - [x] Pass the context.log into the sqlService 
 - [x] Mod the console.log lines to use this.logger
 - [x] Fix bug in SPA tests
 - [x] Fix integration tests


admin/services/data-access/pupil-restart.data.service.js
```diff
@@ -162,7 +162,7 @@ pupilRestartDataService.sqlMarkRestartAsDeleted = async (restartId, userId) => {
 
                    -- IF the new check is not in NEW status we have to bail out
                    IF @newCheckCode <> 'NEW'
-                    THROW 5100, 'New check cannot be deleted as it does not have a NEW status', 1;
+                    THROW 51000, 'New check cannot be deleted as it does not have a NEW status', 1;
```
GuyHarwood: have you tested this on azure, as sp_addmessage does not exist and you should verify the behaviour of custom message codes on sql azure.  should 'just work' or you may have to use 50000.  not ideal, but thats how they do it

jon-shipley: This fixes an issue as the error code needs to >= 51000. Looks like a 0 got missed off.

GuyHarwood:
checks passed but not reporting back to github 🤷‍♂

PR #1603, opened 2020-03-20
Feature/39705 Add new user fields in pupil access arrangements

GeorgeChatzigiannis:
Renamed field to `createdBy_userId ` and added `modifiedBy_userId `


GeorgeChatzigiannis:
Existing implementation is preventing on reporting the user who modified access arrangements

PR #1604, opened 2020-03-20
Feature/40047 Accessiblity statement update

GeorgeChatzigiannis:


admin/views/accessibility-statement.ejs
```diff
@@ -44,21 +39,21 @@
...
                     <th>Access arrangement</th>
-                    <th>Recommended browser/device</th>
+                    <th>Recommended browser and device</th>
                 </tr>
                 </thead>
                 <tbody>
                 <tr>
                     <td>Audible time alert</td>
-                    <td>Not compatible with IE11</td>
+                    <td>Any browser/device</td>
```
jon-shipley: Technical compatibility.  @GeorgeChatzigiannis - can you confirm this advice is correct?

GeorgeChatzigiannis: @jon-shipley updated accordingly, thanks for spotting this.

PR #1609, opened 2020-03-26
Bug/40163 HDF page inacessible

GeorgeChatzigiannis:
Fixed both queries by identifying if the check is complete


activemq:
> LGTM
> @activemq we need to think about how we could catch this in the test suite 🤔

@GuyHarwood will see if i can add a test to capture this now

PR #1611, opened 2020-03-27
Bug/40087 Results page not display HDF unavailability

GeorgeChatzigiannis:
Attempt to access and evaluate redis result after checking results page accessibility and hdf submission.


admin/controllers/results.js
```diff
@@ -59,6 +57,13 @@ controller.getViewResultsPage = async (req, res, next) => {
...
   }
 
+  try {
+    redisResult = await resultService.getPupilResultData(req.user.schoolId)
+    pupilResultData = redisResult && redisResult.pupilResultData
+  } catch (error) {
+    return next(error)
+  }
+
   if (!redisResult) {
```
GuyHarwood: should this check pupilResultData?

GeorgeChatzigiannis: Good spot @GuyHarwood.

PR #1612, opened 2020-03-30
Feature/39556 check view support api

GuyHarwood:
developer check view api.


GuyHarwood:
work parked until helpdesk pin usage page is complete

PR #1618, opened 2020-04-16
Feature/40535 dfe number separator

GuyHarwood:
permits the inclusion of a hyphen or forward-slash at index 3 in the dfe number when performing impersonation.

- fix existing implementation, as it was invalid
- trim input
- cleanse input if separator character detected


admin/services/school-impersonation.service.js
```diff
@@ -13,6 +13,13 @@ const schoolImpersonationService = {}
  * @returns {Object}
  */
 schoolImpersonationService.setSchoolImpersonation = async (user, dfeNumber) => {
+  if (typeof dfeNumber === 'string') {
+    dfeNumber = dfeNumber.trim()
+    if (dfeNumber[3] === '-' || dfeNumber[3] === '/') {
+      dfeNumber = dfeNumber.slice(0, 3) + dfeNumber.slice(4)
```
jon-shipley, line 19: This is pretty precise, which may be what you want.   But as dfeNumber is an int we can simply remove all non-digit input and see what we are left with:

```js
const regex = /[^0-9]/g
const p1 = '123-4567'
const p2 = '123/4567'
const p3 = ' 1234567 '

> p1.replace(regex, '')
'1234567'
> p2.replace(regex, '')
'1234567'
> p3.replace(regex, '')
'1234567'
```

This obviously also removes the need for calling `trim()`.

activemq:
@GuyHarwood @jon-shipley This has been signed off, however it is not ready to merge as it has a bunch of unrelated fails when i ran the whole suite. These have now been resolved and i am re running the suite.

GuyHarwood:
@activemq unit tests now passing, ready for sign-off with business 👍

PR #1639, opened 2020-05-05
resolve issues with connection timeouts etc

activemq:
- [x] Fixed issues with browser-stack time outs

- [x] better reporting in the dashboard

- [x] fix issues with some methods not working when running in browserstack mode

- [x] clean up yml files


test/admin-hpa/Gemfile
```diff
@@ -10,7 +10,7 @@ gem 'rb-readline'
 gem 'pry'
 gem 'require_all'
 gem 'rspec'
-gem 'selenium-webdriver'
+gem 'selenium-webdriver', '4.0.0.alpha1'
```
jon-shipley, line 13: Alpha?

activemq: @jon-shipley -  its the only version that works with it atm.

jon-shipley: Okay.  Let's keep an eye on it then and upgrade when it's stable 😓

activemq: @jon-shipley - For sure, using this version at least gets the tests up and running. Agree to keep an eye on it

PR #1644, opened 2020-05-15
Feature/39556 tech support check summary

GuyHarwood:
Tech support check view summary page.

- [x] implement role based connection pools
- [x] refactor sql service to 'proper' object
- [x] implement check summary page (WIP)
- [x] smarten up main menu
- [x] show 'not found' flash message if no check found
- [x] unzip receivedCheck and render, take logic from test-dev page


admin/services/data-access/sql.role-connection.builder.js
```diff
@@ -0,0 +1,36 @@
...
+   * @description builds a sql connection config object
+   * @param {string} roleName - the role of the user to create a connection config for
+   * @param {boolean} readonly - set to true for a readonly connection, typically to a sql azure replica
+   * @returns {object} a config object that works with mssql
+   */
+  build: function build (roleName, readonly = false) {
+    let cfg
+    switch (roleName) {
+      case roles.teacher:
+        cfg = sqlConfig
```
jon-shipley, line 19: This is assigning a reference to sqlConfig, and there is a line doing mutation later on.

Suggest we freeze sqlConfig once it is read, and return a cloned copy here.
`cfg = R.clone(sqlConfig)`

admin/services/data-access/sql.role-connection.builder.js
```diff
@@ -0,0 +1,36 @@
...
+        break
+      case roles.techSupport:
+        cfg = R.clone(sqlConfig)
```
jon-shipley, line 22: This is clear code, but as you are using Ramda anyway we could do the clone and the assignment in a single call, and then freeze the result, perhaps something like this:

```
const cfg = Object.freeze(R.mergeDeepRight(sqlConfig, {
  user: 'foo',
  password: 'bar',
  pool: {
    min: 0,
    max: 100
  }
}))
```

https://ramdajs.com/docs/#mergeDeepRight

admin/services/data-access/sql.role-connection.builder.js
GuyHarwood: great idea, a lot safer as it could be overlooked in further down the line

admin/data/sql/migrations/20200601114033.do.alter-vewcheckdiagnostic.sql
```diff
@@ -0,0 +1,28 @@
+ALTER VIEW [mtc_admin].[vewCheckDiagnostic]
```
jon-shipley, line 1: Thank you 😊

admin/controllers/tech-support.js
```diff
@@ -0,0 +1,102 @@
...
+    const { checkCode } = req.body
+    try {
+      const validationError = uuidValidator.validate(checkCode, 'checkCode')
+      if (validationError && validationError.hasError && validationError.hasError()) {
+        return controller.getCheckViewPage(req, res, next, validationError)
+      }
+      let notFound = false
+      const checkSummary = await checkDiagnosticsService.getByCheckCode(checkCode)
+      if (!checkSummary) {
+        notFound = true
```
jon-shipley, line 65: I find negative variables harder to read than the positive equivalents: e.g. `found:  false`

admin/controllers/tech-support.js
```diff
@@ -0,0 +1,102 @@
...
+   */
+  getReceivedCheckPayload: async function getReceivedCheckPayload (req, res, next) {
+    const checkCode = req.query.checkCode.trim()
+    let response
+    try {
+      const payload = await payloadService.getPayload(checkCode)
+      response = payload
+    } catch (error) {
+      next(error)
+      // response = { message: 'check not found' }
```
jon-shipley, line 95: Can this be deleted now?

admin/spec/back-end/controllers/tech-support.spec.js
```diff
@@ -0,0 +1,107 @@
...
+    })
+
+    it('POST: should redirect back to GET when validation fails', async () => {
+      const req = getRequest(getReqParams('/tech-support/checkview', 'POST'))
+      const res = getResponse()
+      spyOn(res, 'render')
+      spyOn(sut, 'getCheckViewPage')
+      await sut.postCheckViewPage(req, res, next)
+      expect(next).not.toHaveBeenCalled()
+      expect(res.statusCode).toBe(200)
```
jon-shipley, line 91: This status code does not agree with the test title.  Redirect is mentioned in the title,  but actually, if validation fails the controller issues a  200 response.

admin/spec/back-end/controllers/tech-support.spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+  describe('/received-check-payload', () => {
+    it('GET: should render payload', async () => {
+      const req = getRequest(getReqParams('/tech-support/received-check-payload', 'GET'))
+      const res = getResponse()
+      req.query.checkCode = checkCode
+      spyOn(payloadService, 'getPayload')
+      await sut.getReceivedCheckPayload(req, res, next)
+      expect(next).not.toHaveBeenCalled()
+      expect(payloadService.getPayload).toHaveBeenCalledWith(checkCode)
```
jon-shipley, line 104: Should this test that `render` was called as per the title?

GuyHarwood:
@jon-shipley thanks Jon, all remediated

GuyHarwood:
@activemq had to override checks as it was caching the 'validation containers & migrations' after i had renamed it to 'validation migrations'.  the latter passed.

PR #1652, opened 2020-06-01
feature/41755 switch to github api header token

GuyHarwood:
Switch github API call to header token now that query params are deprecated


GuyHarwood:
travis link appears to be hung. will create new PR

PR #1658, opened 2020-06-03
update required ruby version

activemq:
- [x] upgrade ruby version across project to 2.6.0

- [x] add required version using bundler ruby version to enforce version 2.6.0

### DIFF
--- a/api_tests/Gemfile.lock
+++ b/api_tests/Gemfile.lock
@@ -36,5 +36,8 @@ DEPENDENCIES
   pry
   rspec
 
+RUBY VERSION
+   ruby 2.6.0p0
+
 BUNDLED WITH
-   1.16.6
+   1.17.2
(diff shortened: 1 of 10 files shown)


api_tests/Gemfile.lock
```diff
@@ -36,5 +36,8 @@ DEPENDENCIES
   pry
   rspec
 
+RUBY VERSION
```
GuyHarwood, line 39: did you change this yourself, or auto updated?

activemq: @GuyHarwood this is added automatically as now i have added a ruby version to the Gemfile. Then when we run bundle install, it fails if any other version of ruby is used with the following error message:
"Your Ruby version is 2.4.0, but your Gemfile specified 2.6.0"

activemq:
Please hold off merging as we need to setup ruby version 2.6.0 on the build servers before merging.

api_tests/Gemfile.lock
GuyHarwood: Good 👍

PR #1659, opened 2020-06-04
feature/41756 - fix tsc issues for admin app

GuyHarwood:


.circleci/config.yml
```diff
@@ -24,10 +24,13 @@ jobs:
           key: v1-node-tslib-deps-{{ checksum "tslib/yarn.lock" }}
       - run: cd admin; yarn install --frozen-lockfile
       - run:
-          name: Install JUnit coverage reporter
+          name: Admin lint and ts compile
```
jon-shipley, line 27: I don't see any compilation steps in `yarn lint`

.circleci/config.yml
```diff
@@ -24,10 +24,13 @@ jobs:
...
+          command: cd admin; yarn lint; rm -rf dist
+      - run:
+          name: Install JUnit coverage reporter for admin unit tests
           command: cd admin; yarn add --dev jest-junit
       - run:
-          name: Run tests with JUnit as reporter
+          name: Admin unit tests with JUnit as reporter
           command: cd admin; ./node_modules/.bin/jest --ci --runInBand --reporters=default --reporters=jest-junit --collectCoverage=true
```
jon-shipley, line 34: As far as I know, CircleCI is not surfacing any code coverage details - so why bother to collect code coverage?

admin/helpers/pupil-presenter.js
```diff
@@ -10,7 +10,8 @@ const pupilPresenter = {}
  */
 pupilPresenter.getPupilExampleYear = () => {
   const deductionYears = 8
-  return moment.utc().format('YYYY') - deductionYears
+  const currentYear = parseInt(moment.utc().format('YYYY'), 10)
```
jon-shipley, line 13: Possibly simpler, as it eliminates the need to convert from string to number:

```suggestion
  const currentYear = (new Date()).getFullYear() // returns number
```

admin/lib/validator/check-window-v2/check-window-add-validator.js
```diff
@@ -204,68 +204,88 @@ checkWindowAddValidator.validate = (checkWindowData, validationConfig = null) =>
   // Compare date fields
   // Admin start date
   if (adminStartDate && adminStartDate.isAfter(adminEndDate, 'days')) {
+    // @ts-ignore
```
jon-shipley, line 207: Why are we applying `ts-ignore` to our own code?

admin/lib/validator/pin-validator.js
```diff
@@ -5,7 +5,7 @@ const pinValidator = {}
 /**
  * Validate pin
  * @param {string} pin
- * @param {Moment} pinExpiresAt
+ * @param {import('moment').Moment} pinExpiresAt
```
jon-shipley, line 8:
```suggestion
 * @param {moment.Moment} pinExpiresAt
```

admin/lib/validator/pupil-validator.js
```diff
@@ -153,6 +159,7 @@ module.exports.validate = async (pupilData, schoolId, isMultiplePupilsSubmission
     }
   }
   if (upnErrorArr.length > 0) {
+    // @ts-ignore
```
jon-shipley, line 162: This shouldn't need to be ignored.

admin/package.json
```diff
@@ -4,7 +4,7 @@
   "private": false,
   "license": "GPL-3.0",
   "scripts": {
-    "lint": "standard",
+    "lint": "standard && tsc",
```
jon-shipley, line 7: Why?  Now it does two things...

admin/services/business-availability.service.js
```diff
@@ -105,14 +106,33 @@ businessAvailabilityService.determineAccessArrangementsEligibility = (checkWindo
...
+ * @property {boolean} adminWindowStarted,
+ * @property {boolean} adminWindowClosed,
+ * @property {boolean} hdfAvailable,
+ * @property {boolean} hdfSubmitted,
+ * @property {boolean} canEditArrangements,
+ * @property {boolean} restartsAvailable,
+ * @property {boolean} livePinsAvailable,
+ * @property {boolean} familiarisationPinsAvailable,
+ * @property {boolean} groupsAvailable,
+ * @property {boolean} accessArrangementsAvailable
```
jon-shipley, line 125: Thank you 👏  but please can you also sort it?

admin/services/config.service.js
```diff
@@ -39,10 +39,37 @@ const configService = {
     return R.split(',', commaSeparatedString)
   },
 
+  /**
+   * @typedef {object} RawPupilConfigData
```
jon-shipley, line 43: 💯

admin/services/config.service.js
```diff
@@ -39,10 +39,37 @@ const configService = {
...
+   * @property {number} loadingTime
+   * @property {number} questionTime
+   * @property {number} checkTime
+   * @property {any} accessArrangementCodes
+   * @property {any} fontSizeCode
+   * @property {any} colourContrastCode
+   */
+
+  /**
+  * @typedef {object} StructuredPupilConfig
```
jon-shipley, line 55: 💯 - thank you!

admin/services/data-access/pin-generation.data.service.js
```diff
@@ -39,11 +39,22 @@ const serviceToExport = {
     return sqlService.readonlyQuery(sql, [param])
   },
 
+  /**
+   * @typedef CreateCheck
+   * @property {number} pupil_id
+   * @property {number} checkForm_id
+   * @property {number} checkWindow_id
+   * @property {boolean} isLiveCheck
+   * @property {import('moment').Moment} pinExpiresAt
```
jon-shipley, line 48: Is it meant to be like this?

admin/services/prepare-check.service.js
```diff
@@ -14,10 +14,24 @@ const redisService = require('./data-access/redis-cache.service')
 const startOfDay = () => moment().startOf('day')
 const eightAmToday = () => moment().startOf('day').add(8, 'hours')
 
+/**
+ * Unprepared Check
+ * @typedef {Object} UnpreparedCheck
```
jon-shipley, line 19: Please sort alphabetically

admin/services/pupil-status.service.js
```diff
@@ -54,12 +54,30 @@ const pupilStatusService = {
     return newPupil
   },
 
+  /**
+ * Process Status Object
+ * @typedef {Object} ProcessStatus
```
jon-shipley, line 59: Please consider sorting these props

.circleci/config.yml
GuyHarwood: `yarn lint` now does `standard && tsc`, see the `admin/package.json` edits

jon-shipley: I saw this later on in the PR... so now the name should be lintAndBuild?

.circleci/config.yml
GuyHarwood: We get this...
Screenshot 2020-06-09 10 51 10

Ideally i'd like a gate on the overall coverage metric, which i think would be suitable for codacy?

admin/lib/validator/check-window-v2/check-window-add-validator.js
GuyHarwood: some of the code would require a large refactor to satisfy the typescript compiler.  The Ramda uses are typical of this.
We can certainly consider rewriting them, but i feel its outside the scope of this ticket.
at the very least we now have placeholders for code to address.
I would definitely be interested in having a deep dive with you on some of these, so we can understand the effort and direction we want to take

admin/lib/validator/pupil-validator.js
GuyHarwood: its an array, being passed to a method that expects a string.  From what i recall, the code appeared to be lazy in the sense that the author probably knew there would only ever be one element in `upnErrorArr`.  `validationError.addError` has no concept of dealing with arrays.

.circleci/config.yml
GuyHarwood: i left as-is as we are really only using `tsc` for code quality right now.  However, i think 2 distinct items would be clearer, such as `lint` and `lintAndBuild`.
let me know if you think otherwise.

admin/services/data-access/pin-generation.data.service.js
GuyHarwood: yes, but only in places where moment is not declared/required in the file itself.  There were a couple of these instances where `moment-timezone` was required at the top of the file, but `tsc` couldn't infer the `Moment` type from it.   When `moment` is imported, i simply use `{moment.Moment}`

admin/package.json
GuyHarwood: as per other comment, will separate into two distinct script calls

admin/lib/validator/check-window-v2/check-window-add-validator.js
jon-shipley: Sounds like a plan

admin/lib/validator/pupil-validator.js
jon-shipley: The calling code probably needs to be refactored then.  We can use these hints as a ToDo list...

.circleci/config.yml
jon-shipley: Fair enough then

PR #1681, opened 2020-07-08
feature/41244 retro input assistant persistence

GuyHarwood:
wiring up of the retrospective access arrangement UI.

- validation of form inputs
- binding and resolution of latest check
- data persistence


admin/controllers/retro-input-assistant.js
```diff
@@ -0,0 +1,57 @@
+'use strict'
+
+const checkWindowV2Service = require('../services/check-window-v2.service')
+const businessAvailabilityService = require('../services/business-availability.service')
+const ValidationError = require('../lib/validation-error')
+const retroInputAssistantService = require('../services/retro-input-assistant.service')
+
+const controller = {}
+
+controller.getAddRetroInputAssistant = async (req, res, next, error = null) => {
```
jon-shipley, line 10: It would be better practise, in my view, if we used named functions, as at least then, should we choose to use an APM, the functions will be meaningful.

admin/data/sql/migrations/20200710143820.undo.check-retro-input-assistant.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE [mtc_admin].[check] DROP COLUMN retroInputAssistantFirstName
```
jon-shipley, line 1: These columns can be (more efficiently?) dropped in a single statement:
https://github.com/jon-shipley/patterns/blob/master/mssql-idempotent-sql.md#Columns

admin/services/access-arrangements.service.js
```diff
@@ -13,7 +13,12 @@ const accessArrangementsService = {}
  * @returns {Promise<Array>}
  */
 accessArrangementsService.getAccessArrangements = async () => {
-  return accessArrangementsDataService.sqlFindAccessArrangements()
+  let aaData = await accessArrangementsDataService.sqlFindAccessArrangements()
+  const retroInputAssistantItemIndex = R.findIndex(R.propEq('code', 'RIA'))(aaData)
+  if (retroInputAssistantItemIndex > -1) {
+    aaData = R.remove(retroInputAssistantItemIndex, 1, aaData)
```
jon-shipley, line 19: Just a thought, if RIA was used the previous time, won't it be needed the next time, except hopefully not retrospectively?  Shouldn't RIA be converted to ITA?

admin/services/data-access/retro-input-assistant.data.service.js
```diff
@@ -0,0 +1,80 @@
...
+            p.group_id,
+            p.urlSlug
+          FROM
+              [mtc_admin].[pupil] p
+              LEFT JOIN [mtc_admin].[pupilAccessArrangements] paa ON (paa.pupil_id = p.id)
+         WHERE p.school_id = @schoolId
+           AND p.attendanceId IS NULL
+           AND p.checkComplete = 1
+           AND paa.pupil_id IS NULL
+         ORDER BY lastName;
```
jon-shipley, line 74: Why do the ordering on the db when there is only a single instance of it?  Wouldn't it make more sense to order it on the web tired, which is designed for scaling?

Also, if the idea is to order it so as it pass it into the pupil identification service, this ordering is insufficient. From memory, it would need to be `surname`, `firstname`, `dateOfBirth`, `middleNames`.

admin/spec/back-end/lib/validator/retro-input-assistant-validator.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  })
+
+  it('should invalidate if userId is less than 1', async () => {
+    retroInputAssistantData.userId = 0
+    const validationResult = await sut.validate(retroInputAssistantData)
+    expect(validationResult.hasError()).toBe(true)
+    const checkIdError = validationResult.get('userId')
+    expect(checkIdError).toEqual(errorMessages.userId)
+  })
+})
```
jon-shipley, line 108: nice 👍

admin/services/access-arrangements.service.js
GuyHarwood: this is a distinct access arrangement, and is removed from here so it doesn't end up on the main AA page as an option for selection.  it is only ever added from the dedicated retro input assistant page.  The list above is used in the main AA page

admin/services/data-access/retro-input-assistant.data.service.js
GuyHarwood: this is a mirror image of the function used in the main AA page autocomplete feature, except for the filtering (which only includes pupils with completed checks).

the pupil uuid is captured for the autocomplete, to be submitted with the form.

if both instances of this fall short then we should raise as a separate concern.

PR #1682, opened 2020-07-09
Feature/41488 upgrade admin app node to v12

jon-shipley:
1. Upgrade admin app and pupil-api to node v12
2. Dockerise everything except the ruby tests in order to pin the dependencies

Nothing much has changed: use ./start.sh ./stop.sh and ./restart.sh to bring the env up and down as required.  This will clean up Azure and bring docker-compose up and down.

NOTE: running docker on macOS is known to utilise more cpu than native apps.


admin/dev.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+	    && echo "$SSH_PASSWD" | chpasswd
+
+COPY sshd_config /etc/ssh/
+
+EXPOSE 3001 2222
+
+RUN mkdir -p /mtc/admin
+WORKDIR /mtc/admin
+
+CMD ./docker-start.sh
```
GuyHarwood, line 16: should this be `docker-start-dev.sh`?

admin/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
...
+        WAITFORIT_RESULT=$?
+    fi
+fi
+
+if [[ $WAITFORIT_CLI != "" ]]; then
+    if [[ $WAITFORIT_RESULT -ne 0 && $WAITFORIT_STRICT -eq 1 ]]; then
+        echoerr "$WAITFORIT_cmdname: strict mode, refusing to execute subprocess"
+        exit $WAITFORIT_RESULT
+    fi
+    exec "${WAITFORIT_CLI[@]}"
```
GuyHarwood, line 179: i added this script under the deploy folder many moons ago.  It didn't end up being used, could you please remove it so we don't have duplication.

func-consumption/Dockerfile
```diff
@@ -1,12 +1,5 @@
...
-### execute ../run-docker-compose-travis.sh in order to ensure tslib dependencies are copied in prior to build
-FROM baldy/mtc-azure-functions-base:latest
-
-SHELL ["/bin/bash", "-c"]
-
-RUN mkdir -p /func-app
-WORKDIR /func-app
-COPY . .
-
+FROM mtc_functions_base:latest
```
GuyHarwood, line 1: where is it sourcing `mtc_functions_base` from? the owner prefix has been dropped...

functions-app/db-check.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
```
GuyHarwood, line 2: this has been duplicated twice now, any chance we can source from root, or am i right in assuming that's not possible due to node/yarn sourcing rules?

admin/wait-sqldb.sh
```diff
@@ -5,6 +5,6 @@ if [[ "${WAIT_FOR_SQL_SERVER:=false}" != "false" ]]; then
 
   until [[ "$(curl -sS $SQL_SERVER:$SQL_PORT 2>/dev/stdout)" =~ "(52)" ]]; do
```
jon-shipley, line 6: I don't think anything is actually using this script.

docker-compose.yml
```diff
@@ -9,8 +9,101 @@ services:
       SA_PASSWORD: "Mtc-D3v.5ql_S3rv3r"
       ACCEPT_EULA: "Y"
       HOMEBREW_NO_ENV_FILTERING: 1
+  functions_base:
```
jon-shipley, line 12: This is how mtc_functions_base image gets built.

docker-compose.yml
```diff
@@ -9,8 +9,101 @@ services:
...
+    environment:
+      SQL_SERVER: 'sqldb'
+      REDIS_HOST: 'redis'
+    depends_on:
+      - sqldb
+      - migration
+      - redis
+    command: ["./wait-for-db.sh", "sqldb", "yarn", "start"]
+  func_consumption:
+    build: ./func-consumption
```
jon-shipley, line 52: We should probably change the Dockerfiles in `func-consumption` and `functions-app` to be `dev.Dockerfile` as these are not prod assets, and the name change will make that clearer.

func-consumption/Dockerfile
jon-shipley: See the docker-compose.yml file for details - it gets built on the fly.

admin/wait-for-it.sh
GuyHarwood: (remove the one under deploy dir, not the one you added)

admin/wait-sqldb.sh
GuyHarwood: this could well have been something that was added by DW when we were exploring the ruby in docker thing a while back

functions-app/db-check.js
jon-shipley: If there is a nice solution then let me know.  

`/mtc` is mounted so the whole project is made available - however, the only dependency we need is `.env`.  Adding more dependencies into '/mtc' would slightly complicate using other methods in the future, e.g. mutagen.

However, the node script has it's own dependencies: `mssql` and our own `config.js`, so in de-duplicating a small amount of code, we would need to duplicate the larger quantity of dependencies.  

Perhaps there is a better way?  If not, let's leave it for now.

GuyHarwood: sounds like we should stick with your approach 👍

.circleci/config.yml
```diff
@@ -7,7 +7,8 @@ jobs:
   build:
     working_directory: ~/repo/
     docker:
-      - image: circleci/node:10-browsers
+#      - image: circleci/node:10-browsers
+      - image: ovix/mtc_circleci:latest
```
jon-shipley, line 11: @GuyHarwood - The modified circleci image has produced a green build.

GuyHarwood: great. can we maintain that within MTC/STA org?

PR #1683, opened 2020-07-13
changes for local id and exit code

activemq:
- [x]  Ensure browserstack returns the correct exit code
- [x] Ensure its not a breaking change to existing functionality
- [x] Start binary with local id set


test/admin-hpa/config/browserstack.yml
```diff
@@ -14,4 +14,5 @@ browser_caps:
     "os_version": "10"
     "browser": "Chrome"
     "resolution": "1024x768"
+    "browser_version": "81.0"
```
jon-shipley, line 17: Will this have to be updated as the latest comes out?

test/pupil-hpa/features/support/browserstack_driver_helper.rb
```diff
@@ -10,9 +10,11 @@
     @caps = @caps.merge ({"name" => "#{Time.now.strftime("%d/%m/%y-%H:%M")}"})
     if @caps['browserstack.local'] && @caps['browserstack.local'].to_s == 'true'
       @bs_local = BrowserStack::Local.new
-      bs_local_args = {"key" => "#{ENV['BROWSERSTACK_ACCESS_KEY']}", "force" => 'true'}
+      bs_local_args = {"key" => "#{ENV['BROWSERSTACK_ACCESS_KEY']}", "force" => 'true', 'localidentifier'=> ENV['BS_ID']}
       @bs_local.start(bs_local_args)
     end
+    fail 'Browserstack access key should be alphanumeric' if ENV['BROWSERSTACK_ACCESS_KEY'].match(/\A[a-zA-Z0-9]*\z/).nil?
+    fail 'Browserstack username should be alphanumeric' if ENV['BROWSERSTACK_USERNAME'].match(/\A[a-zA-Z0-9]*\z/).nil?
```
jon-shipley, line 17: The * here will match on zero occurrences - presumably, the length of the expected string is known.  Something like length > 8 might be better.

e.g. `/\A[a-zA-Z0-9]{8,20}\z/` to match between 8 and 20 alpha-numeric characters

test/pupil-hpa/features/support/browserstack_driver_helper.rb
```diff
@@ -10,9 +10,11 @@
     @caps = @caps.merge ({"name" => "#{Time.now.strftime("%d/%m/%y-%H:%M")}"})
     if @caps['browserstack.local'] && @caps['browserstack.local'].to_s == 'true'
       @bs_local = BrowserStack::Local.new
-      bs_local_args = {"key" => "#{ENV['BROWSERSTACK_ACCESS_KEY']}", "force" => 'true'}
+      bs_local_args = {"key" => "#{ENV['BROWSERSTACK_ACCESS_KEY']}", "force" => 'true', 'localidentifier'=> ENV['BS_ID']}
       @bs_local.start(bs_local_args)
     end
+    fail 'Browserstack access key should be alphanumeric' if ENV['BROWSERSTACK_ACCESS_KEY'].match(/\A[a-zA-Z0-9]*\z/).nil?
```
jon-shipley, line 16: The * here will match on zero occurrences - presumably, the length of the expected string is known.  Something like length > 8 might be better.

e.g. `/\A[a-zA-Z0-9]{8,20}\z/` to match between 8 and 20 alpha-numeric characters

test/admin-hpa/config/browserstack.yml
activemq: Yep, I have now removed browser version all together, it will now pick up the latest version

test/pupil-hpa/features/support/browserstack_driver_helper.rb
activemq: Added this in at the earliest point. So we get a faster fail if the regex does not match

test/pupil-hpa/features/support/browserstack_driver_helper.rb
activemq: Added this in at the earliest point. So we get a faster fail if the regex does not match

PR #1687, opened 2020-07-17
Feature/minimal docker compose by default

jon-shipley:
Dev setup change - go back to starting apps manually

The docker stuff is there - and maybe a useful option for testing.


README.md
```diff
@@ -22,42 +23,39 @@ You can now start the admin, pupil-spa & pupil-api applications individually.
...
 
 ### docker-compose.yml
-The default, which runs just SQL Server (main data store) and MongoDB (express session store)
+The default, which runs just SQL Server (main data store) and Redis (express session store)
 Stand up the database containers with: `docker-compose up`
 
-### docker-compose.admin-test.yml and docker-compose.pupil-test.yml
-Runs SQL Server, MongoDB, all web applications, and the pupil and admin test suites as separate containers.
-This compose file depends on `docker-compose.yml` and should not be run directly.
+### docker-compose-other-dev.yml
```
GuyHarwood, line 32: could we call this `docker-compose-full.yml` please, feels more specific

jon-shipley: I would argue that it is not "full" as this compose file lacks Redis and mssql.  To bring everything up requires using 2 compose files - now that is "full".

How about calling it "docker-compose-apps.yml" as one file is for data stores and one for apps.

GuyHarwood: good point, sounds accurate to me 👍

jon-shipley: Changed

PR #1717, opened 2020-09-09
Feature/42736 retro input assistant pt3

GuyHarwood:
Conclusion of the retro input assistant feature.

- retire formal 'retro input assistant' AA type
- use standard input assistant AA type
- flag existing complete check as retrospectively added


jon-shipley (approved):
LGTM

PR #1725, opened 2020-09-25
feature/40901-ie11 browserstack integration

activemq:
Exclude upload scenarios from running in IE11
Refine the Rakefile 
Kill Browserstack binary when run is completed. (This is only temporary)
Create cucumber profile for IE11


jon-shipley (approved):
Looks good to me

PR #1726, opened 2020-09-28
feature/43523 mtc_admin.hdf indexes

GuyHarwood:
- Add relevant indexes to hdf table


admin/data/sql/migrations/20200928103917.do.hdf-indexes.sql
```diff
@@ -0,0 +1,3 @@
+
+CREATE UNIQUE INDEX ix_hdf_school_checkWindow_id ON [mtc_admin].hdf (school_id, checkWindow_id);
```
jon-shipley, line 2: You could make this idempotent by either:

a) checking to see if the index exists before creating it, or
b) dropping the index if it exists, before creating the indexes required

https://github.com/jon-shipley/patterns/blob/master/mssql-idempotent-sql.md#Indexes

jon-shipley (approved):
Looks good, but could be made idempotent

PR #1730, opened 2020-10-07
43539 Store events in results schema

jon-shipley:
* Move all functions to use SQL_FUNCTIONS_APP_USER and SQL_FUNCTIONS_APP_USER_PASSWORD|
* Update the permissions of the user
* Add missing event to the eventTypeLookup table
* Update sync-results-to-sql function to now store event information too


tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    date = new Date()
+  })
+
+  it('calls `createNewEvent` if a new event is seen in the audit log when preparing events', async () => {
+    // setup
+    const eventTypes = new Map<string, DBEventType>()
+    eventTypes.set('TestEvent1', { id: 1, eventType: 'TestEvent1', eventDescription: '' })
+    eventTypes.set('TestEvent2', { id: 1, eventType: 'TestEvent2', eventDescription: '' })
+
+    jest.spyOn(sut as any, 'createNewEventType').mockImplementation(jest.fn())
```
GuyHarwood, line 32: seems odd to need to cast down to `any`. was this causing an issue?

tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    jest.spyOn(sut as any, 'refreshEventTypes')
+      .mockImplementationOnce(() => {
+        // call 1: 2 event types
+        sut['eventType'] = eventTypes
+      })
+      .mockImplementationOnce(() => {
+        // call 2 - now the new event should be added
+        eventTypes.set('TestEvent3', { id: 3, eventType: 'TestEvent3', eventDescription: '' })
+        sut['eventType'] = eventTypes
```
GuyHarwood, line 41: can we not use dot notation here?

tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    // expect
+    expect(res).toBeTruthy()
+    // we expect it to be a map
+    expect(res.size).toBe(1)
+    expect(res.has('1x1')).toBe(true)
+    expect(res.get('1x1')).toEqual({ id: 1, factor1: 1, factor2: 1, code: 'Q001', isWarmup: false })
+  })
+
+  it('sqlGetQuestionData caches data so it does not need to keep fetching the same data', async () => {
+    jest.spyOn(sut['sqlService'], 'query').mockResolvedValue([{ id: 1, factor1: 1, factor2: 1, code: 'Q001', isWarmup: false }])
```
GuyHarwood, line 94: is this to access a private member?

tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
jon-shipley: Spied methods in typescript can be typed as two types: the original object and the jest mock.  This is one method to convert a private method to jest spy.

If it was typed as a SyncResultDataService typescript would not allow it due to it being private.   Often private methods would not be covered when unit testing public interfaces, but this one has a side effect of inserting something into the db.

If there is a better way please let me know.

tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
jon-shipley: No.  It's a private variable.  Using object notation is an ugly work-around.

tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
jon-shipley: Yes.

tslib/src/functions-throttled/sync-results-to-sql/sync-results-data.service.spec.ts
jon-shipley: My understanding is that reflection in typescript is pretty poor.  Another way would be to subclass the test class and make the variable public.

It seems easier to just lean on the fact that this is transpiled to javascript.

GuyHarwood: i see why you are using this now, because it allows you to peek in and inspect the internal variables using reflection.

there are 'different' ways to do it, but that would involve pretty much rewriting what you already have, which isn't really viable for us, as we are coming from a javascript implementation to typescript, rather than writing this from the ground up.

admin/data/sql/migrations/20201006113920.do.add-event-type-lookup.sql
```diff
@@ -0,0 +1,8 @@
+IF ((SELECT id
+       FROM mtc_results.eventTypeLookup
+      WHERE eventType = 'QuestionRendered') IS NULL)
+    BEGIN
+        INSERT INTO mtc_results.eventTypeLookup (eventType, eventDescription)
+        VALUES ('QuestionRendered',
+                'The question screen was shown. For spoken questions this event is registered when the screen is showm, but the question itself is not shown until after the speech has finished.');
+    END
```
GuyHarwood, line 8: new line

GuyHarwood (approved):
LGTM

lets have a chat a little later about the different strategies for unit testing, as i think there are some good concepts that we can pull together from the different patterns in use.

admin/data/sql/migrations/20201006113920.do.add-event-type-lookup.sql
jon-shipley: thanks, added.

PR #1732, opened 2020-10-13
43908-spike-pupil-app-input-handling

jon-shipley:
Notes

1. Use Pointer Events where possible, and fall back to click handling for older devices
2. Add `touch-action: manipulation` to buttons to get iOS fast-tap


GuyHarwood (approved):
LGTM
thanks for doing this, very useful 🏆

PR #1735, opened 2020-10-27
feature/44112 switch census function to use functions sql user

GuyHarwood:
Modify the census import function to use the new functions sql user account, instead of dedicated census user, which is to be removed.

A subsequent PR will be raised to baseline the sql migrations and remove the census user entirely.


jon-shipley (approved):
LGTM

PR #1738, opened 2020-10-30
feature/44207 - fix pipeline fails

activemq:
- [x] Skip the census tests until the bug is fixed

- [x] Attempt to create a blob container if its not already created


jon-shipley (approved):
lgtm

GuyHarwood (approved):
LGTM

PR #1742, opened 2020-11-04
Feature/44247 yaml validate migrations build

GuyHarwood:
finish off yaml build definition for validate migrations.

once merged, this will replace the 'classic' vso build


jon-shipley (approved):
LGTM

PR #1743, opened 2020-11-05
Bug/44168 function log outputs connectionPool instance object

GuyHarwood:
ConnectionPool properties were being sent to stdout


jon-shipley (approved):
LGTM

PR #1744, opened 2020-11-06
bug/44173 delete blob after upload, not container

GuyHarwood:
- [admin app] Removed obsolete pupil census upload code
- [pupil census function] Refactor blob clean-up to remove file, not container


jon-shipley (approved):
LGTM

PR #1746, opened 2020-11-09
removing unused gem

activemq:
The browserstack-fast-selenium gem has a dependancy on the curb gem. We do not use browserstack-fast-selenium, it was left in and not removed after testing


GuyHarwood (approved):
LGTM

jon-shipley (approved):
lgtm

PR #1757, opened 2020-12-03
Add docker build jobs to circleci

jon-shipley:
Add docker build jobs
Fix issue with gem caching


.circleci/config.yml
```diff
@@ -89,13 +90,66 @@ jobs:
...
+    working_directory: ~/repo/
+    docker:
+      - image: cimg/base:stable
+    environment:
+      TZ: Europe/London
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.13
+      - run: cd pupil-api && BUILDKIT=1 docker build .
```
GuyHarwood, line 137: wasn't aware of the new build engine, nice 💯

GuyHarwood (approved):
LGTM

.circleci/config.yml
jon-shipley: It makes it a little faster, but our account with circle does not allow docker layer caching which would make a considerable difference.

PR #1760, opened 2020-12-07
Increase request timeout when asking Azure to create a database

jon-shipley:
Fix for Azure migrations validation


GuyHarwood (approved):
Thanks for sorting this 🎖️

PR #1769, opened 2020-12-10
Feature/43327 ps report function 1

jon-shipley:
This is the first function in the Psychometric report generation pipeline.


tslib/src/functions-throttled/ps-report-1-list-schools/index.ts
```diff
@@ -0,0 +1,25 @@
+import { AzureFunction, Context } from '@azure/functions'
+import { performance } from 'perf_hooks'
+import { SchoolListService } from './list-schools-service'
+
+const functionName = 'ps-report-1-list-schools'
```
GuyHarwood, line 5: whats the purpose of the '1' in the name?

tslib/src/functions-throttled/ps-report-1-list-schools/list-schools-service.ts
```diff
@@ -0,0 +1,48 @@
...
+  private async getSchools (): Promise<School[]> {
+    const sql = 'SELECT id, name, urlSlug as uuid from mtc_admin.school'
+    return this.sqlService.query(sql)
+  }
+
+  public async getSchoolMessages (): Promise<SchoolMessage[]> {
+    this.logger.verbose(`${functionName}: ListSchoolsService called - retrieving all schools`)
+    const schools = await this.getSchools()
+    const schoolMessages: SchoolMessage[] = []
+    for (const school of schools) {
```
GuyHarwood, line 38: you could use `schools.map` here instead to make it less verbose.  trivial though.

GuyHarwood (approved):
LGTM.  just one clarification and minor suggestion

tslib/src/functions-throttled/ps-report-1-list-schools/index.ts
jon-shipley: According to the architecture, there will ultimately be 5 functions working on the PS report pipeline.  This naming scheme is 'ps-report-n-' as the prefix for these functions, where n indicates the pipeline stage.

I'm not committed to this naming scheme if you want to suggest something.

tslib/src/functions-throttled/ps-report-1-list-schools/list-schools-service.ts
jon-shipley: Yes,  good spot.  Happy with your suggestion. 👍

PR #1770, opened 2020-12-10
feature/44764 - analysis schema, role and permissions

GuyHarwood:
support for analysis schema, role and role permissions


db/misc/20201210144200.create-analysis-users.sql
```diff
@@ -0,0 +1,22 @@
+/* example script for creating analysis users and assigning to roles */
+
+CREATE USER [analysisUser1] WITH PASSWORD='', DEFAULT_SCHEMA=[mtc_analysis]
+CREATE USER [analysisanalysisUser2] WITH PASSWORD='', DEFAULT_SCHEMA=[mtc_analysis]
```
jon-shipley, line 4: typo?

db/misc/20201210144200.create-analysis-users.sql
```diff
@@ -0,0 +1,22 @@
+/* example script for creating analysis users and assigning to roles */
```
jon-shipley, line 1: If the intention is that this is a script that will not be run automatically, I think the intention would be clearer if we moved this to markdown and put it in the docs/ directory.

GuyHarwood: none of the scripts in this directory are automatic.  they are manual utility scripts that we have collected over time.

docs area doesn't feel like the right place for these.  they should be available as sql.

i have included a readme in this folder to indicate that they are manual

jon-shipley: Well you certainly got my heart rate up when I saw empty passwords until I realised this was in a different directory 😆

jon-shipley (approved):
LGTM

PR #1771, opened 2020-12-10
Feature/44776 test support api

jon-shipley:
Initial test-support api for low environments.


tslib/src/functions/util-test-support-api/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const httpTriggerFunc: AzureFunction = async function (context: Context, req: HttpRequest): Promise<void> {
+  if (!config.DevTestUtils.TestSupportApi) {
+    context.log('exiting as not enabled (default behaviour)')
+    return
+  }
+  // Respond in 230 seconds or the load balancer will time-out
+  // https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=javascript#limits
+  const start = performance.now()
+  const entity = context.bindingData.entity
```
GuyHarwood, line 24: maybe i've not had enough coffee yet, but how is this entity property populated?  the only binding we have is an incoming http request 🤔

GuyHarwood (approved):
LGTM

nice clean style on the api implementation 💯 

should be standard going forward

tslib/src/functions/util-test-support-api/index.ts
jon-shipley: The `entity` bit is defined as an alpha in the `route` property of the function.json. 

`      "route": "test-support/{entity:alpha}"`

GuyHarwood: ah yes, overlooked the route stuff in the function definition. thanks!

PR #1775, opened 2020-12-11
feature/44806 - rename function, consolidate util flag into one setting

GuyHarwood:
- prefix school pin http sampler with 'util'
- consolidate util enabled flag into one setting


jon-shipley (approved):
LGTM 

Many thanks for this 👍

PR #1783, opened 2020-12-16
Feature/44783 ps report func 2

jon-shipley:
PS report system - extract pupil data for later transformation


tslib/src/functions-throttled/ps-report-2-pupil-data/index.ts
```diff
@@ -0,0 +1,21 @@
+import { AzureFunction, Context } from '@azure/functions'
+import { performance } from 'perf_hooks'
+import { PsReportService } from './ps-report.service'
+import { PupilResult } from './models'
+
+const functionName = 'ps-report-2-pupil-data'
+
+const serviceBusQueueTrigger: AzureFunction = async function (context: Context, incomingMessage: any): Promise<void> {
```
GuyHarwood, line 8: could you type `incomingMessage` to give more clarity on the message structure?

tslib/src/functions-throttled/ps-report-2-pupil-data/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  const start = performance.now()
+  context.log.verbose(`${functionName}: called for school ${incomingMessage.name}`)
+  const outputBinding: PupilResult[] = []
+  context.bindings.psReportPupilMessage = outputBinding
+  const psReportService = new PsReportService(outputBinding, context.log)
+  await psReportService.process(incomingMessage.uuid)
```
GuyHarwood, line 14: i assume `uuid` is the `school.urlSlug`?

tslib/src/functions-throttled/ps-report-2-pupil-data/models.ts
```diff
@@ -0,0 +1,129 @@
...
+  loadingTime: number
+  nextBetweenQuestions: boolean
+  numpadRemoval: boolean
+  practice: boolean
+  questionReader: boolean
+  questionTime: number
+  compressCompletedCheck: boolean
+}
+
+// interface CheckWindow {
```
GuyHarwood, line 44: how come these are commented?

tslib/src/functions/check-marker/index.ts
```diff
@@ -1,34 +1,34 @@
...
-  }
-
-  const end = performance.now()
-  const durationInMilliseconds = end - start
-  const timeStamp = new Date().toISOString()
-  context.log(`${functionName}: ${timeStamp} run complete: ${durationInMilliseconds} ms`)
-}
-
-export default serviceBusQueueTrigger
+/* eslint-disable import/first */
```
GuyHarwood, line 1: is this just re-formatted?

GuyHarwood (commented):
Looking good, just a few questions...

tslib/src/functions-throttled/ps-report-2-pupil-data/index.ts
jon-shipley: Yes, I intend to.

tslib/src/functions-throttled/ps-report-2-pupil-data/index.ts
jon-shipley: yes.

tslib/src/functions-throttled/ps-report-2-pupil-data/models.ts
jon-shipley: They will be uncommented when this draft PR is complete.

tslib/src/functions/check-marker/index.ts
jon-shipley: Yes, just a whitespace change - line endings were changed to \n

func-consumption/package.json
```diff
@@ -7,7 +7,7 @@
     "node": ">= 10.14.1"
```
GuyHarwood, line 7: have you tried manually removing the `devDependencies` from this file to see if they are actually needed and/or re-added?

tslib/src/functions-throttled/ps-report-2-pupil-data/models.ts
```diff
@@ -0,0 +1,139 @@
...
+  data?: object
+  id: number
+  isWarmup: boolean | null
+  question?: string | null
+  questionCode?: string | null
+  questionNumber?: number | null
+  type: string
+}
+
+export type CheckConfigOrNull = CheckConfig | null
```
GuyHarwood, line 110: can you not express these with the optional operator `?`
or is it more the case that doing it like this makes it more explicit....

GuyHarwood (approved):
LGTM
function app, gulp and task stuff looks much better 💯

func-consumption/package.json
jon-shipley: Not as such.

The gulp stuff we clearly need, and I *believe* that IntelliJ in-editor linting *requires* Eslint et al. in order to work for monorepos.

Are there any packages you think could be ditched?

tslib/src/functions-throttled/ps-report-2-pupil-data/models.ts
jon-shipley: Yes, we can express the same thing without the alias but these types are used in the code (mostly the data service) and I felt it was cleaner - as now all the types are defined in a single place making change easier!

This already proved useful during the refactor to make the output readonly.

GuyHarwood: ok great.  makes sense, given that its easy to forget to add the operator everywhere you want to use the type

func-consumption/package.json
GuyHarwood: the jest stuff should not be in there, as the app itself does no testing.  
it may be a case of pruning each one in turn, to figure out what is needed.

there is some other cruft that we deploy which i noticed today, which we should discuss elsewhere.

PR #1795, opened 2021-01-05
feature/44602 replace obsolete app insights package with new one

GuyHarwood:
- [x] add new package
- [x] create dedicated singleton service with initialisation
- [x] update all existing call sites / dependencies
- [x] remove old package
- [x] test locally against app insights azure instance


GuyHarwood:
@jon-shipley as discussed - i've implemented a custom singleton service, which can be injected into components & services when needed.  There is also a dedicated plugin available [here](https://github.com/microsoft/ApplicationInsights-JS/tree/master/extensions/applicationinsights-angularplugin-js), but i feel that would be more difficult to control when we don't have a key present.  Although i haven't explored it, yet.

jon-shipley (approved):
lgtm 💯

PR #1796, opened 2021-01-07
Pupil-api: Update dependencies

jon-shipley:
Security fix

Update dependencies
Security fix: bump node-fetch to 2.6.1


GuyHarwood (approved):
LGTM

jon-shipley:
@GuyHarwood 

Please note this PR changes the behaviour of the `X-XSS-Protection` response header because `helmet` has changed, as has the wider web (e.g. google.com)

Previous behaviour was to set

`X-XSS-Protection` to `1; mode=block`

The new behaviour is to set:

`X-XSS-Protection` to `0`

This appears to be because setting it to `1; block` did not work as expected as t opened up a side-channel attack.  Given that most attacks are in the more advanced spectrum, this seems more preferable than to set it to `block` which only protects against trivial attacks, but opens up side-channel attacks.

Helmet Changelog: https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md
This issue changing the behaviour: https://github.com/helmetjs/helmet/issues/230
MDN documentation: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection


@activemq - should this new behaviour be approved by @GuyHarwood we will need to update the corresponding test.

GuyHarwood:
@jon-shipley sounds like the best approach to me 👍🏼

activemq (approved):

PR #1799, opened 2021-01-11
[42003] sticky-banner-and-pupil-inconsistent

jon-shipley:
UX bug: only count visible pupils, even if they were previously selected.


GuyHarwood (approved):
LGTM

PR #1800, opened 2021-01-12
Feature/45083 test env   timeouts when data sync func is triggered

jon-shipley:
Config update


jon-shipley:
@GuyHarwood 

This sets the tslib sql request timeout default to 10 minutes (was 1 minute) due to timeouts on the test env.  I'm open to other values though.  What do you think?  The timeout was from sync-results-to-sql function which has a large transactional insert that the test env is struggling with.

We have already overridden it on the Azure config to 15 minutes and that works.  Once this goes in we can remove that config and allow the default to surface.

However, the concern is that we could also see timeouts in production under heavy load.  Giving the DB more time to complete the work allows us to downscale further.

GuyHarwood:
@jon-shipley the timeout needs to be set to something that will tolerate realistic loads.  Has this feature been tested with realistic volumes?

jon-shipley:
@GuyHarwood - no - there is a ticket for load testing it which has not been played.

jon-shipley:
@GuyHarwood - as the test environment has limited DTUs (200) it may well be the bottleneck here for a large insert.  We can't really predict production values with this low-scaled DB.

GuyHarwood:
@jon-shipley then figure out through trial and error what DTU and timeout values will work to accommodate the test environment scenarios

tslib/src/config.ts
```diff
@@ -34,7 +35,7 @@ export default {
     database: process.env.SQL_DATABASE ?? 'mtc',
     connectionTimeout: parseInt(parser.valueOrSubstitute(process.env.SQL_CONNECTION_TIMEOUT, oneMinuteInMilliseconds), 10),
     censusRequestTimeout: parseInt(parser.valueOrSubstitute(process.env.SQL_CENSUS_REQUEST_TIMEOUT, twoHoursInMilliseconds), 10),
-    requestTimeout: parseInt(parser.valueOrSubstitute(process.env.SQL_REQUEST_TIMEOUT, oneMinuteInMilliseconds), 10),
+    requestTimeout: parseInt(parser.valueOrSubstitute(process.env.SQL_REQUEST_TIMEOUT, tenMinutesInMilliseconds), 10),
```
GuyHarwood, line 38: putting aside the actual reason for this - other requests using the same config that are performing common db updates could tie up connections for a long time in a bottleneck scenario.  Best practice suggests short timeouts so connections are not tied up for long.

It's like 2 different use cases, and i wonder if we should have a separate setting for 'long running' data operations, that you can utilise for the sync service instead.  Maybe we should discuss further.

PR #1802, opened 2021-01-14
Feature/43328 ps report transformation

jon-shipley:
Create the Psychometric report for a single pupil

Input: service bus queue message
Output: service bus queue message


func-consumption/disable-functions.env
```diff
@@ -1,20 +1,21 @@
...
-AzureWebJobs.pupil-auth.Disabled=false
-AzureWebJobs.pupil-feedback.Disabled=false
-AzureWebJobs.pupil-login.Disabled=false
-AzureWebJobs.pupil-prefs.Disabled=false
-AzureWebJobs.school-pin-generator.Disabled=false
-AzureWebJobs.school-pin-http-service.Disabled=false
-AzureWebJobs.school-results-cache-determiner.Disabled=false
-AzureWebJobs.school-results-cache.Disabled=false
+AzureWebJobs.ps-report-transformed.Disabled=false
+AzureWebJobs.pupil-auth.Disabled=ture
```
GuyHarwood, line 11: misspelling

tslib/src/functions/ps-report-transformer/mocks/answers.ts
```diff
@@ -0,0 +1,43 @@
+import moment from 'moment'
+import { Answer } from '../../../functions-throttled/ps-report-2-pupil-data/models'
```
GuyHarwood, line 2: might be better to pull this back into the root models folder?

tslib/src/functions/ps-report-transformer/report-line.class.ts
```diff
@@ -0,0 +1,95 @@
...
+    DOB: '',
+    Gender: '',
+    PupilID: '',
+    Forename: '',
+    Surname: '',
+    ReasonNotTakingCheck: null
+  }
+
+  constructor (
+    answers: AnswersOrNull,
```
GuyHarwood, line 31: if i recall correctly there is a shorthand way of doing constructor params with backing fields.
not always ideal, but worth knowing

GuyHarwood (commented):
LGSF

tslib/src/functions/ps-report-transformer/report-line.class.ts
jon-shipley: Yes, I am aware.  But here we are using getters and setters so the `answers` param is being set to the `_answers` property.

GuyHarwood (commented):
LGSF

GuyHarwood (approved):
LGTM 🏆

tslib/src/functions/ps-report-transformer/mocks/answers.ts
GuyHarwood: did you look into this @jon-shipley ?

jon-shipley: I did an initial investigation, but more time would be needed to determine the right fit here.

PR #1803, opened 2021-01-18
bug/44250 PR validation databases not being deleted via Azure CLI

GuyHarwood:
[This problem](https://github.com/Azure/azure-cli/issues/7285) is occurring again.

See issue linked above for reasons why this can occur (ARM/CLI cache doesn't see it because it was created via TSQL)

- [x] add step to create database via CLI first, meaning it is already in ARM cache when we come to delete it at the end.


jon-shipley (approved):
lgtm

PR #1807, opened 2021-01-26
bug/45255 - cross platform script for app insights API call to write data annotation

GuyHarwood:
PowerShell script to write data annotations on linux.

Once this is in master it will be integrated into the pipeline


deploy/app-insights/index.js
```diff
@@ -10,5 +10,52 @@ Script Parameters
...
+const yargs = require('yargs').argv
+const uuid = require('uuid').v4
+const moment = require('moment')
+
+const applicationId = yargs.applicationId
+const apiKey = yargs.apiKey
+const releaseName = yargs.releaseName
+const releaseDescription = yargs.releaseDescription
+const author = yargs.releaseAuthor
+const eventTime = moment().format('YYYY-MM-DD[T]HH:MM:SS')
```
jon-shipley, line 23: I think you mean `YYYY-MM-DD[T]HH:MM:ss.SSS` or (with timezone offset) `YYYY-MM-DD[T]HH:MM:ss.SSSZZ` or `.toISOString()`

jon-shipley (commented):
LGSF

deploy/app-insights/index.js
GuyHarwood: good catch, thanks

jon-shipley (approved):

PR #1810, opened 2021-02-01
bug/45311 disable xss protection for admin and pupil-spa

GuyHarwood:
nginx config files were still setting header to active, this should now be disabled.


jon-shipley (approved):
Thanks for picking up on this 👍 
LGTM

PR #1811, opened 2021-02-01
Fix date validator

jon-shipley:
Fix an error inadvertently introduced by the recent linter upgrade when resolving lint issues


GuyHarwood (approved):
LGTM

PR #1814, opened 2021-02-02
Feature/44351 move pupil la codes to db

jon-shipley:
Re-requesting review due to branch name change

Move LA Codes to DB
Mod pupil UPN validator to use DB to obtain LA codes


GuyHarwood (approved):
LGTM

PR #1824, opened 2021-02-15
bug/45543 - verify API headers tests fix

GuyHarwood:
Fix outstanding tests so that they correctly assert for `0` which disables the browser xss prevention check.


jon-shipley (approved):
LGTM

PR #1830, opened 2021-02-22
feature/45652 vso release pipeline app settings

GuyHarwood:
App settings are applied to Azure services during each deployment.

This work collects & consolidates those settings into the source code.

- [x] Collect all unique sets of app settings, per environment
- [x] Establish which environments can be grouped, via identical sets
- [x] Consider a base set, and differentials for other environments


deploy/release-config/z_original-archive/dev-test-demo-load/admin.txt
```diff
@@ -0,0 +1,24 @@
+-ALLOWED_WORDS $(SchoolPinWordSet)
```
jon-shipley, line 1: @GuyHarwood - why are there a bunch of files added to the `/z_original-archive/` folder?

GuyHarwood: these are the original ones, copied verbatim from the release pipeline definition.

They are to be removed before the PR is closed.  Next steps...

1. finalise reviews
2. apply settings
3. test release
4. dispose of original files in archive
5. sign off
5. close and merge PR

jon-shipley: Okay, as this is a special case, we can leave them in the PR for now.

jon-shipley (approved):
LGTM

PR #1833, opened 2021-03-03
43250 - add migration to remove null from mtc_pupil.checkComplete

jon-shipley:
Add migration: remove NULL from `mtc_pupil.checkComplete`


db/migrations/schema-objects/20210303101058.do.remove-null-from-checkcomplete.sql
```diff
@@ -0,0 +1,11 @@
+DROP INDEX IF EXISTS mtc_admin.pupil.pupil_school_id_index;
+GO
+
+ALTER TABLE mtc_admin.pupil
+    ALTER COLUMN checkComplete BIT NOT NULL;
+GO
+
+CREATE INDEX pupil_school_id_index ON mtc_admin.pupil (school_id) include (currentCheckId, attendanceId, checkComplete,
```
GuyHarwood, line 8: is this new, or hangover from a previous years optimisations?

jon-shipley: The index already exists, and it prevents the column from being altered.  I'm not sure it's a hangover though.

GuyHarwood (approved):
LGTM

PR #1839, opened 2021-03-12
Feature/45952 pupil auth rate limiting enhancements

GuyHarwood:
- [ ] change default request count to 1000
- [ ] update env var documentation
- [ ] add rate limiting to redis purge functionality in tech support area

### DIFF
--- a/admin/services/tech-support/redis.service.js
+++ b/admin/services/tech-support/redis.service.js
@@ -56,7 +56,8 @@ const redisService = {
       /^result:/,
       /^sasToken:/,
       /^schoolData.sqlFindOneById/,
-      /^settings$/
+      /^settings$/,
+      /^ratelimit::/
     ]
     return allowedPrefixes.some(regex => { return regex.test(key) })
   },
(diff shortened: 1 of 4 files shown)


admin/services/tech-support/redis.service.js
```diff
@@ -56,7 +56,8 @@ const redisService = {
       /^result:/,
       /^sasToken:/,
       /^schoolData.sqlFindOneById/,
-      /^settings$/
+      /^settings$/,
+      /^ratelimit::/
```
jon-shipley, line 60: This only allows us to delete a single rate-limit entry at a time.  More changes would be needed to add this to the batch-delete feature.

jon-shipley (approved):
LGTM 💯 

But rate limit items cannot be batch-deleted as it stands.

admin/services/tech-support/redis.service.js
GuyHarwood: this has now been removed as we are not supporting single drops for the rate limit sorted set

PR #1850, opened 2021-03-30
Fix #46179 - Bugfix for AA keypad removal

jon-shipley:
When the keypad was removed via an Access Arrangement the SPA was crashing during question initialisation
as the buttons it was binding the listener functions to was not present.

The fix detects the presence of the button before binding, rather than the corresponding AA as this is deemed to be more reliable.


GuyHarwood (approved):
LGTM

PR #1853, opened 2021-03-31
Feature/46162 load test scripts update

GuyHarwood:
- [x] update dummy data gen scripts
- [x] update jmeter scenario to target func-consumption based proxy


jon-shipley (commented):
LGTM

PR #1854, opened 2021-04-06
Bug/46203 multiple tio pins for same pupil

jon-shipley:
Fix bug in SQ view that was incorrectly allowing pupils who already had a TIO pin to generate another.

Bump dependencies


GuyHarwood (approved):
LGTM

PR #1857, opened 2021-04-09
Bug/46725 ps report reasons for not taking check not mapping to correct values

jon-shipley:
Bugfix

This PR fixes the Not Taking Check reason codes in the PS Report

A new spec change means this PR also has a feature change - the 'reasons' are now mapped to codes rather than numbers.


GuyHarwood (approved):
LGTM

PR #1863, opened 2021-04-14
[admin] feature/46211-pin-expiry-not-available-for-ntc-when-pin-expires

jon-shipley:
Allow pupils who have an expired pin to be set as not taking the check


GuyHarwood (approved):
LGTM

PR #1867, opened 2021-04-15
Feature/46217 code z pupil not taking check

jon-shipley:
Update data export mapping for reasons for not taking a check


GuyHarwood (approved):
LGTM, nice one for adding the comments for 2022 💯

PR #1875, opened 2021-04-23
Feature/46532 the retro input assistant link does not display until official check

jon-shipley:
Hide the retrospective input assistant link until the live check period.

### DIFF
--- a/admin/helpers/access-arrangements-overview-presenter.js
+++ b/admin/helpers/access-arrangements-overview-presenter.js
@@ -1,26 +1,41 @@
 'use strict'
+const ejsUtil = require('../lib/ejs-util')
 
-const accessArrangementsOverviewPresenter = {}
-
-/**
- * Return list of pupils with appropriate styling properties
- * @param {Array} pupils
- * @param {Array} hl
- * @param {Object} availabilityData
- * @returns {Array} pupil data
- */
-accessArrangementsOverviewPresenter.getPresentationData = (pupils, availabilityData, hl) => {
-  return pupils.map(pupil => {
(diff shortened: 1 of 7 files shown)


admin/helpers/access-arrangements-overview-presenter.js
```diff
@@ -1,26 +1,41 @@
...
+    return pupils.map(pupil => {
+      if (hl && hl.includes(pupil.urlSlug)) {
+        let arrangementsLn = pupil.arrangements.length
+        if (pupil.showDoB && arrangementsLn === 1) {
+          arrangementsLn = 2
+        }
+        pupil.verticalBarStyle = arrangementsLn > 1 ? `height:${235 - 35 * (7 - arrangementsLn)}px` : 'height:0px'
       }
-      pupil.verticalBarStyle = arrangementsLn > 1 ? `height:${235 - 35 * (7 - arrangementsLn)}px` : 'height:0px'
+      pupil.hasAAEditDisabled = !availabilityData.canEditArrangements || pupil.hasCompletedCheck || pupil.notTakingCheck
```
GuyHarwood, line 21: i was going to suggest moving this business rule to the aa service rather than the presenter.  But it is about controlling the presentation, and would require significant refactoring of the AA service.

GuyHarwood (approved):
LGTM

admin/helpers/access-arrangements-overview-presenter.js
jon-shipley: These lines were already there, I only moved them around to tidy it up.   There is quite a lot of refactoring to do here - the controller is looking fat.

PR #1879, opened 2021-04-30
Feature/44399 tech support queue replay tools 

GuyHarwood:
New feature for tech support module
- [ ] view storage & service bus queue counts (including dead letter / poison queues)
- [ ] allow replay of a specified queue
- [ ] design with consideration for replay of 'queue sets' for known support scenarios

### DIFF
--- a/admin/package.json
+++ b/admin/package.json
@@ -39,7 +39,7 @@
     "**/**/ajv": "^6.12.6"
   },
   "dependencies": {
-    "@azure/service-bus": "^1.1.0",
+    "@azure/service-bus": "^7.0.5",
     "applicationinsights": "^1.8.0",
     "axios": "^0.21.1",
     "azure-storage": "^2.10.2",
--- a/tslib/src/functions/queue-replay-service-bus/index.ts
+++ b/tslib/src/functions/queue-replay-service-bus/index.ts
@@ -1,0 +1,20 @@
+import { AzureFunction, Context } from '@azure/functions'
+import { performance } from 'perf_hooks'
+import { IQueueMessageReplayRequest } from './message-replay.service'
+
+const functionName = 'queue-replay-service-bus'
+
+
+
+function finish (start: number, context: Context): void {
+  const end = performance.now()
+  const durationInMilliseconds = end - start
+  const timeStamp = new Date().toISOString()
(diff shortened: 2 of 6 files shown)


admin/package.json
```diff
@@ -39,7 +39,7 @@
     "**/**/ajv": "^6.12.6"
   },
   "dependencies": {
-    "@azure/service-bus": "^1.1.0",
+    "@azure/service-bus": "^7.0.5",
```
jon-shipley, line 42: I didn't expect us to upgrade this for `admin` until July at the earliest. One alternative would be to move this to a HTTP function to interface with queues.

GuyHarwood: Yes i considered this, and we use it in both tslib and admin.  It has a smaller footprint on the admin app, and is used only to put prepare-check messages on the queue.

There are advantages of using version 7 in comparison to 1 for this PR, and there is nothing compelling against not using it, as the pin generation journey is tested on each PR.

tslib/src/functions/queue-replay-service-bus/index.ts
```diff
@@ -0,0 +1,20 @@
...
+function finish (start: number, context: Context): void {
+  const end = performance.now()
+  const durationInMilliseconds = end - start
+  const timeStamp = new Date().toISOString()
+  context.log(`${functionName}: ${timeStamp} run complete: ${durationInMilliseconds} ms`)
+}
+
+const queueReplayServiceBus: AzureFunction = async function (context: Context, busMessage: IQueueMessageReplayRequest): Promise<void> {
+  const start = performance.now()
+  context.log('nothing to do, yet...')
```
jon-shipley, line 16: This function doesn't do anything yet.  Is this for future work?

admin/spec/back-end/controllers/tech-support.spec.js
```diff
@@ -110,4 +111,19 @@ describe('tech-support controller', () => {
       expect(payloadService.getPayload).toHaveBeenCalledWith(checkCode)
     })
   })
+
+  describe('/queue-overview', () => {
+    it('GET: should send queueInfo as array', async () => {
```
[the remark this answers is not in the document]
GuyHarwood: ive added an assertion for that at the service level and renamed the controller test.

PR #1896, opened 2021-05-11
feature/46614-hdf-start-of-the-journey-copy-change

jon-shipley:
Copy change for the HDF


jon-shipley:
Consolidating into new PR

PR #1897, opened 2021-05-11
feature/46616-hdf-review-pupil-details-read-only

jon-shipley:
Copy change for hdf pupil review page


jon-shipley:
Consolidating into new PR

PR #1901, opened 2021-05-13
bug/46827 - tech support service bus queue shows total count instead of active

GuyHarwood:
data service was using wrong property `totalMessageCount` instead of `activeMessageCount`


jon-shipley (approved):
lgtm

PR #1902, opened 2021-05-13
bug/46824 - per environment service bus configuration

GuyHarwood:
- [x] enable 'per environment' configuration of service bus via environment variables
- [x] support staged integration of new functionality in Azure Pipelines for each environment 
- [x] allow re-configuration of existing queues
- [x] default to production values 
- [x] add shell support to configuration file for ruby


deploy/deploy.config.js
```diff
@@ -0,0 +1,79 @@
...
+      {
+        name: 'check-sync'
+      },
+      {
+        name: 'check-validation'
+      },
+      {
+        name: 'ps-report-schools',
+        DefaultMessageTimeToLive: sixDays,
+        MaxSizeInMegabytes: {}.hasOwnProperty.call(process.env, 'SERVICE_BUS_QUEUE_MAX_SIZE_PS_REPORT') ? parseInt(process.env.SERVICE_BUS_QUEUE_MAX_SIZE_PS_REPORT, 10) : fiveGigabytes
```
jon-shipley, line 56: Consider the failure scenario's here.

By setting the default to the setting fort a lower environment, if we forget to override the setting for prod then the queue will be initialised to 5GB, a figure that is too low.  An error will not be immediately raised and future errors are likely in the overnight runs when there is little oversight.

If we were to make the production values the default, then I believe the deployment would fail immediately and it would be immediately possible to remedy by re-releasing with the override setting.  Releases are watched by several people so it has a high chance of being noticed and corrected.

What do you think?

deploy/service-bus/create.v2.js
```diff
@@ -0,0 +1,60 @@
...
+
+function debug () {
+  const data = queues.map(q => {
+    console.dir(getQueueProperties(queueDefaults, q))
+  })
+}
+
+function getQueueProperties (queueDefaults, queueInfo) {
+  const props = R.mergeRight(queueDefaults, queueInfo)
+  return R.omit([ "name" ], props)
```
jon-shipley, line 39: 👍  for clarity

deploy/deploy.config.js
GuyHarwood: yes, i see the logic here.  Prioritise production over all others.

it does mean that other environments have more default config, but as i have preserved the older scripts it can be staged as necessary.

jon-shipley (commented):
LGTM 💯

jon-shipley (approved):
LGTM 💯

PR #1903, opened 2021-05-14
Feature/46815 add new role sta admin

jon-shipley:
Add new role: STA-ADMIN

Tickets:

* 46815
* 46816
* 46741


GuyHarwood (approved):
LGTM

PR #1923, opened 2021-06-10
feature/47176 results sync

GuyHarwood:


tslib/src/functions-throttled/sync-results-to-sql/sync-results.data.service.ts
```diff
@@ -36,6 +42,104 @@ export class SyncResultsDataService implements ISyncResultsDataService {
...
+      name: 'checkResultId',
+      type: TYPES.Int,
+      value: checkResultInfo.checkResultId
+    })
+    params.push({
+      name: 'userDeviceId',
+      type: TYPES.Int,
+      value: checkResultInfo.userDeviceId
+    })
+    return this.sqlService.modify(sql, params)
```
jon-shipley, line 140: I would have expected `modifyWIthTransaction()` but the code above does not have a TRY handler or explicit Rollback. Curious as to why?

GuyHarwood: good idea, keeps it consistent with the rest

jon-shipley (approved):
LGTM

PR #1929, opened 2021-06-22
Feature/42515 pupil status page content changes

jon-shipley:
* copy change for pupil status page for after the check window closes
* fix 2 timer functions to not run if the timer is past due


GuyHarwood (approved):
LGTM

PR #1932, opened 2021-06-24
Bug/47418 sync results parameters

jon-shipley:
[tslib/sync-results-to-sql] better transaction handling for very large events and inputs


GuyHarwood (approved):
LGTM

PR #1938, opened 2021-06-26
[tslib/sync-results-to-sql] allow unusually large battery discharge v…

jon-shipley:
Results sync
Allow unusually large battery discharge values


GuyHarwood (approved):
LGTM

PR #1947, opened 2021-07-07
47429 - [db] mask checkResult.mark with a better default

jon-shipley:
Was the default of 0
Now: 99

Screenshot 2021-07-07 at 15 42 35


GuyHarwood (approved):
odd that according to the docs it doesn't allow a static numeric value!
LGTM

jon-shipley:
Well, it's a very restricted range.

PR #1952, opened 2021-07-15
chore - delete sql migrations archive

GuyHarwood:
delete the old migrations folder now that we are successfully running on the new db app.


jon-shipley (approved):
LGTM

PR #1953, opened 2021-07-15
Feature/47662 docker build agent pt.2

GuyHarwood:
- [x] configure for non root user
- [x] fix ruby mismatch


jon-shipley (commented):
lgsf 👍

PR #1954, opened 2021-07-19
Bug/47447 not attending can select completed check pupils

jon-shipley:
* fix bug in not attending selection
* add integration test
* bump minor versions in pupil-api
* update deprecated faker calls


GuyHarwood (approved):
LGTM. great to see an integration test 💯

PR #1956, opened 2021-07-20
Feature/46157 remove columns from ps report table

jon-shipley:
Remove DeviceType and DeviceTypeModel from mtc_results.psychometricReport
Bump versions in `db` migration app


GuyHarwood (approved):
LGTM

PR #1961, opened 2021-07-27
chore - remove unused check window service from tslib

GuyHarwood:
delete check window service from tslib. unused


jon-shipley (approved):
lgtm

PR #1969, opened 2021-08-17
Feature/copy changes rollback 47644 47638

jon-shipley:
Various copy changes to revert the 2021 special copy back to standard.
Re-allow all pupil-not-taking-check reasons


GuyHarwood (approved):
LGTM

PR #1970, opened 2021-08-18
Security updates: path-parse, ioredis

jon-shipley:
Consolidated security updates for path-parse and ioredis


GuyHarwood (approved):
thanks for bundling these up
LGTM

PR #1979, opened 2021-08-20
[admin] major dependency upgrades

jon-shipley:
* gulp-sass now no longer bundles sass, so `sass` (dart-sass) package added and a fix made in _tables.scss
* jest upgrade to v27: we still have plenty of jasmine api usage so the testrunner has been set to `jest-jasmine2` rather than the newer default.
* integration-tests: added some wait time before and after the test to allow things to settle and avoid an error
* packages removed: the following packages were no longer referenced in the admin app and were removed: postgrator, serve-favicon, iconv-lite, htst
* upgrades of note: mssql upgraded to the latest
* SQL config change - now verifies the server cert


GuyHarwood (approved):
LGTM

PR #1982, opened 2021-08-24
[admin] Bug fix -Allow pupil restart by STA-Admin

jon-shipley:
Bug fix - allow pupil restart by STA-Admin
Additionally, make the link between the detached JS-enabled submit button and the form it is submitting explicit.


GuyHarwood (approved):
LGTM

PR #1983, opened 2021-09-02
Bug/41813 pupil identification flag service

jon-shipley:
Bug fix for pupil identification service
Refactoring for callers


GuyHarwood (approved):
much better than the old one 💯

PR #1997, opened 2021-09-14
Feature/47696 show pupil alias in printed version

jon-shipley:
Use the pupil aliases (if available) on the pin slips

Maintenance:
  - pupil-api dependency updates
  - admin updates
  - db updates


GuyHarwood (approved):
LGTM
nice to see jasmine being phased out 🎖️

PR #2003, opened 2021-09-28
bug/42526 lookup table name issues

GuyHarwood:
- refactor table names
- refactor associated column names
- adjust code as necessary


jon-shipley:
@activemq - there is no point testing this yet as I am waiting on @GuyHarwood to submit a change to rename the table to "Lookup" rather than "LookUp"

jon-shipley (approved):
lgtm

PR #2004, opened 2021-09-28
Spike/49003 replace obsolete azure-storage with new libraries

GuyHarwood:
- create poc for queue, blob and sas token methods
- refactor admin app to use new libraries
- refactor tslib to use new libraries


admin/bin/init-dev-storage.js
```diff
@@ -19,68 +19,69 @@ try {
...
+const connectionString = process.env.AZURE_STORAGE_CONNECTION_STRING
+
 const queueNames = names.queues
 const tableNames = names.tables
 const poisonQueues = queueNames.map(q => q + '-poison')
 const allQueues = queueNames.concat(poisonQueues)
-const tableService = getPromisifiedService(azure.createTableService())
-const queueService = getPromisifiedService(azure.createQueueService())
+
+async function deleteTableEntries (tableName) {
```
jon-shipley, line 38: Should we actually be calling our own interface here, rather than the MS implementation directly?

admin/bin/init-dev-storage.js
```diff
@@ -19,68 +19,69 @@ try {
...
-      }
-      console.log(`Found ${entities.length} entities to delete in batch ${batch++} from ${table}`)
-      entities.forEach(entity => {
-        deletions.push(tableService.deleteEntityAsync(table, entity))
-      })
-      await Promise.all(deletions)
-    }
+    deletions.push(deleteTableEntries(table))
   }
+  await Promise.all(deletions)
```
jon-shipley, line 54: I think `Promise.allSettled()` offers better functionality in this case, as deletions are not dependent on each other.  Available since node 12.9.0 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

admin/controllers/pupil.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const azureFileDataService = require('../services/data-access/azure-file.data.service')
+const azureBlobDataService = require('../services/data-access/azure-blob.data.service')
```
jon-shipley, line 3: Isn't this the equivalent of having the data service in the controller?

admin/services/data-access/azure-blob.data.service.js
```diff
@@ -1,32 +1,37 @@
 'use strict'
 
-const azureStorage = require('azure-storage')
-const bluebird = require('bluebird')
+const { BlobServiceClient } = require('@azure/storage-blob')
 const config = require('../../config')
 
-let blobService
+// TODO cache the client once loaded, or is it better to recreate each time?
```
jon-shipley, line 6: As blob uploads are fairly uncommon let's agree to recreate the service as needed each time?

admin/services/data-access/payload.data.service.js
```diff
@@ -27,8 +27,8 @@ const payloadDataService = {
     }
     const schoolUrlSlug = res[0].urlSlug
     const table = 'receivedCheck'
-    const tableService = azureTableService.getPromisifiedAzureTableService()
-    return tableService.retrieveEntityAsync(table, schoolUrlSlug.toLowerCase(), checkCode.toLowerCase())
+    // GUY check the return on this is preserved
```
jon-shipley, line 30: Note to self

admin/controllers/pupil.js
GuyHarwood: yes, i'm not finished with the refactor yet.  this is still a draft PR.

step 1 - merge file and blob into 1 serviced
step 2 - eradicate superfluous methods from service
step 3 - refactor data services out

admin/bin/init-dev-storage.js
GuyHarwood: done

admin/services/data-access/azure-table.data.service.js
```diff
@@ -1,36 +1,41 @@
...
-    })
+    }
+  },
 
-    return azureTableService
+  clearTable: async function clearTable (tableName) {
+    const tableClient = TableClient.fromConnectionString(connectionString, tableName)
+    const entityIterator = tableClient.listEntities()
+    const deletions = []
+    for await (const entity of entityIterator) {
```
jon-shipley, line 27: This could presumably take a really long time to complete if the table was really big.   I presume this is only for non-prod workloads.  If so, how can we surface that?

admin/tests-integration/azure-table.data.service.spec.js
```diff
@@ -0,0 +1,161 @@
...
+        return deleteTable(t)
+      })
+      await Promise.all(deletions)
+    } catch (error) {
+      fail(`failed to delete one or more tables, azure storage may need manual clean up.\n${error}`)
+    }
+  })
+
+  describe('retrieveEntity', () => {
+    it('returns raw entity when exists', async () => {
```
jon-shipley, line 60: really new tests should be written in jest, so `test` rather than `it` is preferred.

admin/tests-integration/azure-table.data.service.spec.js
```diff
@@ -0,0 +1,161 @@
...
+      const randomPrefix = getUniqueTableName()
+      const tableNames = [
+        getUniqueTableName(),
+        `${randomPrefix}-Bad&chars^1`
+      ]
+      try {
+        await sut.createTables(tableNames)
+        fail('should have thrown an error due to hyphen in table name')
+      } catch (error) {
+        expect(error).toBeDefined()
```
jon-shipley, line 157: Jest has a better, more readable, idiom for async tests that are expected to throw:
`await expect(sut.createTables(tableNames)).rejects.toThrow(/some error message/)`

that would remove the entire try/catch block.

GuyHarwood: updated both new tests that use try..catch

PR #2019, opened 2021-10-19
Feature/45930 add new school page

jon-shipley:
Add feature for the service manager to manually add a new organisation

Screenshot 2021-10-19 at 13 21 23


admin/spec/back-end/controllers/service-manager.spec.js
```diff
@@ -1,4 +1,4 @@
-/* global jasmine describe expect it beforeEach spyOn test jest afterEach */
```
GuyHarwood, line 1: 🏆  we should make a point of including this with every piece of work now

PR #2021, opened 2021-10-22
Fix: admin session timeout modal

jon-shipley:
Add ajax call to refresh the session on the server to '/keep-alive'

Modify session-expiry.js to pick up the initial expiryDate on page load from the body tag.  If the session is extended by the user, the new expiryDate is sent back in the JSON.

The frontend app.js code has had the vendor files split from it to aid in code-coverage testing.


GuyHarwood (approved):
LGTM

PR #2023, opened 2021-10-27
feature/46909 - pupil auth include build info

GuyHarwood:
- include build info in auth request from pupil app
- send build info in submitted check
- persist build info in pupil-api when received


jon-shipley (approved):
LGTM 💯

PR #2025, opened 2021-10-27
Feature/46036 load testing

paulflood5:
Added the Flood.io Load Agent Install for Windows Walkthrough.

I have added verification steps for each task step to support with automating the Task in Azure Pipeline.


GuyHarwood (approved):
LGTM

PR #2033, opened 2021-11-10
Add service message to service manager home page

jon-shipley:
- [x] add serviceMessage to service-manager home page
- [x] allow paragraphs in service message
- [x] add message preview when setting the message
- [x] black text in the service message


admin/package.json
```diff
@@ -20,17 +20,17 @@
     "start": "yarn build && node --max-http-header-size 15000 ./dist/server.js",
     "start:dev": "yarn dev-build && concurrently -r 'nodemon --inspect --delay 1500ms --watch ./dist --max-http-header-size 15000 ./dist/server.js' 'gulp watch'",
     "test": "yarn lint && yarn jest",
-    "test:watch": "yarn jest --watch --coverage=no",
-    "test:nocoverage": "yarn lint && yarn jest --coverage=no",
-    "test:coverage": "yarn lint && yarn jest --coverage=yes",
+    "test:watch": "yarn jest --watch",
+    "test:nocoverage": "yarn lint && yarn jest",
```
GuyHarwood, line 24: i'm a bit confused, as `test:nocoverage` appears to be the same as just `test`
are you doing this on purpose for compatibility purposes with any existing usages in the CI pipeline?

GuyHarwood (commented):
LGTM.  just one question regarding the yarn scripts

admin/package.json
jon-shipley: Tbh, I don't think anything is actually using `test:nocoverage`.  I think the default, of no coverage, is perfect during development.  Shall I just remove `test:nocoverage`?

GuyHarwood: yes please do.

GuyHarwood (approved):
LGTM
great to see the sanitizer and various tests 💯

PR #2035, opened 2021-11-17
Feature/48948 2022 teacher pin gen scenario

paulflood365:
Updated teacher journey for 2022 saved as v3.jmx

Created a directory 2022_PR as a temp for the updated JMX, some issue with folder 2022 previously created.

PR #2036, opened 2021-11-17
Delete 2022

paulflood365:


GuyHarwood:
commit to be done on main PR #2035

PR #2042, opened 2021-11-25
Feature/load test data gen

jon-shipley:
Add utilities for test data generation

1. Stored procedure to relatively quickly add test pupils
2. Utility to generate the UPN check letter from the base serial number
3. Update deps for `db`
4. Update deps for `pupil-spa`


GuyHarwood (approved):
LGTM

paulflood365:
Yes nice Jon, will have a try, I would like to put a dataSetID as a suffix on the pupil name and on the school (Last name), as part of the call, 

EXEC [mtc_admin].[spGenTestPupils] @numPupils = 10, @schoolIdFrom = 2, @schoolIdTo = 5, dataSetID = 'aac';

It is good practise when running the assertions on checking the data returned on the GUI, and if there is any test data already in when the database is set up,

And for the ClearRedisCache script to verify.

I am doing it manually at the moment without a variable.

CONCAT('Pupilaab', CAST(temp30.rowNumber AS VARCHAR)), -- foreName 
    CONCAT('Schoolaab', CAST(school.id AS VARCHAR)), -- lastName

jon-shipley:
@paulflood365 - I've called the parameter `dataSetCode` as IDs should be reserved for PKs, and in any case, this is a varchar(12), not an int.

PR #2043, opened 2021-11-26
Feature/48515 appinsights metadata enhancements

jon-shipley:
Add custom properties to AI


GuyHarwood (approved):
LGTM

this PR reminds me - the SPA has semi colon terminators everywhere.  all the other apps don't.  we also need to bring the linting rules in line with the other apps too.

would be good to do this as a ticket if it doesnt clash with all the angular code

PR #2048, opened 2021-12-01
Fix/replay node14 pr

GuyHarwood:
this PR replays the node 14 upgrade PR that had to be reverted due to it being merged too early.


jon-shipley (approved):
Looks good to me 👍 

On some future PR, we should be able to alter the compilation target for the admin app and functions.

PR #2053, opened 2021-12-06
Feature/50724 postgrator update

GuyHarwood:
implement own connection and query executor, ahead of v5 changes, which remove the integrated sql client.

our scripts fail due to inclusion of baseline with multiple batches


GuyHarwood:
closing as work is significant to make compatible at this time

PR #2091, opened 2022-01-25
update page objects

activemq:
Updated names of page objects along with var names to be more in line with current functionality


GuyHarwood (approved):
LGTM

PR #2069, opened 2022-01-13
Feature/41055 allow removal of retro input assistant

GuyHarwood:
- remove wholesale AA removal link from overview page
- introduce edit button on overview page
- allow removal of retro input assistant from overview page

PR #2092, opened 2022-01-31
Fix/41055 orphaned commits

GuyHarwood:
5 commits were mistakenly missed out when #2069 was merged to master.

This PR introduces those commits.


jon-shipley (approved):
LGTM 💯 
and thanks for converting these tests over 🏅

PR #2097, opened 2022-02-02
Feature/security roll up

activemq:
Rollup of security prs


GuyHarwood (approved):
LGTM

jon-shipley:
@GuyHarwood @activemq I don't understand why we are patching code under `z_archive`?

I would think either deleting the code or instructing Dependabot to ignore that path would be appropriate as it is historical, isn't it?

GuyHarwood:
good spot @jon-shipley 

i'd prefer to just delete it, as part of this PR

GuyHarwood:
regarding ignores - would be good to do that for the spikes folder

PR #2130, opened 2022-02-16
bug/51752 school audit user display

GuyHarwood:
When viewing the school audit history in the SM area, the user.identifier value was displayed.

When the system uses DfE sign-in, this value is populated with a UUID.

This fix prioritises the user.displayName value, and falls back to the identifier.

'system' is displayed for non user changes.


jon-shipley (commented):
lgtm 👍

PR #2147, opened 2022-03-07
feature/51834 - remove usage of old service bus library from deploy/service-bus

GuyHarwood:
- update calls to `create` to `createV2`
- refactor `delete` to use newer service bus package
- remove `azure` package from `deploy/service-bus`


deploy/service-bus/package.json
```diff
@@ -1,10 +1,9 @@
 {
   "name": "service-bus-deploy",
   "version": "1.0.0",
-  "description": "creates service bus queues",
+  "description": "manages service bus entities",
   "main": "index.js",
   "scripts": {
-    "createqs": "node ./create.js",
```
jon-shipley, line 7: why change the script? I think it's only an implementation change?

deploy/service-bus/readme.md
```diff
@@ -11,7 +11,7 @@ Creates service bus queues defined in queues-topics.json, if they do not already
 ## Usage
 
 `yarn deleteqs` or `npm run deleteqs` - deletes all queues specifieid in `queues-topics.json`
-`yarn createqs` or `npm run createqs` - creates all queues specifieid in `queues-topics.json`
+`yarn createV2` or `npm run createV2` - creates all queues specifieid in `queues-topics.json`
```
jon-shipley, line 14: Again, why change the script target?

jon-shipley (approved):
LGTM 👍 

Many thanks for this PR.  I am not sure why we are changing the script name, but seeing as the changes are consistent I am comfortable leaving it as you suggest with the new name.

deploy/service-bus/package.json
GuyHarwood: `createqs` is legacy and was only left in to support the transition while it was tested in dev, before eventually implemented all the way down to production.
I am removing it because it's no longer used now.

GuyHarwood:
@jon-shipley i would actually prefer the scripts to be aligned as `createqs` and `deleteqs`, but i feel the pipeline disruption might bite us so close to go live

PR #2149, opened 2022-03-09
Bug/51933 Access Arrangements edit validation failure

GuyHarwood:
- diagnose cause of error
- switch param from DFE number to school id


GuyHarwood:
polluted with commits from another PR

PR #2160, opened 2022-03-21
bug/52200-remove-download-link-from-testdev-homepage

jon-shipley:
Remove a link from the test developer home page that is no longer needed.  Pupil data is no longer downloaded, and the service functionality has already been removed, but this link got left.


GuyHarwood (approved):
LGTM

PR #2172, opened 2022-03-31
Feature/49323 service message border colour

jon-shipley:
* new colours added: blue, orange, green and the previously available red
* fixed editing / error pages
* new notification banner style


GuyHarwood (approved):
LGTM

PR #2201, opened 2022-04-21
Feature/52719 tslint to eslint

GuyHarwood:


func-consumption/package.json
```diff
@@ -38,10 +38,7 @@
     "gulp-typescript": "^5.0.1",
     "gulp-yarn": "^3.0.0",
     "standard": "^17.0.0",
```
jon-shipley, line 40: Why are we keeping standard if we are moving to eslint? `func-throttled` also includes `standard`, which I think would be nice if we could get rid of it too.

func-ps-report/package.json
```diff
@@ -28,10 +28,7 @@
     "gulp-typescript": "^5.0.1",
     "gulp-yarn": "^3.0.0",
     "standard": "^16.0.3",
```
jon-shipley, line 30: Again, why keep standard?

jon-shipley (approved):
lgtm

PR #2214, opened 2022-04-29
Feature/52865 Service Manager pupil search

GuyHarwood:
Add pupil search & summary to Service Manager area


admin/package.json
```diff
@@ -141,7 +141,8 @@
   "standard": {
     "ignore": [
       "assets/javascripts/jquery-3.5.1.js",
-      "assets/vendor-js/*"
+      "assets/vendor-js/*",
+      "**/*.ts"
```
jon-shipley, line 145: I thought `standard` could lint ts files.  Why ignore them?

jon-shipley (approved):
LGTM - lovely tests 😃 

One question, though.

admin/package.json
GuyHarwood: not completely - https://github.com/standard/standard#can-i-use-a-javascript-language-variant-like-flow-or-typescript

if you open the admin app as the root in your editor, you get syntax errors.
we should assess this carefully with eslint and figure out the best path.
i will create a ticket for that

PR #2217, opened 2022-05-18
Feature/46732 SPA queue submissions & core code improvements

GuyHarwood:
- Remove obsolete azure storage library and replace with raw HTTP calls
- Tighten tsconfig settings to improve codebase and require strict typings


pupil-spa/.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
jon-shipley, line 2: I am surprised that this `//comment is allowed in a JSON file.

jon-shipley (approved):
Nice work - thank you for doing this! 🏅

pupil-spa/.vscode/launch.json
GuyHarwood: it will no doubt be auto generated, as it's VsCode specific, which has a file type called 'JSON with comments'

PR #2218, opened 2022-05-20
feature/53146 - add option on SPA queue messages so they do not expire

GuyHarwood:
add option for message TTL so it does not expire

FYI - no unit tests for this due to existing issues with implementation.  This is being refactored in #2217


jon-shipley (approved):
lgtm

PR #2222, opened 2022-06-01
Feature/53220 freeze a pupil 

GuyHarwood:
enable freezing a pupil so it cannot be modified on the core journeys


admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    const sql = 'UPDATE mtc_admin.pupil SET frozen=1 WHERE urlSlug=@pupilUrlSlug'
+    const params = [{
+      name: 'pupilUrlSlug',
+      type: TYPES.UniqueIdentifier,
+      value: pupilUrlSlug
+    }]
+    return sqlService.modify(sql, params)
+  }
+
+  public static async thawPupil (pupilUrlSlug: string): Promise<void> {
```
jon-shipley, line 71: Docs missing
Returns something when should be void.

What happens why if we call thaw() on an already thawed pupil? Do we care enough to throw? Or emit an info log?

admin/services/service-manager/pupil-annulment/pupil-annulment.data.service.ts
```diff
@@ -0,0 +1,135 @@
...
+      @currentUnconsumedRestartId = pr.id
+    FROM
+      [mtc_admin].[pupilRestart] pr
+    WHERE
+      pr.isDeleted = 0
+    AND
+      pr.pupil_id = @pupilId
+    AND
+      pr.check_id IS NULL
+    ORDER BY 1 DESC
```
jon-shipley, line 54: Please enlighten me: what does `ORDER BY 1 DESC` actually do?

admin/views/service-manager/annul-pupil.ejs
```diff
@@ -0,0 +1,41 @@
...
+            <%- partial('partials/_form-error', {
+              pageFields: ['upn'],
+              error
+          }) %>
+
+            <form action="" method="post">
+                <input type="hidden" name="_csrf" value="<%= csrftoken %>"/>
+                <div class="govuk-form-group">
+                    <label class="govuk-label" for="upn">
+                        Confirm UPN for <%=pupil.lastName %>,<%=pupil.firstName %>
```
jon-shipley: How does this work if the UPN is not unique? I was expecting a pupil urlSlug...

db/migrations/schema-objects/20220614111014.do.vewpupilseligibleforlivepingeneration-frozen.sql
```diff
@@ -0,0 +1,38 @@
+DROP VIEW IF EXISTS [mtc_admin].[vewPupilsEligibleForLivePinGeneration]
+GO
+
+CREATE VIEW [mtc_admin].[vewPupilsEligibleForLivePinGeneration] AS
```
jon-shipley, line 4: Why not just ALTER the view?

db/migrations/schema-objects/20220614111014.undo.vewpupilseligibleforlivepingeneration-frozen.sql
```diff
@@ -0,0 +1,38 @@
+DROP VIEW IF EXISTS [mtc_admin].[vewPupilsEligibleForLivePinGeneration]
+GO
+
+CREATE VIEW [mtc_admin].[vewPupilsEligibleForLivePinGeneration] AS
```
jon-shipley, line 4: Would be simpler to ALTER the view?

db/migrations/schema-objects/20220614143015.do.modify-vewpupilswithactivelivepins.sql
```diff
@@ -0,0 +1,33 @@
+DROP VIEW IF EXISTS [mtc_admin].[vewPupilsWithActiveLivePins]
```
jon-shipley, line 1: An ALTER view may be less disruptive during the deployment?

jon-shipley (commented):
LGSF 👍 

Nice work. I've added some comments for your consideration.

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    }]
+    return sqlService.readonlyQuery(sql, params)
+  }
+
+/**
+ * @description look up a pupils frozen flag by their urlSlug
+ * @param {string} pupilUrlSlug
+ * @returns {Promise<Array<any>>}
+ */
+  public static async isFrozenByUrlSlug (pupilUrlSlug: string): Promise<any> {
```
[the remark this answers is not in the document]
GuyHarwood: the unpacking is being done in the service when its an implementation specific detail of the sql service. 

I will refactor this

admin/services/service-manager/pupil-annulment/pupil-annulment.data.service.ts
GuyHarwood: it returns the last entry, as its ordering by the column at ordinal 0, the primary key.

admin/views/service-manager/annul-pupil.ejs
GuyHarwood: regards uniqueness of UPN - the UX journey has you search via UPN, you choose the pupil you want to view in the summary, then you choose to annul (all guided by the URL slug), and then you simply confirm the UPN.  its not used in the selection during annulment, just to confirm the UPN is the same as the one you just viewed

db/migrations/schema-objects/20220614111014.do.vewpupilseligibleforlivepingeneration-frozen.sql
GuyHarwood: updated all to `CREATE OR ALTER VIEW`

admin/services/service-manager/pupil-service/service-manager.pupil.service.ts
```diff
@@ -42,6 +44,8 @@ export class ServiceManagerPupilService {
     if (p.length === 0) return undefined
 
     const status = await this.getPupilStatus(p[0].id)
+    console.log(`GUY: attendanceCode:${p[0].attendanceCode}`)
```
jon-shipley, line 47: Debug?

admin/views/service-manager/annul-pupil-undo.ejs
```diff
@@ -0,0 +1,41 @@
...
+              error
+          }) %>
+
+            <form action="" method="post">
+                <input type="hidden" name="_csrf" value="<%= csrftoken %>"/>
+                <div class="govuk-form-group">
+                    <label class="govuk-label" for="upn">
+                        Confirm UPN for <%=pupil.lastName %>, <%=pupil.firstName %>
+                    </label>
+                    <input class="govuk-input" id="upn" name="upn" type="input" spellcheck="false"
```
jon-shipley, line 19: Let's have an online chat about this form.

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {Promise<void>}
+ */
+  public static async freezePupil (pupilUrlSlug: string): Promise<void> {
+    const sql = 'UPDATE mtc_admin.pupil SET frozen=1 WHERE urlSlug=@pupilUrlSlug'
+    const params = [{
+      name: 'pupilUrlSlug',
+      type: TYPES.UniqueIdentifier,
+      value: pupilUrlSlug
+    }]
+    return sqlService.modify(sql, params)
```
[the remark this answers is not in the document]
GuyHarwood: because you should return the task/work unless you specifically need to work with the outcome

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {Promise<void>}
+ */
+  public static async thawPupil (pupilUrlSlug: string): Promise<void> {
+    const sql = 'UPDATE mtc_admin.pupil SET frozen=0 WHERE urlSlug=@pupilUrlSlug'
+    const params = [{
+      name: 'pupilUrlSlug',
+      type: TYPES.UniqueIdentifier,
+      value: pupilUrlSlug
+    }]
+    return sqlService.modify(sql, params)
```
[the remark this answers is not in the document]
GuyHarwood: as above

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+   public static async countFrozenByPupilIds (pupilIds: Array<number>): Promise<any> {
+    const paramData = sqlService.buildParameterList(pupilIds, TYPES.Int,
+      'pupilId')
+    const sql = `SELECT COUNT(id) as [frozenCount]
+      FROM mtc_admin.pupil WHERE frozen=1 AND id IN (${paramData.paramIdentifiers.join(', ')})`
+    return sqlService.readonlyQuery(sql, paramData.params)
+  }
+
+  public static async freezePupil (pupilUrlSlug: string): Promise<void> {
```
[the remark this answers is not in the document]
jon-shipley: Why would we change the modify function? The return statement is in this method.

admin/services/service-manager/pupil-annulment/pupil-annulment.data.service.ts
```diff
@@ -0,0 +1,228 @@
...
+      attendanceId=@attendanceCode_id,
+      frozen=1,
+      checkComplete=0,
+      restartAvailable=0
+    WHERE
+      id=@pupilId
+    `
+    // get previously active restart record id and attendance record id for rollback
+    // delete unconsumed restarts
+    return sqlService.modifyWithTransaction(sql, params)
```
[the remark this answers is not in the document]
GuyHarwood: as above - its returning a promise / piece of work, that itself can return something on completion

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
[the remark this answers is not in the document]
jon-shipley: That's fine but it isn't returning <void>, it's returning a "modify response object".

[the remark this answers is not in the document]
jon-shipley: There is no requirement for an async function to have a `return` statement at all.  It might be slightly more efficient to just return the underlying promise from the sqlService, but this code is not on the critical path.

```
> async function doNothing () {
... console.log('nothing doing');
... }
undefined
> const p = doNothing()
nothing doing
undefined
> p
Promise {
  undefined,
  [Symbol(async_id_symbol)]: 301,
  [Symbol(trigger_async_id_symbol)]: 5,
  [Symbol(destroyed)]: { destroyed: false }
}
```

Async function always return a Promise, and undefined is the default return value.

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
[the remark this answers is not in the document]
GuyHarwood: overall, the comments in this PR are more about the fundamentals of async/await, returning work and what the outcome of that work is.  I think we need to clarify on best practice - should you avoid holding on for the outcome of a piece of work if you don't need to do anything with the return value? or does the high level promise pattern just 'deal with that under the hood'?

there are some subtle semantics in other languages for this particular scenario, but i would rather we get straight on best practices for typescript and keep it that way.

ultimately, my concern is that you want to hang about here (via the `await`) for a piece of work to finish, and then just return void, which seems like a waste of an asynchronous piece of work, because you are essentially blocking until its finished, then returning nothing.  what if the caller wanted the piece of work, then to wait until many of them were complete and do something with the aggregated outcome?

admin/services/pupil-frozen.service/pupil-frozen.data.service.ts
GuyHarwood: i pondered this, and then got distracted.

lets discuss

jon-shipley (approved):
LGTM 🏅 

Many thanks for the changes

PR #2227, opened 2022-06-14
bug/53226 - function error logging defaults and ps report timeout

GuyHarwood:
- allow PS report to run for just under 23 hours
- set logging default to 'Error' for all functions


jon-shipley (approved):
LGTM

PR #2232, opened 2022-06-25
bug/53461-ps-report-restart-code-not-available

jon-shipley:
Expired pins break the relationship between the pupilRestart table and the
check that was actually used for the restart.  This code  provides a
work-around by searching the restarts to find the code.


GuyHarwood (approved):
LGTM

PR #2233, opened 2022-06-28
bug/53617 - fix query to include received checks that failed processing

GuyHarwood:
accommodate situation where received check failed processing, and teacher would like to exclude from attendance


admin/services/data-access/pupils-not-taking-check.data.service.js
```diff
@@ -78,8 +78,8 @@ const pupilsNotTakingCheckDataService = {
            AND (
              -- No check has been generated for the pupil
                  p.currentCheckId IS NULL OR
-                 -- Or, a check was assigned but not received - cover new, collected and not received checks
-                 (p.currentCheckId IS NOT NULL AND p.checkComplete = 0 AND c.received = 0))
+                 -- Or, a check was assigned but is not complete - cover new, collected and not received checks
+                 (p.currentCheckId IS NOT NULL AND p.checkComplete = 0))
```
jon-shipley, line 82: I suspect this can be reduced to:

```sql
SELECT
	p.foreName,
	p.middleNames,
	p.lastName,
	p.dateOfBirth,
	p.urlSlug,
	p.group_id
FROM
	[mtc_admin].[pupil] p
WHERE
	p.attendanceId IS NULL
	AND p.checkComplete = 0
```

GuyHarwood: it will need `p.school_id = @schoolId`

PR #2248, opened 2022-07-15
Feature/ps doc update

jon-shipley:
Update PS report documentation


GuyHarwood (approved):
LGTM
although there is a question to be answered (outside of this PR) about the use of omnigraffle files.  We should probably opt for an open format such as draw.io
We need a spike on that to understand the limitations of conversion etc 
I looked into this last year and was able to open a graffle file in draw.io after conversion, but it did make the source quite limited for further editing.

PR #2252, opened 2022-07-27
Feature/53962 post live check window state tactical

jon-shipley:
Tactical solution for Read-Only mode at end of check


admin/app.js
```diff
@@ -275,6 +276,14 @@ app.use(function (req, res, next) {
   next()
 })
 
+app.use(function (req, res, next) {
+  // set up the current check window phase
+  // do this for every request
+  // @ts-ignore - var declared in server.js
+  global.checkWindowPhase = checkWindowPhaseConsts.readOnlyAdmin // TODO: query data and check window period to perform this check dynamically
```
GuyHarwood, line 283: great idea

admin/availability/middleware.js
```diff
@@ -29,4 +40,20 @@ async function isAdminWindowAvailable (req, res, next) {
...
+/**
+   * The UK academic year starts in September and continues through to August of the following year.
+   * TODO: move to lib
+   * @param {moment.Moment} date
+   * @return {number}
+   */
+function getAcademicYear (date) {
+  const i = date.month()
+  const y = date.year()
+  const september = 8
```
GuyHarwood, line 52: september is the 9th month - what am i not understanding when reading the code?

GuyHarwood (commented):
LGSF
just a couple of questions

admin/availability/middleware.js
jon-shipley: moment.month() is zero-based, so it will return 0-11 when doing a get, and it expects a value of 0-11 when setting passing a value in.

I have no idea why dates want to be zero-based, as we have considerable mental models of Jan being the 1st month, and not the 0th.

GuyHarwood: i did wonder if it was that.  very odd design choice

jon-shipley: I've added a comment as a reminder

GuyHarwood (approved):
LGTM

PR #2255, opened 2022-08-02
Feature/54015 post check window state service manager UI

jon-shipley:
Add UI element to allow the service manager to toggle the behaviour of the post check admin period between read-only and unavailable states.

Screenshot 2022-08-02 at 13 34 03


GuyHarwood (approved):
LGTM

PR #2257, opened 2022-08-11
Feature/54124 add created at to sm pupil search results

jon-shipley:
Add the createdAt date to the service manager pupil results (when more than one result) for disambiguation

Screenshot 2022-08-11 at 16 51 07


GuyHarwood (approved):
LGTM

PR #2260, opened 2022-08-24
Bug/53881 ps report initial non numeric key used as tfirstkey

jon-shipley:
Fixes negative recall time under certain situations with Audio version enabled.


GuyHarwood (approved):
LGTM

PR #2268, opened 2022-09-05
Bug/54381 check marker treat azure errors as soft errors

jon-shipley:
Update admin app to sql retry on socket errors
Update tslib sql retry to parity with admin app


GuyHarwood (approved):
LGTM

PR #2295, opened 2022-09-13
bug/54798 assets image contains no assets

GuyHarwood:
- add `yarn build` to assets dockerfile now that install does not automatically trigger it


jon-shipley (approved):
lgtm

PR #2302, opened 2022-09-15
bug/54886 pupil auth failing to start

GuyHarwood:
move swagger deps to non dev dependency section, as always imported whether enabled or not

### DIFF
--- a/pupil-api/src/App.ts
+++ b/pupil-api/src/App.ts
@@ -35,7 +35,7 @@ class App {
   public express: express.Application
 
   // Run configuration methods on the Express instance.
-  constructor () {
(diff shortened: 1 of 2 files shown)


pupil-api/src/App.ts
```diff
@@ -35,7 +35,7 @@ class App {
   public express: express.Application
 
   // Run configuration methods on the Express instance.
-  constructor () {
+  constructor() {
```
jon-shipley, line 38: Lint is moaning about this change.

GuyHarwood: corrected

PR #2319, opened 2022-10-05
Bug/55188 service unavailable page shows wrong academic year

jon-shipley:
Update copy on service unavailable page


GuyHarwood (approved):
LGTM

PR #2321, opened 2022-10-06
add hooks to tests which are to assert for check started timestamp

activemq:
- [x] Added "@check_started" tag to test to ensure it is a live check and that it is completed

- [x] Add "@check_started" tag to the After hook, to ensure that once the scenario is finished that the check started timestamp is tested


GuyHarwood (approved):
LGTM

PR #2323, opened 2022-10-10
Update school_import_steps.rb

activemq:
Check for the last school to be imported instead of the first to ensure all schools have been imported before checking individually


GuyHarwood (approved):
LGTM

PR #2324, opened 2022-10-12
bump runtime to node 16 LTS

GuyHarwood:
- bump to 16 LTS
- add documentation for guidance on procedure


func-consumption/.nvmrc
```diff
@@ -1 +1 @@
-v14.18.1
+v16.17.1
```
jon-shipley, line 1: According to [1] we should be on azure-functions-core-tools version 4 to support node16 on the production target. 
The same comment applies to the other two function apps.

We may need appropriate settings on Azure too?  FUNCTIONS_EXTENSION_VERSION = 4 and WEBSITE_NODE_DEFAULT_VERSION = 16

[1] https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=azure-cli%2Cwindows%2Cin-process%2Cv4&pivots=programming-language-javascript

GuyHarwood: good catch, i had overlooked this.

annoyingly those settings are in the local.settings.json file which is not tracked by git, and the azure environments will need upgrading manually.

i will figure this out.

GuyHarwood: functions have been tested on the following...
- functions v3, node 14
- functions v4, node 14
- functions v4, node 16 (now operational in azure)

jon-shipley (approved):
lgtm 💯

PR #2328, opened 2022-10-27
Feature/54224 ctf download change version dynamically

jon-shipley:
* Dynamically determine the ctf version in the CTF file (results)
* Bump admin app deps


GuyHarwood (approved):
LGTM

PR #2329, opened 2022-10-27
Feature/54225 school import exclude proposed to open

jon-shipley:
* Remove proposed to open school from the organisation upload
* Bump deps for tslib


GuyHarwood (approved):
LGTM

PR #2335, opened 2022-11-11
Feature/54232 Ps Report - manual execution

GuyHarwood:
manually execute the Ps Report from the tech support area.


admin/controllers/tech-support.js
```diff
@@ -492,11 +493,11 @@ const controller = {
   },
 
   postCheckResultsResyncAll: async function postCheckResultsResyncAll (req, res, next) {
-    res.locals.pageTitle = 'Resync All Results'
+    res.locals.pageTitle = 'Resync Uncalculated Results'
```
jon-shipley, line 496: It's not really "uncalculated" though is it?  More like 'Un-synchronised".

Even better, as the role is tech support here, perhaps something like "Synchronise new checks from Table Storage to Azure SQL".  


If the checkbox is ticked then the historical stuff is resynchronised too.

admin/controllers/tech-support.js
```diff
@@ -552,6 +553,41 @@ const controller = {
...
     }
+  },
+
+  getPsReportRun: async function getPsReportRun (req, res, next) {
+    try {
+      res.locals.pageTitle = 'PS Report Run'
+      req.breadcrumbs('PS Report Run')
+      res.render('tech-support/ps-report-run', {
+        breadcrumbs: req.breadcrumbs(),
+        runReport: req.body?.runReport ?? false,
```
jon-shipley, line 564: So this will be either a string (possibly empty) or a boolean. I don't like giving it more than one type.  Is there a reason for it?

admin/controllers/tech-support.js
```diff
@@ -552,6 +553,41 @@ const controller = {
...
+        response: ''
+      })
+    } catch (error) {
+      return next(error)
+    }
+  },
+
+  postPsReportRun: async function postPsReportRun (req, res, next) {
+    const runReport = req.body.runReport === 'true'
+    const response = runReport === true ? 'PS Report Requested' : 'Report NOT Requested - checkbox tick required'
```
jon-shipley, line 574: I don't see why you are mixing types here.  `runReport` is sometimes a string and sometimes a bool.   If we need a boolean flag for the server logic, then I would prefer a separate variable that is always a bool that is derived from the incoming form post.

admin/services/tech-support/ps-report-exec/ps-report-exec.data.service.ts
```diff
@@ -0,0 +1,54 @@
...
+      displayName: record.displayName
+    }
+  }
+
+  public static async sendPsReportExecMessage (message: IExecPsReportRequest): Promise<void> {
+    if (this.sbClient === undefined || this.sbSender === undefined) {
+      this.sbClient = new ServiceBusClient(config.ServiceBus.connectionString)
+      this.sbSender = this.sbClient.createSender('ps-report-exec')
+    }
+    console.dir(message)
```
jon-shipley, line 49: debug?

admin/controllers/tech-support.js
GuyHarwood: yeah doesn't need to be high level, and makes more sense to us.
I've changed the checkbox to `Include checks that were previously synchronised`

admin/controllers/tech-support.js
GuyHarwood: it's a string value, as its the state of the checkbox HTML element.  I've tried the various combinations and it has to be parsed from its string value to work.  I've updated that line to reflect this

admin/controllers/tech-support.js
GuyHarwood: i am creating a boolean and setting its state based upon the string value of the checkbox element.
It does read a bit wordy, so i'll trim it back and just use the form element string value.

admin/services/tech-support/ps-report-exec/ps-report-exec.data.service.ts
GuyHarwood: yep, caught this earlier, already removed

admin/controllers/tech-support.js
```diff
@@ -492,11 +493,11 @@ const controller = {
   },
 
   postCheckResultsResyncAll: async function postCheckResultsResyncAll (req, res, next) {
-    res.locals.pageTitle = 'Resync All Results'
+    res.locals.pageTitle = 'Synchronise new checks from Table Storage to main SQL database'
     const resyncAll = req.body.resyncAll === 'true' // convert string to bool- the user must tick the checkbox
     try {
       await resultsResyncService.resyncAllChecks(resyncAll)
-      req.breadcrumbs('Check Results - Resync All')
+      req.breadcrumbs('Check Results - Resync Uncalculated')
```
jon-shipley, line 500: I believe the "All" is correct here - it will resynchronise checks that are already in the db from table storage.

jon-shipley (approved):
LGTM

PR #2340, opened 2022-11-24
disallow restart deletion if pin generated against restart

GuyHarwood:
prevent restart deletion if a pin has been generated against a restart.


jon-shipley (approved):
LGTM

PR #2345, opened 2022-11-30
Feature/54227 test schools

jon-shipley:
* new db column on the schools table `isTestSchool`
* db migration for existing environments populate the flag
* prevent test school data from being output into the ps report


GuyHarwood (approved):
LGTM

PR #2350, opened 2022-12-01
feature/54228 sign hdf in readonly mode 

GuyHarwood:
- remove obsolete results functionality from HDF
- allow signing of hdf once checks disabled


jon-shipley (approved):
LGTM

PR #2375, opened 2022-12-13
feature/55601 QR Code route for SPA

GuyHarwood:
- create new route to identify logins via QR code
- update QR code to link to new route


pupil-spa/src/app/qr-code-arrival/qr-code-arrival.component.html
```diff
@@ -0,0 +1,14 @@
+<main id="content" class="js-content" role="main">
+        <div class="grid-row">
+            <header>
+                <h1 class="heading-xlarge">Redirecting</h1>
+            </header>
+            <div class="notice">
+                <strong class="visually-hidden">Please wait</strong>
+                <strong class="aa-copy-size">
+                    Redirecting to home page
```
jon-shipley, line 9: It would be good to have a link to click here, in case the redirection failed.

jon-shipley (approved):
Looks good to me 💯 

One suggestion made for the landing page.

pupil-spa/src/app/qr-code-arrival/qr-code-arrival.component.html
GuyHarwood: good call.  this page never shows, but just in case...

PR #2427, opened 2023-01-27
feature/54122 browser tab Id

GuyHarwood:
explore recording an id for the current tab in check data packets.


GuyHarwood:
API is browser extension specific.

Too much work at this point in the annual cycle.
POC should be merged for reference

PR #2449, opened 2023-02-13
Feature/55611 audit pupil age reason

jon-shipley:
Add further auditing to the mtc_admin.pupilAgeReason table

* createdAt
* updatedAt + trigger
* recordedBy_userId - FK to user.id table
* lastUpdatedBy_userId - FK to user.id table.  Useful for auditing the changed reason.

Fix: the pupilAgeReasonService will only be called from pupilEditService when the the pupil age reason has changed, not every time the pupil is edited.


GuyHarwood (approved):
LGTM

PR #2451, opened 2023-02-15
Feature/57438 nginx custom error pages

GuyHarwood:
- add custom error pages for nginx on SPA
- add catch-all handler for angular


jon-shipley (approved):
lgtm

PR #2452, opened 2023-02-15
Feature/57884 update pupil api deps

jon-shipley:
Dependency Updates
----------------------

* @azure/service-bus v7.8.0
* applicationinsights v2.4.1
* dotenv v16.0.3
* helmet v6.0.3
* ioredis v5.3.1
* uuid v9.0.0
* winston-azure-application-insights v4.0.0

+ dev dependency updates including eslint and new lint fixes 😬 

Security
---------

Remove deprecated directive block-all-mixed-content from Content-Security-Policy
Remove deprecated header: Expect-CT

Add headers:--

Cross-Origin-Embedder-Policy: require-corp
Cross-Origin-Opener-Policy: same-origin
Cross-Origin-Resource-Policy: same-origin
Origin-Agent-Cluster: ?1

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Resource-Policy
https://whatpr.org/html/6214/origin.html#origin-keyed-agent-clusters


pupil-api/src/controllers/api-response.ts
```diff
@@ -1,4 +1,4 @@
-import { Response } from 'express'
+import type { Response } from 'express'
```
GuyHarwood, line 1: just read up about this, we should use it more where appropriate

GuyHarwood (approved):
LGTM

thanks for tidying up the code!

PR #2454, opened 2023-02-17
feature/54226 - Dfe Sign in Missing org info error distinction

GuyHarwood:
provide distinction on why the user could not be logged in


admin/error-types/DsiSchoolNotFoundError.js
```diff
@@ -9,6 +9,22 @@ class DsiSchoolNotFoundError extends mtcError {
...
+  constructor (dfeSignInUser) {
+    if (dfeSignInUser.organisation === undefined) {
+      const message = `User ${dfeSignInUser.providerUserId} has no organisation data`
+      super(message, 'No information about your associated school could be found')
+    } else if (dfeSignInUser.organisation.urn === undefined) {
+      const message = `User ${dfeSignInUser.providerUserId} has no urn associated with their organisation data`
+      super(message, 'We could not find a URN in your associated school data')
+    } else {
+      const message = `User ${dfeSignInUser.providerUserId} has no urn associated with their organisation data`
+      super(message, 'We could not find a URN in your associated school data')
```
jon-shipley, line 22: This message can't be disambiguated from the above.  That may be fine, but I don't know just from reading this.  Is it likely we would want to know the code path taken?

PR #2458, opened 2023-02-20
feature/56369 unsupported browser app insights tracking

GuyHarwood:
add app insights to unsupported browser page
- add App Insights to unsupported page
- update docker setup to inject instrumentation key into static page
- manually call page view event as rewrite hides 'real' URL
- fix NGINX `try_files` config (single URL error was being handled by NGINX, fixed in #2451 )
- update rewrite config for NGINX


jon-shipley (approved):
LGTM

PR #2465, opened 2023-03-03
Feature/58072 mtc closed guard

GuyHarwood:
redirect teachers logging on when the system is out of service.

return the unavailable page and short circuit the loading of school data to prevent any mismatch errors


jon-shipley (approved):
lgtm

PR #2475, opened 2023-03-16
Feature/security 15 3 23

activemq:


jon-shipley (changes requested):
tslib package.json needs to be bumped to 7.8.1 as per the change in func-throttled which wil auto-roll it out to other function apps after ('./setup.sh') has been run.

jon-shipley (approved):
LGTM :)

Many thanks for the changes

PR #2489, opened 2023-04-11
Feature/58445 freeze a pupil

jon-shipley:
SM Feature: Pupil Thaw

Change a pupil from a frozen (read-only) state back to an editable state.


admin/views/service-manager/pupil-summary.ejs
```diff
@@ -23,10 +23,17 @@
               <li>
               <% if (pupil.isAnnulled) { %>
                 <a class="govuk-link" href="/service-manager/annul-pupil-undo/<%=pupil.urlSlug %>" >Undo Annulment</a>
-                <% } else { %>
+                <% } else if (!pupil.isFrozen) { %>
```
GuyHarwood, line 26: Hiding the annulment when frozen could be confusing for the user.  I would have thought it be perfectly acceptable to go straight to annulment from frozen, given that this is one of the two outcomes from the freeze period.  Should we ask for clarification?

GuyHarwood (commented):
Looks good, just one item to address

admin/views/service-manager/pupil-summary.ejs
jon-shipley: That's a good point.  I'll change it.

GuyHarwood (approved):
LGTM

activemq:
Closing, covered in #2504

PR #2504, opened 2023-05-05
Feature/58445 freeze pupil 2

activemq:
SM Feature: Pupil Thaw

Change a pupil from a frozen (read-only) state back to an editable state.


jon-shipley (approved):
lgtm

PR #2505, opened 2023-05-05
Bug/58880 pupil add edit validation bug

jon-shipley:
Fix issue where the upn was validated as the raw user input, but was transformed before saving to the database (trim, uppercase).

Bump admin app minor/patch dependencies.


GuyHarwood (approved):
LGTM

PR #2507, opened 2023-05-11
feature/58890 Attendance Code Visibility for Service Manager

GuyHarwood:
- surface attendance code visibility as service manager page


GuyHarwood:
@jon-shipley the service manager controller is huge, and getting unwieldy to work with.

i think we should look at breaking it down into areas/specifics

admin/routes/service-manager.js
```diff
@@ -221,4 +221,14 @@ router.post('/annul-pupil-undo/:slug',
   serviceManagerController.postPupilAnnulmentUndo
 )
 
+router.get('/attendance-codes',
+  isAuthenticated([roles.serviceManager]),
+  serviceManagerController.getAttendanceCodes
+)
+
+router.post('/attendance-codes-submit',
```
jon-shipley, line 229: I usually use the same URL as the GET for the POST.  There is nothing wrong with this, just pointing out it is different from the GET req. 😄

admin/services/service-manager/attendance/service-manager.attendance.data.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        value: attendanceCode,
+        type: TYPES.Char(5)
+      },
+      {
+        name: 'visible',
+        value: visible,
+        type: TYPES.Bit
+      }
+    ]
+    await sqlService.query(sql, params)
```
jon-shipley, line 31: Why is this using sqlService.query() and not sqlService.modify()?

admin/services/service-manager/attendance/service-manager.attendance.data.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+export interface AttendanceCode {
+  id: number
```
jon-shipley, line 36: I don't think exposing the db id is useful to the app.  The code is unique isn't it?

admin/services/service-manager/attendance/service-manager.attendance.service.ts
```diff
@@ -0,0 +1,23 @@
...
+   * @param visibleAttendanceCodes - The attendance codes to be visible
+   */
+  static async setVisibleAttendanceCodes (visibleAttendanceCodes: string[] | undefined): Promise<void> {
+    if (!visibleAttendanceCodes || visibleAttendanceCodes.length === 0) {
+      visibleAttendanceCodes = []
+    }
+    const allAttendanceCodes = await ServiceManagerAttendanceDataService.getAttendanceCodes()
+    for (const attendanceCode of allAttendanceCodes) {
+      const visible = visibleAttendanceCodes.includes(attendanceCode.code)
+      await ServiceManagerAttendanceDataService.setVisibility(attendanceCode.code, visible)
```
jon-shipley, line 20: I would normally discourage making sql calls in a for loop.  This should really be one call to set all attendance codes in one go, or perhaps all modified attendance codes. However, as the service manager will only do this very rarely it will probably be okay. 🤔

jon-shipley (commented):
yeah, the service manager controller has grown arms and legs.  Good call on the refactor 👍 

This looks good!  Just some comments to think about first.

admin/services/service-manager/attendance/service-manager.attendance.data.service.ts
GuyHarwood: naughty copilot.  will have a word with the mischievous robot about how to write code properly

admin/services/service-manager/attendance/service-manager.attendance.data.service.ts
GuyHarwood: i reckon i can make it work without it, as i dont use it for anything meaningful, other than naming repeated HTML elements, which i can use the attendance code for anyway.

admin/services/service-manager/attendance/service-manager.attendance.service.ts
GuyHarwood: yes, i brooded on this for a while, and its extremely unlikely that there will ever be enough entries to cause a concerning amount of calls.  Plus, as you say - its for the service manager and unlikely to be used once a year, if at all.

GuyHarwood: i decided to refactor it into one transaction for all the updates.  i'd rather set the right example!

admin/services/service-manager/attendance/service-manager.attendance.data.service.ts
```diff
@@ -0,0 +1,65 @@
...
+  static async getAttendanceCodes (): Promise<AttendanceCode[]> {
+    const sql = `
+      SELECT code, reason, [order], visible
+      FROM mtc_admin.attendanceCode
+      WHERE isPrivileged = 0
+      ORDER BY [order]`
+    return sqlService.readonlyQuery(sql)
+  }
+
+  static async setVisibility2 (attendanceCodes: AttendanceCodeVisibility[]): Promise<void> {
```
jon-shipley, line 14: Do we still need setVisibility() and setVisibility2()?

jon-shipley (approved):
LGTM :)

admin/services/service-manager/attendance/service-manager.attendance.data.service.ts
GuyHarwood: good catch, will amend!

PR #2510, opened 2023-05-17
feature/59053 Refactor pupil feedback form to one question

GuyHarwood:
Refactor pupil feedback form to a single question


pupil-api/yarn.lock
```diff
@@ -17,7 +17,7 @@
   dependencies:
     tslib "^2.2.0"
 
-"@azure/core-amqp@^3.3.0":
+"@azure/core-amqp@^3.1.0":
```
jon-shipley, line 20: Why the downgrade here as well as in PR #2511 ?  Any security issue for us?

jon-shipley (approved):
LGTM :)

One question on a package downgrade.  Unsure why package.lock is being changed anyway...

pupil-api/yarn.lock
GuyHarwood: not sure why this is in here either.  will roll it back

PR #2555, opened 2023-06-23
bug/59439 prevent attendance changes after HDF signed

GuyHarwood:
bug in view logic was allowing removal of attendance codes after HDF signed
- fix view logic
- implement guard at routing level


admin/routes/pupils-not-taking-the-check.js
```diff
@@ -28,6 +27,7 @@ router.get(
   '/remove/:pupilId',
   isAuthenticated([roles.teacher, roles.helpdesk, roles.staAdmin]),
   isAdminWindowAvailable,
+  refuseIfHdfSigned,
```
jon-shipley, line 30: Good addition! 👍

PR #2559, opened 2023-06-28
bug/53295 guard against errors when obtaining number of questions

GuyHarwood:
- implement guard against undefined reference to `this.questions` in SPA


jon-shipley (approved):
LGTM 👍

PR #2561, opened 2023-06-30
bug/59558 PS Report: set attemptId to null if never logged in

GuyHarwood:
- set AttemptId, FormID and TestDate to null if the pupil never logged into the check
- if an attendance code is present, do not initialise any of the check data fields


jon-shipley (approved):
LGTM

PR #2562, opened 2023-06-30
Bug/59512 ps report pupil status mod

jon-shipley:
Prioritise the pupil status for Not taking the check in the PS Report.


GuyHarwood (approved):
LGTM

PR #2564, opened 2023-07-03
Bug/59557 restart number null or int

jon-shipley:
Mod for mtc_results.psychometricReport. RestartNumber will be:

1. NULL for pupils set to not taking check
2. NULL for incomplete pupils
3. a number for pupils who have logged in to take a check (whether we received it back or not)


GuyHarwood (approved):
LGTM

PR #2565, opened 2023-07-03
Bug/59543 ps report restart available overrides checkcomplete for 2023

jon-shipley:
Fix pupil status for cases where the pupil has been given a restart and is also check complete.


GuyHarwood (approved):
LGTM

PR #2647, opened 2023-11-06
feature/61995 Helpdesk access to Pupil History

GuyHarwood:
Enable the helpdesk role to access the pupil history area via the pupil register


jon-shipley (approved):
LGTM 👍

PR #2654, opened 2023-11-17
feature/61920 Remove ps report logger and associated utilities

GuyHarwood:
- remove PS Report logger function
- redirect log calls to context.log functions
- remove associated pages from tech support
- remove utility function


jon-shipley (approved):
lgtm 👍

PR #2679, opened 2023-12-22
Feature/61927 ps report output to new table each run

jon-shipley:


db/migrations/schema-objects/20240103120853.do.rename-table-psychometricreport.sql
```diff
@@ -0,0 +1,7 @@
+IF EXISTS(SELECT *
+            FROM INFORMATION_SCHEMA.TABLES
+           WHERE TABLE_NAME = 'psychometricReport'
+             AND TABLE_SCHEMA = 'mtc_results')
+BEGIN
+    EXEC sp_rename 'mtc_results.psychometricReport', 'psychometricReport_2023'
```
GuyHarwood, line 6: remember that these migrations won't get run on the 2023 production instance (i have literally just retargeted prod to a new 2024 instance).

So should we be dropping the table instead?  seems like the cleaner route...

func-consumption/relauncher.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
GuyHarwood, line 1: what is the purpose of this?

would be good to add a header comment to explain

tslib/src/functions-ps-report/ps-report-1-list-schools/index.ts
```diff
@@ -5,27 +5,24 @@ import { PsReportLogger } from '../common/ps-report-logger'
...
   logger.verbose(`requested at ${jobInfo.dateTimeRequested} by ${jobInfo.requestedBy}`)
   const start = performance.now()
   const meta = { processCount: 0, errorCount: 0 }
   const jobDataService = new JobDataService()
   try {
-    await jobDataService.setJobStarted(jobInfo.jobUuid)
+    // We need to store a filename for all the data to be written to during the staging process.
+    const now = moment()
+    const filename = `ps-report-staging-${now.format('YYYY-MM-DD-HHmm')}.csv`
+    await jobDataService.setJobStarted(jobInfo.jobUuid, { meta: { filename } }, context.log)
```
GuyHarwood, line 21: 💯

GuyHarwood (commented):
Great job on this huge piece of work!

Just a couple of comments to address

db/migrations/schema-objects/20240103120853.do.rename-table-psychometricreport.sql
jon-shipley: Happy to change to drop now that prod has been retargeted to a new DB.

func-consumption/relauncher.sh
jon-shipley: This is save me from Microsofts decision to just let the `func` process exit every time it gets a network timeout.  By using this handy-dandy `relaucnher` script every time the process dies it will get relaunched.

I use it for long running end to end tests. 

Will add a header as suggested. 👍

GuyHarwood (approved):
LGTM

GuyHarwood (approved):
LGTM

PR #2698, opened 2024-02-20
Feature/61924 load test data

GuyHarwood:
- Tidy up `load-test` directory and remove superfluous 'versioned' files
- upgrade compatibility level of database to 160 to allow use of [GENERATE_SERIES](https://learn.microsoft.com/en-us/sql/t-sql/functions/generate-series-transact-sql?view=sql-server-ver16) function
- switch from sql edge docker image back to standard now apple chipset is supported
- create 'db setup' scripts directory for reset and dummy data generation


jon-shipley (approved):
Lgtm 👍

PR #2716, opened 2024-03-12
Feature/62546 maladmin codes 2024

GuyHarwood:
- update annulment and maladministration codes for 2024 cycle
- refactor service manager area to accommodate new codes


jon-shipley (approved):
LGTM

jon-shipley:
Blocked as this contains changes to the PS report which would be better to merge in after the #2679

PR #2726, opened 2024-03-18
Feature/qa debug

activemq:
- [x] identify cause of issue test fails
- [x] Redis ruby access blocked by firewall
- [x] Redis ruby code refactored to be explicit about why value not found
- [x] Redis inaccessible - fail entire test suite run and exit early

### DIFF
--- a/test/admin-hpa/features/support/env.rb
+++ b/test/admin-hpa/features/support/env.rb
@@ -37,8 +37,8 @@ require_relative '../../features/support/app'
 require 'jwt'
 include Helpers
 
-# logger = Selenium::WebDriver.logger
-# logger.level = :debug
+ logger = Selenium::WebDriver.logger
+ logger.level = :debug
 
 Dotenv.load('../../.env')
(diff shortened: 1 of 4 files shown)


test/admin-hpa/features/support/env.rb
```diff
@@ -61,7 +61,6 @@
   config.exact = true
   config.ignore_hidden_elements = false
   config.visible_text_only = true
-  config.default_max_wait_time = 7
```
GuyHarwood, line 64: when its not clear what unit the config item requires we create a variable that describes it, and assign that to the property.
Like this...
```
sevenSeconds = 7
config.default_max_wait_time = sevenSeconds
```

test/pupil-hpa/features/support/redis_helper.rb
```diff
@@ -2,16 +2,17 @@ class RedisHelper
 
   def self.wait_for_prepared_check(school_password,pin)
     p school_password, pin
-    found = false
     begin
       retries ||= 0
       sleep 5
```
GuyHarwood, line 7: same as above, use a named variable with the unit in its name

test/admin-hpa/features/support/env.rb
activemq: Will do going forward, this config was removed as part of this commit

test/pupil-hpa/features/support/redis_helper.rb
activemq: Will do as above

GuyHarwood (approved):
LGTM

PR #2746, opened 2024-03-28
Feature/62241 404 should not raise an exception

jon-shipley:
* Fix bug in error handler
* Create new 404 page
* Remove the Error when creating a 404 to prevent an exception being raised in AI

Screenshot 2024-03-27 at 16 45 31


GuyHarwood (approved):
LGTM

PR #2748, opened 2024-04-04
Feature/64026 remove radio buttons and free text box from access arrangements

jon-shipley:
UI changes for access arrangements

Further reasons when selecting Question Reader are no longer required.


GuyHarwood (approved):
LGTM

PR #2782, opened 2024-04-18
Feature/64176 school pins & 10k seed fix

GuyHarwood:
- fix laCode issue on 10k inserts
- move school pin words to table
- include default set of random words
- only use default set of random words if `ALLOWED_WORDS_SET` is not present
- if present, insert to table via the env var above


db/migrations/core-data/20240417115737.do.school-pin-data.js
```diff
@@ -0,0 +1,884 @@
+'use strict'
+
+module.exports.generateSql = function () {
+
+  if (process.env.ALLOWED_WORDS_SET !== undefined) {
+    const sql = process.env.ALLOWED_WORDS_SET.split(',').map(word => {
+      return `INSERT INTO mtc_admin.[schoolPin] (val) VALUES ('${word}')`
```
jon-shipley, line 7: I would personally like to see whitespace trimmed and the word lowercased to enforce correct behaviour.

jon-shipley (approved):

PR #2785, opened 2024-05-03
feature/64193 Request PS Report for specific schools

GuyHarwood:
- Add ability to specify CSV list of URNs when requesting PS report


tslib/src/functions-ps-report/common/ps-report-service-bus-messages.ts
```diff
@@ -5,6 +5,7 @@ export interface PsReportListSchoolsIncomingMessage {
   requestedBy: string
   dateTimeRequested: string
   jobUuid: string
+  urns?: number[]
```
jon-shipley, line 8: or undefined?

jon-shipley (approved):
LGTM 👍

tslib/src/functions-ps-report/common/ps-report-service-bus-messages.ts
GuyHarwood: that's what the question mark is for

jon-shipley: I think the question mark means the property may be omitted.  But in the definition in the admin app the property may be there with an undefined value.

GuyHarwood: yeah good point, i will update this

PR #2789, opened 2024-05-14
Feature/61884 audit helpdesk impersonation

GuyHarwood:
- audit school impersonation events


db/migrations/schema-objects/20240514111912.do.school-impersonation-audit-table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE [mtc_admin].[schoolImpersonationAudit] (
```
jon-shipley, line 1: Whilst there is no _application_ need for an updatedAt col, I would suggest one anyway, with a DB trigger to catch any updates e.g. by DBAs or malicious actors.  This is standard for all tables I believe. 

I don't know what size we expect this table to reach but we could consider adding indexes on `school_id` and `user_id` to allow ad-hoc reporting during the day without adversely affecting performance.   Or we could wait and see if the index is actually required based on table growth.

jon-shipley (approved):
LGTM 👍 

A couple of suggestions made.

db/migrations/schema-objects/20240514111912.do.school-impersonation-audit-table.sql
GuyHarwood: I decided not to add the `updatedAt` column, as it seemed pointless.  However, it would allow us to identify if something had been modified, which it never should.

Will add the column, but will leave the indexes until such time that they are required.

PR #2790, opened 2024-05-14
feature/pupil-spa-localstorage-error-page-not-shown

jon-shipley:
Early use of the QR code system was using localstorage before it was detected, causing an unhandled error.  
This mod defers QrCodeUsageService initialisation until later.


GuyHarwood (approved):
LGTM

PR #2791, opened 2024-05-14
Feature/64320 pin gen disable buttons after click

jon-shipley:
Disable sticky banner button after form submit


GuyHarwood (approved):
LGTM

PR #2798, opened 2024-05-21
spike/64121-enable-call-stack-capture-for-app-insights

jon-shipley:
* generate sourcemaps in prod
* remove optimisation in prod (this only removed uglification) same size files
* update the size threshold to get it to compile


GuyHarwood (approved):
LGTM

PR #2799, opened 2024-05-24
Add bash script to run local e2e tests in parallel

jon-shipley:
* requires `parallel`


test/admin-hpa/run-local-tests.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/sh
+#set -x
+
+# Run the tests locally in parallel
+JOB_COUNT=6
```
GuyHarwood, line 5: would be nice to make this an optional param with a default of 6, or is there a particular reason you chose this?

test/admin-hpa/run-local-tests.sh
```diff
@@ -0,0 +1,83 @@
...
+
+writeSuccess ()
+{
+    MSG=$1
+  echo -e "\033[92m ${MSG} \033[0m"
+}
+
+test ()
+{
+  FILE=$1
```
GuyHarwood, line 58: can you document these params at the top of the file please, and also check they exist?

test/pupil-hpa/run-local-tests.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/sh
+
```
GuyHarwood, line 2: same comments apply to this as the admin one...

GuyHarwood (changes requested):
superb bit of work 🏆  this will make a huge difference

just a few comments that would improve the script for someone who's not used it before

test/admin-hpa/run-local-tests.sh
jon-shipley: 6 seems a good choice for an 8 core system, and was chosen after a lot of tests runs from 5-10.  It works well.  If anyone has wildly more cores we can make it a param.

test/admin-hpa/run-local-tests.sh
jon-shipley: 😀. This is the documentation.

In a bash function $1 refers to the 1st parameter passed to the function, $2 to the 2nd and so on.   So while it's possible to just use $1 or $2 within the function body, this assignment at least makes it readable.

GuyHarwood: yes, i am aware of the parameter indexing and $0 being the script itself.  

If you check for the existence of any required params near the top of the file and inform the user if anything is missing, it explains itself and stops it running without anything missing.

Here is [an example](https://github.com/DFEAGILEDEVOPS/MTC/blob/master/deploy/keys/function-masterkey-update-keyvault.sh)

GuyHarwood: It's no dealbreaker, so if you're averse to such an addition just leave as is...

GuyHarwood: i've now realised after discussion that these are local function arguments, not script arguments.
resolving this...

GuyHarwood (approved):
LGTM

PR #2806, opened 2024-06-06
Increase limit for express JSON requests

jon-shipley:
[pupil-api] Increase limit for express JSON requests

* new limit is 1MB


GuyHarwood (approved):
LGTM

PR #2807, opened 2024-06-07
bug/64795 Tech support - send service bus messages

GuyHarwood:
add functionality to send ad-hoc service bus messages manually

### DIFF
--- a/admin/services/queue-management.service.js
+++ b/admin/services/queue-management.service.js
@@ -31,6 +31,13 @@ const service = {
     const activeMessageCount = await serviceBusQueueAdminService.getQueueActiveMessageCount(queueName)
     if (activeMessageCount === 0) return
     return serviceBusQueueAdminService.clearQueue(queueName, activeMessageCount)
+  },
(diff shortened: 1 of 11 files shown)


admin/services/queue-management.service.js
```diff
@@ -31,6 +31,13 @@ const service = {
     const activeMessageCount = await serviceBusQueueAdminService.getQueueActiveMessageCount(queueName)
     if (activeMessageCount === 0) return
     return serviceBusQueueAdminService.clearQueue(queueName, activeMessageCount)
+  },
+
+  sendServiceBusQueueMessage: async function sendServiceBusQueueMessage (queueName, message, contentType) {
+    if (contentType === 'application/json') {
+      message = JSON.parse(message)
```
jon-shipley, line 38: Parsing could fail if the user supplied bad JSON.  Could you catch the error and bail out?

jon-shipley (approved):
LGTM

One comment for consideration

PR #2810, opened 2024-06-09
Increase maxAutoRenewDuration to align with functionTimeout

jon-shipley:
Ps report lock renewal update

Align maxAutoRenewDuration with functionTimeout


GuyHarwood (approved):
LGTM

PR #2811, opened 2024-06-09
Increase ttl on ps-report-export queue

jon-shipley:
Increase TTL on the ps-report-export queue to align with functionTimeout and maxAutoRenewDuration


GuyHarwood (approved):
LGTM

PR #2846, opened 2024-07-08
feature/64752-delete-deploy-sql

jon-shipley:
Remove unused code


deploy/sql/vso-validate-migrations/vso-migrate-down.sh
```diff
@@ -1,14 +0,0 @@
-#!/usr/bin/env bash
-
```
GuyHarwood, line 2: This is used to [validate all migrations](https://github.com/DFEAGILEDEVOPS/MTC/blob/master/deploy/build-defs/validate-pr-db-migrations.yml) as a gated PR check.  
We should move it to the `deploy/build-defs` directory or add the content from here and make it inline

deploy/sql/vso-validate-migrations/vso-migrate-up.sh
```diff
@@ -1,16 +0,0 @@
-#!/usr/bin/env bash
```
GuyHarwood, line 1: This is used to [validate all migrations](https://github.com/DFEAGILEDEVOPS/MTC/blob/master/deploy/build-defs/validate-pr-db-migrations.yml) as a gated PR check.  
We should move it to the `deploy/build-defs` directory or add the content from here and make it inline

deploy/sql/anonymise-sql-data/pupil-anon.sql
```diff
@@ -1,8 +0,0 @@
-UPDATE mtc_admin.pupil SET 
-    foreName=CAST(NEWID() AS VARCHAR(255)),
```
GuyHarwood, line 2: gut feel is we should save these anonymisation scripts as they may come in useful in the future.  maybe add under a misc section in the db app?

PR #2878, opened 2024-08-20
bug/59556 preserve restart state when changing attendance state

GuyHarwood:
Ensure that a deleted restart is reinstated if the attendance code is removed.
- capture the attendance instance when deleting the restart, during the addition of an attendance code
- undelete the restart when the attendance code is removed


jon-shipley (approved):
LGTM

PR #2911, opened 2024-09-16
bug/64863-check-received-error

jon-shipley:
Prevent pupils from creating keypress events unnecessarily
Implement Gzip (de)compression on the payload and modules that need access to it


GuyHarwood (approved):
LGTM 🏆

PR #2924, opened 2024-09-27
65515-copy-change

jon-shipley:
Update copy when site unavailable rather than read-only


jon-shipley:
@GuyHarwood @activemq I'll probably just merge this (after testing) so hopefully we get a nightly build ready for Monday.

PR #2925, opened 2024-10-02
Bug/65538 show system unavailable page for teachers

jon-shipley:
Bugfix: allow teachers to see the system unavailable page and not the DFE Sign In error page.


GuyHarwood (approved):
LGTM

PR #2926, opened 2024-10-03
Feature/65264 purging old blobs

GuyHarwood:
- utility script to clear down a storage account while preserving function containers (manual execution use)
- script to delete all blobs from a target container (used for ps report housekeeping)


deploy/storage/delete-old-blobs-key-auth.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
jon-shipley, line 1: Same portability mention as above.

GuyHarwood:
thank you @jon-shipley  all comments addressed

PR #2927, opened 2024-10-03
Feature/65300 app insights connection string

jon-shipley:
Maintenance

* Upgrade @microsoft/applicationinsights and @microsoft/applicationinsights-web from v2 to v3 
* Mandatory switch imposed by Microsoft: use a Connection String rather than an InstrumentationKey.


admin/config.js
```diff
@@ -123,10 +123,10 @@ module.exports = {
   },
   Monitoring: {
     ApplicationInsights: {
-      Key: process.env.APPINSIGHTS_INSTRUMENTATIONKEY,
+      Key: process.env.APPINSIGHTS_INSTRUMENTATIONKEY, // @deprecated
```
GuyHarwood, line 126: should we instead remove this if we are wholesale moving away from it, and there are no more usages?

docs/environment-variables.md
```diff
@@ -21,10 +21,9 @@ Env Var | Type | Default value | Required | Components | Description
...
- APPINSIGHTS_INSTRUMENTATIONKEY | String | NULL | Optional | AA,PA | The Azure Application Insights account key, required for logging / monitoring.
  APPINSIGHTS_LIVE_METRICS | Boolean | true | Optional | AA | Set to false if you do not want App Insights to collection live metrics.
  APPINSIGHTS_WINSTON_LOGGER | Boolean | false | Optional | AA | Boolean flag.  Set to 1 to allow the winston logger to send to Application Insights, or 0 to disable. If enabled, you should also set `EXPRESS_LOGGING_WINSTON`to `1` and `APPINSIGHTS_INSTRUMENTATIONKEY` to the Azure-specified value.
- APPLICATIONINSIGHTS_CONNECTION_STRING | String | '' | Optional, but should be set for Prod | AA (Others to follow in v3 rollout) | Connection string for v3 of Application Insights node package.  This env var name is automatically supported by AI.
+ APPINSIGHTS_CONNECTION_STRING | String | '' | Optional, but should be set for Prod | AA, FT, FC, FP, PAPI | Connection string for v3 of Application Insights node package.  This env var name is automatically supported by AI.
```
GuyHarwood, line 26: if its automatically picked up, should we be passing it in explicitly?

tslib/src/azure/app-insights.ts
```diff
@@ -1,32 +1,48 @@
-import * as appInsights from 'applicationinsights'
 import config from '../config'
+import { isNil } from 'ramda'
+const appInsights = require('applicationinsights')
+
+const cloudRoleName = 'TsLib'
```
GuyHarwood, line 5: this will get handed to all the functions.  i see there is a specific one for the pupil-api, but i wonder if we can override it for each function?  Out of scope for this PR i would say....

GuyHarwood (commented):
LGTM, happy to approve.  just a couple of questions...

docs/environment-variables.md
jon-shipley: Oh. That's an editing gap.  The 'automatically picked up' bit should have deleted when we did the name change, which is for consistency with other APPINSIGHTS env vars. 

I'll update it shortly.

admin/config.js
jon-shipley: I agree - good spot!!

tslib/src/azure/app-insights.ts
jon-shipley: This is only a default.  There is an override available by passing the cloudRole string into the startAppInsightsIfConfigured()` func.

GuyHarwood (approved):
LGTM

PR #2937, opened 2024-10-16
Bug/65517 axios requests failing to talk to azure functions API

GuyHarwood:
Axios requests fail with a refused connection.

Forcing IPV4 fixes the issue.
Configured at global level to ensure parity across solution.


jon-shipley (approved):
LGTM 👍

PR #2941, opened 2024-10-22
feature/65545 node 20 upgrade

GuyHarwood:
- upgrade to Node 20 LTS


jon-shipley:
There also needs to be an update to the various local.settings.json as, for some reason, we have 

`    "WEBSITE_NODE_DEFAULT_VERSION": "~18",`

Perhaps it's purpose is to remind us that this needs updating in Azure?

jon-shipley (approved):
lgtm

PR #2965, opened 2024-11-07
feature/65864 remove legacy check submission process

GuyHarwood:
remove legacy check submission assets, code and feature toggle


jon-shipley (approved):
LGTM

PR #2979, opened 2024-11-18
Feature/65231 disable gestures in pupil checks

jon-shipley:
Disable pull to refresh gestures in iOS and Chrome


GuyHarwood (approved):
LGTM

PR #2982, opened 2024-11-21
bug/66011 azure storage cannot send message using package v12.24.0

GuyHarwood:
sending message returns undefined, revert to 12.17.0


jon-shipley (approved):
LGTM 👍

PR #3004, opened 2024-12-16
bug/66196 build agent creation script failing due to os type

GuyHarwood:
MS [introduced a bug to the Azure CLI recently](https://github.com/Azure/azure-cli/issues/30439)  which causes our build agent creation script to fail.

Added os default to the call to fix this


jon-shipley (approved):
LGTM 👍

PR #3012, opened 2025-01-30
Test/windows setup

alison0:
Changes to page load and windows specific testing setup

### DIFF
--- a/test/se_manager/.cache/selenium/se-metadata.json
+++ b/test/se_manager/.cache/selenium/se-metadata.json
@@ -1,15 +1,31 @@
 {
-  "browsers": [],
+  "browsers": [
+    {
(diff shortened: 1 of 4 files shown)


test/se_manager/.cache/selenium/se-metadata.json
```diff
@@ -1,15 +1,31 @@
...
+      "browser_ttl": 1737992097
+    }
+  ],
   "drivers": [],
   "stats": [
     {
       "browser": "chrome",
       "browser_version": "132",
-      "os": "macos",
+      "os": "linux",
```
activemq, line 15: @alison0 I believe we do not need this. I would try running the tests without this, as it means it will use this for everyone running the tests

PR #3027, opened 2025-03-20
Reverting changes done as part of 66709 - see PR 3019

martincarson:
Reverted - SQL changes shouldn't be added to existing migration scripts

https://github.com/DFEAGILEDEVOPS/MTC/pull/3019


alison0 (approved):
Good to go

PR #3029, opened 2025-04-03
Revert "Feature/65678 upgrade angular"

martincarson:
Reverting this change to narrow down root cause of failure in pipeline


martincarson:
Not required

PR #3086, opened 2025-09-02
Updated version of selenium driver to 4.35.0

martincarson:
Also:
- Updated the approach used for installing Google Chrome

- Added one update to the edit pupil test, the verification is split in two, fixed bug locally


alison0:
Approved

PR #3321, opened 2026-06-25
Documentation for CTF not updating - workaround

martincarson:
Added documentation explaining a workaround for when the name is updated after the hdf is signed - could automate in future


alison0:
Excellent overview capturing the issues involved.

alison0 (approved):
GTG

PR #3355, opened 2026-09-14
Feature/e2e test fix

hannahmorgan24:
- Stabilized the Playwright pupil flow by making the intro screen button clicks more resilient when the SPA swaps/detaches elements mid-navigation.
- Increased the overall test timeout and extended the completion-check polling window to handle asynchronous backend processing before the admin count updates.
- Improved failure diagnostics so the test reports a clearer error when the check never reaches the “Complete” state, including the current pupil status summary.
- Some extra cleanup steps to ensure check window is as it was prior to testing
- Replaced hardcoded dates for maintainability


martincarson (approved):
Looks good to merge

